PR #68, opened 2023-07-05
Test for nonce reuse

nickfarrow:
Resolves https://github.com/frostsnap/frost-esp32/issues/14
Builds on and based on testing approach in https://github.com/frostsnap/frost-esp32/pull/66

### DIFF
--- a/frostsnap_core/tests/malicious.rs
+++ b/frostsnap_core/tests/malicious.rs
@@ -1,6 +1,8 @@
 //! Tests for a malicious actions. A malicious coordinator, a malicious device or both.
-use frostsnap_core::message::{CoordinatorToDeviceMessage, KeyGenProvideShares};
-use frostsnap_core::FrostSigner;
+use frostsnap_core::message::{
+    CoordinatorSend, CoordinatorToDeviceMessage, DeviceSend, KeyGenProvideShares, SignTask,
+};
+use frostsnap_core::{FrostCoordinator, FrostSigner};
 use rand_chacha::rand_core::SeedableRng;
 use rand_chacha::ChaCha20Rng;
 use schnorr_fun::frost;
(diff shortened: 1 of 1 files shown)


LLFourn:
Hey can we put all the tests in `malicious.rs` that are to do with a malicious device or coordinator. What's the difference between this new malicious poly test and the old one?

frostsnap_core/tests/malicious.rs
```diff
@@ -34,3 +36,56 @@ fn keygen_maliciously_replace_public_poly() {
...
+        .unwrap();
+
+    let _device_responses = device
+        .recv_coordinator_message(sign_request.clone())
+        .unwrap();
+
+    let _device_sends = device.sign_ack(true).unwrap();
+
+    // Receive the same sign request again with the same nonce
+    let sign_request_result = device.recv_coordinator_message(sign_request.clone());
```
LLFourn, line 80: Nit: shouldn't this be a different sign request. If it's precisely the same one it's technically fine to carry on right?. I guess in practice we will not "recall" the sign request that occurs so even the same sign request will throw an error so maybe to fine to leave as is.

frostsnap_core/tests/malicious.rs
```diff
@@ -34,3 +36,56 @@ fn keygen_maliciously_replace_public_poly() {
...
+
+    // TODO: Specific error messages and match on NonceReuse
```
LLFourn, line 82: Not sure this is a good idea see: https://github.com/frostsnap/frostsnap/pull/91#issuecomment-1732983696

LLFourn (approved):
Thanks! Left two minor comments.

nickfarrow:
Now creates a new sign request and that has the same nonces as the previous signing session

frostsnap_core/tests/malicious.rs
```diff
@@ -34,3 +36,69 @@ fn keygen_maliciously_replace_public_poly() {
...
+        .start_sign(SignTask::Plain(b"utxo.club!".to_vec()), device_ids.clone())
+        .unwrap();
+
+    let _device_responses = device
+        .recv_coordinator_message(sign_request.clone())
+        .unwrap();
+
+    let _device_sends = device.sign_ack(true).unwrap();
+
+    // Receive the same sign request again with the same nonces
```
LLFourn, line 79:
```suggestion
    // Receive the same sign request again with the same nonces
```
This comment is outdated :)

PR #77, opened 2023-07-18
Coordinator core

LLFourn:
Re-engineers the coordinator code quite a bit and extracts it into its own module. It's going to be interesting to figure out how to generalize the UI over signing keygen on the coordinator side but that is not done here with all UI stuff being handled in `coordiantor-cli`.


frostsnap_coordinator/src/usb_serial_manager.rs
```diff
@@ -183,178 +152,212 @@ impl Ports {
...
+        outbox.retain_mut(|send| {
+            let mut ports_to_send_on = HashSet::new();
+            let mut wire_message = send.clone();
+            wire_message.target_destinations.clear();
+
+            send.target_destinations.retain(|destination| {
+                match self.device_ports.get(destination) {
+                    Some(port) => {
+                        ports_to_send_on.insert(port.clone());
+                        wire_message.target_destinations.insert(*destination);
```
nickfarrow, line 314: Does this `wire_message` accumulate new target destinations, what if the target destinations are on different ports?

Would the `wire_message` have both target destinations and get sent along both serial ports?

LLFourn: > Does this `wire_message` accumulate new target destinations, what if the target destinations are on different ports?
> 
> Would the `wire_message` have both target destinations and get sent along both serial ports?

It sends the same wire message on all the relevant ports (but i didn't test this because I only have one USB cable!).

nickfarrow: Tested 👍

nickfarrow:
I've realised there's code in `coordinator-cli` which we probably want to move into `coordinator-core`.

Things like `signer.rs` and maybe some logic from `main.rs`, previously I reused this code to try add a rust gui on top.

A good state to be in would be to have a `frostsnap_coordinator` library that provides very high level coordinator functions like `do_keygen`, `run_signing`.

LLFourn:
> A good state to be in would be to have a `frostsnap_coordinator` library that provides very high level coordinator functions like `do_keygen`, `run_signing`.

@nickfarrow right but I made the conscious decision to avoid doing that in this PR because I don't know how it should look yet.

Issue #58, opened 2023-06-28
Investigate and fix allocation failures


LLFourn:
We're running out of memory frequently especially after doing multiple signings after a keygen. It always happens while decoding a new message.

This is most likely due to the way we are buffering things. We are making too many allocations to decode a single message. We should try and aim for no allocations for read buffers. There are lots of ways we are being memory inefficient all around the place but I would not fix those as yet because they will just push the problem later down the track.

I'd suggest at this stage also displaying some raw numbers on the blue boards about how much memory is being used while debugging this. I think you can do this by querying the allocator.

nickfarrow:
After some timing practice I can get the same allocation failure (`127800`) bytes every time, just after powering on and upon the first signing attempt. You have to click OK on the downstream device ~50ms before OK on the device closer to the coordinator.

So I do think there is a read being blocked or timing issue. Signing computation on the upstream device is preventing it from reading the signature.

LLFourn:
Ok sure. That's one way of making this happen. But you can make it happen without any timing edge cases. I've done it plenty of times. Are you saying you can't reproduce this?

LLFourn:
Fixed by #79

PR #79, opened 2023-07-21
Serial Ring Buffer

nickfarrow:
Fixes allocation failures https://github.com/frostsnap/frostsnap/issues/58 , I haven't got any after signing many times (5+, 4-of-4) in a row with similar timed presses.

Things feel much faster in general.

Marking draft because
- ~[ ] `upstream_buffer_uart` and `upstream_buffer_jtag` could be from the same buffer.~
- [x] I took the consuming outside of `find_and_remove_magic_bytes` because i'm keeping track of `buffer_filled` and rotating outside. Edit: just renamed functions
- [x] Broke magic bytes tests
- [x] Sometimes the end device (of 4) is failing to register
- [x] Get rid of RingBuffer `Vec`s. Edit: i think i got rid of all the vecs that I can? Only remaining vec is the peek for magic bytes

### DIFF
--- a/device/src/io.rs
+++ b/device/src/io.rs
@@ -19,12 +19,11 @@ use frostsnap_comms::Direction;
 use frostsnap_comms::Downstream;
 use frostsnap_comms::MagicBytes;
 use frostsnap_comms::Upstream;
-use ringbuffer::RingBufferExt;
 use ringbuffer::RingBufferRead;
 use ringbuffer::RingBufferWrite;
 use ringbuffer::{AllocRingBuffer, RingBuffer};
 
-pub const RING_BUFFER_SIZE: usize = 2usize.pow(13);
+const RING_BUFFER_SIZE_LOG_2: usize = 8; // i.e. 256 bytes
 
 pub struct SerialInterface<'a, T, U, D> {
     io: SerialIo<'a, U>,
@@ -38,7 +37,7 @@ impl<'a, T, U, D> SerialInterface<'a, T, U, D> {
     pub fn new_uart(uart: uart::Uart<'a, U>, timer: &'a Timer<T>) -> Self {
         Self {
             io: SerialIo::Uart(uart),
-            ring_buffer: AllocRingBuffer::with_capacity(RING_BUFFER_SIZE),
+            ring_buffer: AllocRingBuffer::with_capacity_power_of_2(RING_BUFFER_SIZE_LOG_2),
             magic_bytes_progress: 0,
             timer,
             direction: PhantomData,
@@ -54,7 +53,7 @@ impl<'a, T, U> SerialInterface<'a, T, U, Upstream> {
     pub fn new_jtag(jtag: UsbSerialJtag<'a>, timer: &'a Timer<T>) -> Self {
         Self {
             io: SerialIo::Jtag(jtag),
-            ring_buffer: AllocRingBuffer::with_capacity(RING_BUFFER_SIZE),
+            ring_buffer: AllocRingBuffer::with_capacity_power_of_2(RING_BUFFER_SIZE_LOG_2),
             magic_bytes_progress: 0,
             timer,
             direction: PhantomData,
@@ -67,32 +66,28 @@ where
     U: uart::Instance,
     T: esp32c3_hal::timer::Instance,
 {
-    pub fn poll_read(&mut self) -> bool {
-        let mut read_something = false;
+    fn fill_buffer(&mut self) {
         while let Ok(c) = self.io.read_byte() {
             self.ring_buffer.push(c);
-            read_something = true;
+            if self.ring_buffer.is_full() {
+                break;
+            }
         }
-        read_something
     }
 
     pub fn find_and_remove_magic_bytes(&mut self) -> bool
     where
         D: Direction,
     {
-        self.poll_read();
-        if self.ring_buffer.len() == 0 {
+        self.fill_buffer();
+        if self.ring_buffer.is_empty() {
             return false;
         }
-
-        let (consume, progress, found) = frostsnap_comms::make_progress_on_magic_bytes::<D>(
-            &self.clone_buffer_to_vec()[..],
+        let (progress, found) = frostsnap_comms::make_progress_on_magic_bytes::<D>(
+            self.ring_buffer.drain(),
             self.magic_bytes_progress,
         );
         self.magic_bytes_progress = progress;
-        for _ in 0..consume {
-            self.ring_buffer.skip();
-        }
         found
     }
 }
@@ -123,8 +118,16 @@ where
 
     pub fn receive_from_downstream(
         &mut self,
-    ) -> Result<DeviceSendSerial<Downstream>, bincode::error::DecodeError> {
-        bincode::decode_from_reader(self, bincode::config::standard())
+    ) -> Option<Result<DeviceSendSerial<Downstream>, bincode::error::DecodeError>> {
+        self.fill_buffer();
+        if !self.ring_buffer.is_empty() {
+            Some(bincode::decode_from_reader(
+                self,
+                bincode::config::standard(),
+            ))
+        } else {
+            None
+        }
     }
 }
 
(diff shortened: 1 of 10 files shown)


nickfarrow:
I had a go at implementing a shared upstream buffer for jtag and uart.

This commit uses raw pointers and requires an unsafe https://github.com/frostsnap/frostsnap/commit/65dce7b9ee92edb485744728c94db0c3db063bea. I assume we want to avoid.

Initially I tried using a Refcell but it broke existing code by moving serial_interface within our loop. Might have another go.

Having a duplicate & unused serial buffer is probably not a big concern anyway..?

nickfarrow:
BTW if a device ever seems completely bricked, you might be able to spam flash while plugging in
```
while true; do
    cargo run --release --bin blue --features blue
    exit_code=$?

    if [[ $exit_code -eq 0 ]]; then
        break
    fi
done
```

LLFourn (commented):
Haven't reviewed code in depth yet -- what do you think about using a `no_std` ring buffer from crates.io e.g. https://crates.io/crates/ringbuffer (there are others). The advantage would be not having to review or test this code.

it would be awesome if we could unify the code around reading messages and magic bytes into one type that is used from both coordinator and device as well. The problem with doing this now is that the coordinator uses `std::io` read buffer. We could just chose a `no_std` impl and implement the bincode traits for it and used it in both places.

LLFourn:
Another approach we could take would be to use [`core2`](https://crates.io/crates/core2) to unify the code between coordinator and device. If we're on `std` then it will use the normal `std::io::*` traits but if not it will export it's own `no_std` compatible triats and reader. This would allow us to implement something like `FramedSerialPort` that exists in `frostsnap_coordiantor` in `frostsnap_comms`. There are two problems with this:

1. `core2` is not really maintained anymore (not a big issue just paste the code into our project and credit the author).
2. We need more than the `Read` trait because we need to know if there's a byte in the buffer without reading it. I don't think our hardware gives us this ability so we might be stuck with manually buffering stuff (so you're current approach is fine).

LLFourn:
> BTW if a device ever seems completely bricked, you might be able to spam flash while plugging in
> 
> ```
> while true; do
>     cargo run --release --bin blue --features blue
>     exit_code=$?
> 
>     if [[ $exit_code -eq 0 ]]; then
>         break
>     fi
> done
> ```

The problem here is likely we are not re-enabling JTAG in the panic handler. We have an issue on trello board for that.

nickfarrow:
Past attempts to fix end-device registration:
* Lower/higher read magic bytes periods
* Random/different magic byte intervals between devices
* Not soft resetting immediately, confirmed we are not restarting.

Commit im using to try debug
e8e69f0df81cbb40376cde3ce0b783b28342276a

device/src/io.rs
```diff
@@ -59,19 +68,32 @@ where
...
         D: Direction,
     {
         self.poll_read();
-        frostsnap_comms::find_and_remove_magic_bytes::<D>(&mut self.read_buffer)
+        if self.ring_buffer.len() == 0 {
+            return false;
+        }
+
+        let (consume, progress, found) = frostsnap_comms::make_progress_on_magic_bytes::<D>(
+            &self.clone_buffer_to_vec()[..],
```
LLFourn, line 89: The point of this PR is to remove allocations during the decoding on messages process but this does allocations. Why can't make progress on magic bytes just work with the ring buffer. Surely the API can be changed to work with whatever you pass in from the ring buffer.

LLFourn:
IMO the 

> Past attempts to fix end-device registration:
> 
>     * Lower/higher read magic bytes periods
> 
>     * Random/different magic byte intervals between devices
> 
>     * Not soft resetting immediately, confirmed we are not restarting.
> 
> 
> Commit im using to try debug [e8e69f0](https://github.com/frostsnap/frostsnap/commit/e8e69f0df81cbb40376cde3ce0b783b28342276a)

Isn't the problem finding the difference in behavior of your new code vs the old code. I didn't quite understand how doing the above is going to help you do that. I'd love to work on this with you. Perhaps Tuesday?

nickfarrow:
I haven't been as rigorous as I should be when documenting my experiments while debugging this. Something I want to improve. 

* I am almost certain this bug existed on master with 4 devices (maybe could see if it occurs with your 4+ printed boards)
* But now I only have 3 devices so can't confirm this
* My belief is that changes in read/write speed or timing with this branch has caused it to occur with 3 devices

But this is why I am not looking into these code changes specifically. Tuesday would be great

nickfarrow:
One source of ~large memory usage when unplugging- the end device is
```
Err(e) => {
    // sends_upstream.push(DeviceSendMessage::Debug {
    //     message: format!("Failed to decode on downstream port: {e}"),
    //     device: frost_signer.device_id(),
    // });
...
```

Even though the `DeviceSendMessage::Debug` is 168 bytes, it might be allocating the max `3248` bytes from `DeviceSendMessage::Core`. Investigating. And wondering whether this has anything to do with clippy's suggestion to Box certain enum variants.

nickfarrow:
The `used/free` numbers aren't changing smoothly as I would have expected, testing like
```
let mut waste_of_space = vec![];
loop {
    waste_of_space.push(0xff);
    ui.set_workflow(ui::Workflow::OnScreenDebug(debug_log.clone())); // calls render
}
```

nickfarrow:
Confirmed that high memory usage is being caused by large variants of `DeviceSendMessage`, that are filling up `sends_upstream` while the device is waiting for a coordinator. Using `Box` dramatically helped 0553a2bdcb46c3dfe3e879e7c76a480de34f6267

Unfortunately it is still possible to get the crazy near doubling of used memory though! This happens more if we heavily we use the `sends_upstream` vector. To me it feels as though something is wrong with freeing up memory when allocating within this vector... 

A reason I think this: if I queue up a whole bunch of messages and then allow them to drain by connecting a coordinator, the free memory only returns to roughly where it last doubled. Potentially related? https://github.com/esp-rs/esp-alloc/issues/6

 If we keep `sends_upstream` relatively empty by immediately forwarding (`frostsnap-cli listen`: 6a1710f96dee3a07c776df484e7e98493b93f203). The device maintains relatively constant memory usage. This listen command also reveals a flurry of `DeviceSendMessage::Debug`s seen when unplugging the end device, potentially explaining the alloc failure as a bombardment of messages into `sends_upstream`.

frostsnap_core/src/message.rs
```diff
@@ -77,7 +78,7 @@ pub enum CoordinatorToStorageMessage {
 #[derive(Clone, Debug, bincode::Encode, bincode::Decode)]
 pub struct DeviceToCoordindatorMessage {
     pub from: DeviceId,
-    pub body: DeviceToCoordinatorBody,
+    pub body: Box<DeviceToCoordinatorBody>,
```
LLFourn, line 81: Ok so the reason this helps is that DeviceToCoordinatorBody is a huge size because `KeyGenResponse` is huge which makes `DeviceSend` huge. I don't think this solves the root problem though which is that `KeyGenResponse` is huge.  `KeyGenResponse` is still making `DeviceToCoordinatorBody::SignatureShare` much bigger than it needs to be. The more targeted solution is to `Box` the array:

```rust
#[derive(Clone, Debug, bincode::Encode, bincode::Decode, Eq, PartialEq)]
pub struct KeyGenResponse {
    pub encrypted_shares: KeyGenProvideShares,
    pub nonces: Box<[Nonce; NONCE_BATCH_SIZE]>,
}
```

I wonder if we can come up with a more general way to stop this from happening again. It seems like a bug you could easily add in a release and wait for end users to notice it. Did clippy warn you of this? Maybe we can configure that check to be even more sensitive.

LLFourn (changes requested):
Thanks for doing this investigation. Nice fix. See comment.

frostsnap_core/src/message.rs
nickfarrow: Yep this was noticed from a clippy warning, we could make the clippy test fail on warnings, or something specific with [enum-variant-size-threshold](https://doc.rust-lang.org/nightly/clippy/lint_configuration.html?highlight=enum#enum-variant-size-threshold)

nickfarrow:
Dropped memory display commit 57331627a7b916c3d1df951c33436bb58aadf7b4 to leave for appropriate PR https://github.com/frostsnap/frostsnap/pull/83, best rebase onto that branch for any further testing of this PR

LLFourn (approved):
I pushed 42b14ec4e9816f0d0a441abc187c7efcff7605d3 to clean up some dependencies (it wasn't compiling for me due to ringbuffer `0.14.1` being yanked.

LLFourn:
I confirmed this fixes the main problem of way too much memory being allocated per insertion. Nice work @nickfarrow.

PR #80, opened 2023-08-09
Add Frostsnap app aka Frostsnapp

LLFourn:
Two main commits that can be reviewed separately:

## 1

- Rearranged serial interfaces to that they were as simple as possible
while being flexible enough for what we needed on android side.
- Made poll_ports return a kind of changeset which makes notifying UI of
changes easier and actually makes the cli workflow more robust as well.
- We strictly set the interface to UART mode first so we don't have the JTAG interface appear to the OS until it's its turn. (note this makes flashing kind of annoying since you have to hard reset each time you want to flash -- we could make this easier with a feature flag).

## 2 

Dumps the flutter app code which at the moment detects devices that are connected and disconnected and lets you give them names. The main file to look are:

- `native/src/api.rs`
- everything in `lib` that wasn't auto-generated

See if you can run the app on your laptop and or phone (for android phone you'll need android studio). Low chances any of this will work first time and we'll probably have to get on a call to figure it out.

Note that I think the whole flutter app should be in this repo for now otherwise it will be hell to develop features at the same time for the app and devices.


frostsnapp/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+This repository serves as a template for Flutter projects calling into native Rust
+libraries via `flutter_rust_bridge`.
+
+## Install
+
+To begin, ensure that you have a working installation of the following items:
+- [Flutter SDK](https://docs.flutter.dev/get-started/install)
+- [Rust language](https://rustup.rs/)
+- `flutter_rust_bridge_codegen` [cargo package](https://cjycode.com/flutter_rust_bridge/integrate/deps.html#build-time-dependencies)
```
nickfarrow, line 11: You may also need `cargo-expand` for flutter rust bridge codegen:
```
cargo install cargo-expand
```

frostsnapp/README.md
```diff
@@ -0,0 +1,33 @@
...
+- Appropriate [Rust targets](https://rust-lang.github.io/rustup/cross-compilation.html) for cross-compiling to your device
+- For Android targets:
+    - Install [cargo-ndk](https://github.com/bbqsrc/cargo-ndk#installing)
+    - Install [Android NDK 22](https://github.com/android/ndk/wiki/Unsupported-Downloads#r22b), then put its path in one of the `gradle.properties`, e.g.:
+
+```
+echo "ANDROID_NDK=.." >> ~/.gradle/gradle.properties
```
nickfarrow, line 18: Might also need to use `flutter config --android-sdk  <ANDROID_NDK_DIR>`, or reopen terminal

nickfarrow:
Currently getting stuck at `just gen` with
```
flutter_rust_bridge_codegen --rust-input native/src/api.rs --dart-output lib/bridge_generated.dart --c-output ios/Runner/bridge_generated.h --extra-c-output-path macos/Runner/ --dart-decl-output lib/bridge_definitions.dart --wasm
2023/08/25 00:40:20 [INFO] Running cargo expand in '/home/nick/repos/frostsnap/frostsnapp/native'
2023/08/25 00:40:20 [ERROR] panicked at 'not yet implemented: generate code for type "PortDesc"', /home/nick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-1.81.0/src/ir/ty_unencodable.rs:59:9
thread 'main' panicked at 'not yet implemented: generate code for type "PortDesc"', /home/nick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-1.81.0/src/ir/ty_unencodable.rs:59:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: Recipe `gen` failed on line 11 with exit code 101
```
Do i recall you mentioning something along these lines? Will update as i progress.

frostsnapp/README.md
LLFourn: Hmmm I wonder why that would help. grade.properties is used by gradle not by your shell unless there's something I don't understand.

LLFourn:
> Currently getting stuck at `just gen` with
> 
> ```
> flutter_rust_bridge_codegen --rust-input native/src/api.rs --dart-output lib/bridge_generated.dart --c-output ios/Runner/bridge_generated.h --extra-c-output-path macos/Runner/ --dart-decl-output lib/bridge_definitions.dart --wasm
> 2023/08/25 00:40:20 [INFO] Running cargo expand in '/home/nick/repos/frostsnap/frostsnapp/native'
> 2023/08/25 00:40:20 [ERROR] panicked at 'not yet implemented: generate code for type "PortDesc"', /home/nick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-1.81.0/src/ir/ty_unencodable.rs:59:9
> thread 'main' panicked at 'not yet implemented: generate code for type "PortDesc"', /home/nick/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-1.81.0/src/ir/ty_unencodable.rs:59:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> error: Recipe `gen` failed on line 11 with exit code 101
> ```
> 
> Do i recall you mentioning something along these lines? Will update as i progress.

No I don't think so. `PortDesc` is a mirrored type (the only one I think) so it is likely something going wrong there. I just upgraded flutter_rust_bridge to the version you're using locally (pushed a commit to upgrade it on the repo) and it seems to work. Maybe it's because you were using `1.81` CLI while everything else was on `1.79`. I've updated everything to `1.81` now and included in the README.md that you should use the `flutter_rust_bridge_codegen` version from the `Cargo.lock`.

frostsnapp/README.md
nickfarrow: I think it came up as a suggestion at some point during the install process. Maybe ignore for now

frostsnapp/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+This repository serves as a template for Flutter projects calling into native Rust
+libraries via `flutter_rust_bridge`.
+
+## Install
+
+To begin, ensure that you have a working installation of the following items:
+- [Flutter SDK](https://docs.flutter.dev/get-started/install)
+- [Rust language](https://rustup.rs/)
+- `flutter_rust_bridge_codegen` [cargo package](https://cjycode.com/flutter_rust_bridge/integrate/deps.html#build-time-dependencies) -- **Make sure to install the same version as in the project's [`Cargo.lock`](../Cargo.lock)**
```
nickfarrow, line 11: Note: on non-debian linux you need to set `export CPATH="$(clang -v 2>&1 | grep "Selected GCC installation" | rev | cut -d' ' -f1 | rev)/include"`

frostsnapp/README.md
```diff
@@ -0,0 +1,34 @@
+# frostsnapp
+
+This repository serves as a template for Flutter projects calling into native Rust
+libraries via `flutter_rust_bridge`.
+
+## Install
+
+To begin, ensure that you have a working installation of the following items:
+- [Flutter SDK](https://docs.flutter.dev/get-started/install)
+- [Rust language](https://rustup.rs/)
```
nickfarrow, line 10: * [Dart](https://dart.dev/get-dart)
* [just](https://github.com/casey/just) for running commands from justfile
* [ninja](https://ninja-build.org/) for making the flutter app

nickfarrow:
That fix works! After a few more dependencies / env additions I got the app running on linux. Awesome

LLFourn:
I've pushed a few commits to make things work (android app was stuttering for me).

I've also done another refactor of github actions. It's not simpler and does clippy on everything. Added a root `justfile` too with a bunch of handy things. The app is being built for linux and android in CI now!

I think this is ready for final review.

frostsnapp/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+``` sh
+flutter run
+```
+
+## Build
+```
+just build linux
+just build apk
+```
```
nickfarrow, line 45: To run this APK on an android phone, you need to install [android-tools](https://wiki.archlinux.org/title/Android_Debug_Bridge) so you can use `adb` (android debug bridge) . The device must be in [debug mode](https://www.lifewire.com/enable-usb-debugging-android-4690927) and show as "authorized" under `adb devices`.

Then
```
adb install -r build/app/outputs/flutter-apk/app-release.apk
```

LLFourn: flutter run should do this though

frostsnapp/LICENSE-MIT
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright 2022 Viet Dinh.
```
nickfarrow, line 3: Note to self: make commit changing/removing all these licenses

frostsnapp/README.md
nickfarrow: Oh interesting I see my device now, I guess it wasnt in debug mode & authorised before.

Will add commit with some suggestions

frostsnapp/native/src/api.rs
```diff
@@ -0,0 +1,196 @@
...
+
+#[derive(Debug)]
+pub struct PortOpen {
+    pub id: String,
+    pub baud_rate: u32,
+    pub ready: RustOpaque<PortOpenSender>,
+}
+
+impl PortOpen {
+    pub fn satisfy(&self, err: Option<String>) {
```
nickfarrow, line 85: I don't quite understand the purpose of this function. Something like, if err is None then the port is open?

Is this the same `.satisfy()` i see being called in the dart code? If so, how does passing bytes into here work?

nickfarrow:
Notes after going through project so far:

---

`frostsnapp/native/src/api.rs`

We have event streams:
* General coordinator `EVENT_STREAM`
* a `DEVICE_EVENT_STREAM`
* and `PENDING_DEVICE_EVENTS`

When emitting device events we take them from pending and add them to `DEVICE_EVENT_STREAM`.

General coordinator events are currently just port management.


Device events `DeviceChange` are currently (Added, Registered, Disconnected). Would `KeyGenResponded` and `KeyGenerated` be an example of a future variants?

We have port controlling structs `PortOpen`, `PortRead`, `PortWrite`, and `PortBytesToRead` each with a `satisfy` method that will be called in dart code.

---

`frostsnapp/native/src/coordinator.rs`

We have an `FfiCoordinator` which creates a `FrostCoordinator`. If `host_handles_serial` (i.e. android) then we have a separate `UsbSerialManager` and `ffi_serial`. On desktop everything is accessible through just the manager, but on android we need to access through this `ffi_serial`.


When creating the `FfiCoordinator` we spawn a thread that processes serial messages. We continuously loop, polling the manager and emit and device changes. For new messages from devices we receive them to the coordinator and append responses to an outbox. Currently I can not see where we process this outbox and send messages out.

The `FfiCoordinator` can also set device labels on `new_device_labels` which will at some point in the future need to write to some local state.

Implementations of `io::Read` and `io::Write` on the `FfiSerialPort` emit coordinator events `PortRead` and `PortWrite` (above).


---

`frostsnapp/lib/coordinator.dart`



Creates the FFICoordinator (handling serial if android).

Listens to usb serial event stream and updates openPorts with connects/disconnects.

Handles events (spawned from that thread loop in `coordinator.rs`), calling appropriate read/writes.

Does this `Coordinator() {...}``  loop?




---

nickfarrow:
Thanks @LLFourn

Ah I see the `global_coordinator.subDeviceEvents().forEach(...` in the dart code now.

Tested on desktop and mobile

nickfarrow (approved):

PR #81, opened 2023-08-14
Frostsnap Dev Board v1.1

musdom:
bin and features to support the new custom PCB

### DIFF
--- a/device/.cargo/config.toml
+++ b/device/.cargo/config.toml
@@ -1,5 +1,5 @@
 [target.riscv32imc-unknown-none-elf]
-runner = "espflash flash --baud 921600 --bootloader bootloader.debug.bin"
+runner = "espflash flash --baud 921600 --bootloader bootloader.bin"
 
 [build]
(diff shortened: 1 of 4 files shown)


device/.cargo/config.toml
```diff
@@ -1,5 +1,5 @@
 [target.riscv32imc-unknown-none-elf]
-runner = "espflash flash --baud 921600 --bootloader bootloader.bin"
+runner = "espflash flash --baud 921600 --bootloader bootloader.debug.bin"
```
LLFourn, line 2: what's going on here?

musdom: Ah, this was me looking at 2nd stage bootloader logs before trying to enable flash encryption. Pushing new commit my bad

LLFourn (approved):
tACK 06a3be3647e1941620ff6da6d9c2be5ad22e87d6

looks like we have officially chritened this device the frostypede :hand_over_mouth:

PR #83, opened 2023-08-17
Onscreen debugging

nickfarrow:
This branch is particularly helpful for seeing memory usage and messages on the device, both shown onscreen.


nickfarrow:
Closing this because it is so far out of date and ui has had a lot of changes.
There are some useful snippets in here for displaying memory usage and a message log if we ever need one again.

PR #84, opened 2023-08-26
Clippy test

nickfarrow:


nickfarrow:
I don't understand why there is a broken dependency with rust bitcoin now
```
    Compiling ahash v0.3.8
error[E0432]: unresolved import `core::sync::atomic::AtomicUsize`
 --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.3.8/src/random_state.rs:3:5
  |
3 | use core::sync::atomic::AtomicUsize;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `AtomicUsize` in `sync::atomic`
```
The problem is fixable by bumping versions of a lot of things, but why has this problem suddenly popped up in the first place?

LLFourn:
I'm not sure but I think that `atomic` doesn't exist on esp32 so this would never have worked. I think we shouldn't depend on `ahash`.

LLFourn:
> I don't understand why there is a broken dependency with rust bitcoin now
> 
> ```
>     Compiling ahash v0.3.8
> error[E0432]: unresolved import `core::sync::atomic::AtomicUsize`
>  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.3.8/src/random_state.rs:3:5
>   |
> 3 | use core::sync::atomic::AtomicUsize;
>   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `AtomicUsize` in `sync::atomic`
> ```
> 
> The problem is fixable by bumping versions of a lot of things, but why has this problem suddenly popped up in the first place?
> 
> Edit: comes with adding general rust env `-Dwarnings` as opposed to just on the clippy command.
> 
> Weird to me that this change [b9d7570](https://github.com/frostsnap/frostsnap/commit/b9d7570c5295c69bf5a135ee59d0ebf7584d44c7) causes the previously unseen build error above..

Note I think this means you are somehow compiling the device dependencies for `riscvimc` but not on a nightly compiler somehow. I haven't been able to find the issue online which details why we need nightly compiler to stop atomic int stuff from failing to compile.

PR #87, opened 2023-09-06
Run clippy in ci

LLFourn:
- Run clippy in ci
- Box the things
- Make DeviceId a `[u8;33]` to satisfy clippy in some areas
- Rename some things around `sign_task` because they bothered me

replaces #84

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,20 +7,19 @@ on:
   pull_request:
  
 env:
-  rust_toolchain: stable
-  
+  LIB_PACKAGES: -p frostsnap_comms -p frostsnap_core -p frostsnap_coordinator
+
 jobs:
   device-build:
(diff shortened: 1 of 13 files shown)


nickfarrow:
Looks good to me! Confirmed it improves and includes all changes of https://github.com/frostsnap/frostsnap/pull/84

I might spend a moment trying to understand internals of `frostsnap_core/src/macro.rs` a bit better. Did you create just now for this purpose or did it come from somewhere?

LLFourn:
oh yes. It's code I paste around in almost every project I do to make things that can be represented as fixed length byte array implement a bunch of traits like serde serialize, deserialize and a few others. It should probably be turned into its own crate but I find myself tweaking small things about it each time.

nickfarrow (approved):

nickfarrow:
Tested on blue devices!

Issue #86, opened 2023-09-06
Devices should report when downstream device is disconnected to coordinator


LLFourn:
- device should quickly report when downstream is disconnected on screen. It should use GPIO for this.
- should send a message to coordinator right away
- coordinator should somehow know which devices this implies are disconnected.

nickfarrow:
> should send a message to coordinator right away

I think we currently send a debug message, but are you suggesting we send something like a `DeviceToCoordinator::DownstreamDisconnect` to the coordinator that the app can react to?

LLFourn:
Yes :)

LLFourn:
This was done a while ago

PR #89, opened 2023-09-08
DownstreamDisconnect message

nickfarrow:
A start on implementing https://github.com/frostsnap/frostsnap/issues/86

Currently the device has no knowledge of the device_id connected immediately downstream. Perhaps we can get this through ordering of messages which are received. I.e. first "Announce" gets stored to the variable

```
downstream_connected = Some(device_id)
```
instead of `true/false`.

We also need to decide how to handle these messages when the coordinator receives these messages during keygen or signing.

### DIFF
--- a/device/src/esp32_run.rs
+++ b/device/src/esp32_run.rs
@@ -12,7 +12,8 @@ use frostsnap_comms::{
 };
 use frostsnap_comms::{DeviceReceiveMessage, Downstream};
 use frostsnap_core::message::{
-    CoordinatorToDeviceMessage, DeviceSend, DeviceToCoordinatorBody, DeviceToUserMessage,
+    CoordinatorToDeviceMessage, DeviceSend, DeviceToCoordinatorBody, DeviceToCoordindatorMessage,
+    DeviceToUserMessage,
 };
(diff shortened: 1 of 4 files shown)


musdom:
GPIO13 low means downstream device is present

LLFourn:
What I was thinking was that the announce message would just have a `upstream: Option<DeviceId>` which the upstream device would set to `Some(my_device_id)` as it passes through it (if it was set to `None`).

device/src/esp32_run.rs
```diff
@@ -130,6 +131,12 @@ where
                                 message: format!("Failed to decode on downstream port: {e}"),
                                 device: frost_signer.device_id(),
                             });
+                            sends_upstream.push(DeviceSendMessage::Core(
+                                DeviceToCoordindatorMessage {
+                                    from: frost_signer.device_id(),
+                                    body: DeviceToCoordinatorBody::DownstreamDisconnect,
```
LLFourn, line 137: Just noting that this message should be in `frostsnap_comms` not in core. What devices are "connected" is not a concept in `core`. I'm working on this issue now.

LLFourn:
Replaced by #90

PR #90, opened 2023-09-20
Device disconnected message

LLFourn:
Builds on #80.

The purpose of the PR is to notify the coordinator when a downstream device physically disconnects.

So far I've done two refactorings before this:

- All messages sent from a device include a `from` field
- Simplify message sending by improving how we use the `Direction` type parameter to the messages

Depending on how things go it might be worth reviewing those two changes and merging them first.


musdom:
Tested frostsnapp. Mistakenly tested branch devices with master frostsnapp and found bug.

LLFourn:
> Tested frostsnapp. Mistakenly tested branch devices with master frostsnapp and found bug.

So I looked for how I could have changed the protocol in this branch. I couldn't find anything so I guess we'll look the other way on this. As long as it all works on this PR :P

device/src/esp32_run.rs
```diff
@@ -159,10 +143,7 @@ where
                 if downstream_serial.find_and_remove_magic_bytes() {
                     downstream_active = true;
                     ui.set_downstream_connection_state(true);
-                    sends_upstream.push(DeviceSendMessage::Debug {
-                        message: "Device read magic bytes from another device!".to_string(),
-                        device: frost_signer.clone().device_id(),
-                    });
+                    sends_upstream.debug("Device read magic bytes from another device!");
```
nickfarrow, line 146: Could we maybe change .`debug` to `.send_debug`?

nickfarrow:
Nice I like what has been done with ` Direction`.

Everything looks good to me so far and the fix for duplicate device showing in app works.

Where in the coordinator code will we model connections so we can disconnect multiple downstream devices from a single `Disconnected` message?

device/src/esp32_run.rs
LLFourn: I don't mind either way so I did it.

LLFourn:
Somewhere is `usb_serial_manager.rs`. I think the harder problem is informing the coordinator about which device is connected to which.

LLFourn:
I've done the device disconnected message as well. It's working well on linux app but weird stuff is happening on android.

musdom:
Works on both android and linux. Is the menu bar gone?

LLFourn:
> Works on both android and linux. Is the menu bar gone?

Yes it was not doing anything so I thought to simplify. Want it back?

musdom:
Ok no need, all good

device/src/esp32_run.rs
```diff
@@ -61,15 +66,15 @@ where
                         signer: frost_signer.clone(),
                     })
                     .unwrap();
-                ui.misc_print("New secret generated and saved");
+                ui.dispaly_debug("New secret generated and saved");
```
nickfarrow, line 69: typo
dispaly_debug -> display_debug

frostsnap_coordinator/src/usb_serial_manager.rs
```diff
@@ -273,6 +273,24 @@ impl UsbSerialManager {
...
+                    DeviceSendBody::DisconnectDownstream => {
+                        if let Some(device_list) = self.reverse_device_ports.get_mut(&serial_number)
+                        {
+                            if let Some((i, _)) = device_list
+                                .iter()
+                                .enumerate()
+                                .find(|(_, device_id)| **device_id == message.from)
+                            {
+                                let index_of_disconnection = i + 1;
+                                while device_list.len() > index_of_disconnection {
```
nickfarrow, line 285: Nice LGTM. Confirmed working on multiple ports

nickfarrow (approved):

Issue #13, opened 2023-03-16
Specific error messages in frostsnap_core


nickfarrow:

LLFourn:
We decided against this see convo in #91.

PR #91, opened 2023-09-22
Typed error variants for signer and coordinator

nickfarrow:
Implements https://github.com/frostsnap/frostsnap/issues/13

There is a little bit more code but I think it makes sense to separate coordinator and signer errors.

Types will allow for any appropriate error handling on device, as well as more resilient testing (in place of matching on generic errors or error string content).

I have left the `MessageKind` error as is.

### DIFF
--- a/frostsnap_core/src/lib.rs
+++ b/frostsnap_core/src/lib.rs
@@ -29,7 +29,7 @@ use alloc::{
 use rand_chacha::ChaCha20Rng;
 use rand_core::RngCore;
 use schnorr_fun::{
-    frost::{self, generate_scalar_poly, FrostKey, SignSession},
+    frost::{self, generate_scalar_poly, FinishKeyGenError, FrostKey, NewKeyGenError, SignSession},
     fun::{derive_nonce_rng, hex, marker::*, KeyPair, Point, Scalar, Tag},
     musig::{Nonce, NonceKeyPair},
     nonce, Message,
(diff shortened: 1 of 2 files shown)


LLFourn:
> Types will allow for any appropriate error handling on device, as well as more resilient testing (in place of matching on generic errors or error string content).

what's inappropriate with current error handling?

LLFourn:
Ok I see the motivation probably comes from wanting to check in tests that you got the right error variant over in https://github.com/frostsnap/frostsnap/pull/68/files

I'm not sure this is a good idea. I think we can cope with "stringly" typed error variants -- it's fine to match against the strings in tests. The key thing is that I don't think there is a motivation to behave differently for a particular error case in business logic. Stringly typed is less friction until there's an case where we need to match on the error type.

nickfarrow:
> The key thing is that I don't think there is a motivation to behave differently for a particular error case in business logic.

I think i started to realise this after making this branch and have jumped the gun on this. 

I agree we don't really need this matching at this point in time, but thought particular error handling would exist in the future. E.g. a `RanOutOfNonces` error could prompt generating new nonces dialogue.

One small benefit is ease of seeing the scope of possible errors (we can aim to make a test for each), the friction could lead to increased thought when adding new error variants.

LLFourn:
I agree with the above. It's a bit too early for this change as the benefits to it are delivered later in the project and the costs are paid from now onwards. Closing for now. Can revisit this when we can see more benefit from doing it.

PR #92, opened 2023-09-23
One button hold to confirm UI

musdom:
One button interface requiring user to long press to confirm pending action required on device. This is meant to showcase a capacitive touch module but should also work with the current GPIO9 button until we get our next batch of devboards.

### DIFF
--- a/device/src/bin/frostypede.rs
+++ b/device/src/bin/frostypede.rs
@@ -1,5 +1,5 @@
 // Frostsnap custom PCB rev 1.1
-// GPIO8 Downstream detection
+// GPIO13 Downstream detection
 // GPIO5 Left button
 // GPIO9 Right button
 // GPIO0 RGB LED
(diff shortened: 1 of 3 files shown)


musdom:
Will rebase on #90 so the tests pass

musdom:
Changed SplashState to AnimationState and let new instances specify duration.

device/src/bin/frostypede.rs
```diff
@@ -445,17 +451,34 @@ where
...
-                self.changes = true;
+                match self.confirm_state.poll() {
+                    AnimationProgress::Progress(progress) => {
+                        self.led
+                            .write(brightness(
+                                [RGB::new(0, (128.0 * progress) as u8, 0)].iter().cloned(),
+                                30,
+                            ))
+                            .unwrap();
+                        self.display.confirm_bar(progress).unwrap();
```
LLFourn, line 462: An alternative to doing this here is to update the state of `prompt` by adding some `progress` field to it and letting render handle dong the progress bar. This would be more inline with the design but this is also more efficient so I'm not really bothered.

LLFourn (approved):
Tested. Works well. Thanks.

PR #93, opened 2023-10-05
App Keygen

nickfarrow:
## Coordinator Outbox

Current approach for handing coordinator outbox is to add a Arc<Mutex<VecDeque<..>> to the FfiCoordinator `pending_for_outbox`.

* In the message handling loop we read messages into recv_device_message.
* CoordinatorSend responses extend `pending_for_outbox`
* At the end of the loop for this thread we drain all the messages, sending them to the appropriate recipient (ToDevice)

Outside of this loop, we can also add messages to`pending_for_outbox` with a function like `coordinator.do_keygen()`. Which the other message thread above will process.

edit: there's also this branch [app-sign](https://github.com/frostsnap/frostsnap/tree/app-sign) as a v early demo of single device signing


musdom:
Working in Android but not in Linux

nickfarrow:
Workflow:

* The `Generate Key` button is hidden until at least one device is named.

* The threshold slider is hidden until there are at least two named devices.

* After finishing keygen, the app will alert prompt the user to confirm what they see matches the devices, the user response acks/nacks keygen on the coordinator.

* If acked, the user is then taken to a screen which shows the public key (will later become a home-screen for the wallet).

Not addressed in this PR:

* MaterialApp routes (to get to a permanent `KeyDisplayPage` page), you'll see you can still click the back arrow.
* Saving of coordinator state to flash.
* The workflow of "Yes/No" between the devices and coordinator still does not interact (e.g. Clicking `No` on coordinator does not send `Cancel` to devices, or otherwise)

I'd like to squash the keygen related commits before merging (messy, forgot to add files)

LLFourn (changes requested):
I think we should expand the scope to making the whole keygen bit coherent which would make this PR add far more value.

Bugs:
- Landscape mode not working
- slider doesn't let you choose every number

Things needed to make this PR coherent:

- It shouldn't be possible to go "back" from public key display which bugs the UI. Instead you should have a delete key button which takes you to the device list keygen screen.
- Cancelling in the UI should cancel the device keygen
- Rather than a spinny thing by itself while the devices are doing things have a list of devices we're waiting for (by name) and tick them off once you are no longer waiting for them.
- We need a keygen ack message to coordinator from devices. And then a final keygen confirm from coordinator to devices (sent when every device has ack'd). This would force us to ack the key on each device one by one in the UI which needs to happen for this workflow to make sense.

frostsnapp/native/src/coordinator.rs
```diff
@@ -31,55 +36,88 @@ impl FfiCoordinator {
...
+
+                        manager_thread
+                            .lock()
+                            .unwrap()
+                            .queue_in_port_outbox(send_message);
+                    }
+                    CoordinatorSend::ToUser(msg) => match msg {
+                        CoordinatorToUserMessage::Signed { .. } => {}
+                        CoordinatorToUserMessage::CheckKeyGen { .. } => {
+                            // Don't want to ack this until button press!!!
```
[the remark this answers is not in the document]
nickfarrow: I had this previously but in did it, since it was superfluous to the keygen progression UI.

This is because the keygen thread polling for a `FrostKey` state would finish at the same time as `bool: coordinator.is_awaiting_keygen` became true. So it was doubling up and had the feeling that I was adding too many state variables to the coordinator api. But I'll add it back and try with other UX changes

frostsnapp/native/src/api.rs
```diff
@@ -215,3 +218,12 @@ pub fn finish_naming(coordinator: RustOpaque<FfiCoordinator>, id: DeviceId, name
...
     coordinator.send_cancel(id);
 }
+
+pub fn generate_new_key(coordinator: RustOpaque<FfiCoordinator>, threshold: usize) -> String {
+    let coordinator_frostkey = coordinator.generate_new_key(threshold);
+    format!("{}", coordinator_frostkey.frost_key().public_key())
+}
+
+pub fn keygen_ack(coordinator: RustOpaque<FfiCoordinator>, ack: bool) {
+    coordinator.keygen_ack(ack)
```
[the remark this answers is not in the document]
nickfarrow: We also need to reset the coordinator out of state `Keygen` by passing `false` when we click cancel

frostsnapp/native/src/coordinator.rs
```diff
@@ -31,55 +36,88 @@ impl FfiCoordinator {
...
-                            device_changes
-                                .into_iter()
-                                .map(crate::api::DeviceChange::from)
-                                .collect(),
+                new_messages
+            };
+
+            for (from, message) in new_messages {
+                match coordinator_loop
+                    .lock()
```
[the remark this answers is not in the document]
nickfarrow: Changing everywhere except for the `manager`, holding it for longer than needed causes message reading/writing lag

frostsnapp/native/src/coordinator.rs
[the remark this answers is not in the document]
LLFourn: I think this all works nicely now.

LLFourn:
I've made major additions to this PR. We now land on "key list" page (which for now only ever has one key). You can start generating a new key from there. This brings us to the threshold selection (in the future probably also device selection page). From there you start the actual work of generating and end up having to confirm on each device and the coordinator.

I didn't end up doing the final confirmation message from coordinator to devices since the PR was big enough and I think you get the idea.

I've figured out a bit about how widget positioning, sizing and how the `Navigator` and learned a lot about designing flutter things.

One thing I also noticed is that the app takes a lot of CPU on linux. This seems to be because of our device polling loop. I'm surprised at this but haven't tried to debug it.

LLFourn (approved):

musdom:
tested ACK

musdom:
> One thing I also noticed is that the app takes a lot of CPU on linux. This seems to be because of our device polling loop. I'm surprised at this but haven't tried to debug it.

I'm seeing 20% CPU usage

frostsnap_core/src/key_id.rs
```diff
@@ -0,0 +1,50 @@
...
+                .into(),
+        )
+    }
+}
+
+pub trait FrostKeyExt {
+    fn key_id(&self) -> KeyId;
+}
+
+impl FrostKeyExt for FrostKey<Normal> {
```
nickfarrow, line 45: I can't yet see what this trait is for?

frostsnap_core/src/lib.rs
```diff
@@ -667,19 +682,30 @@ impl FrostSigner {
         *self = Self::new(self.keypair.clone())
     }
 
-    pub fn cancel_action(&mut self) {
-        match &self.state {
-            SignerState::KeyGen { .. } => {
+    #[must_use]
```
nickfarrow, line 685: 👍

frostsnapp/lib/device_listener.dart
```diff
@@ -0,0 +1 @@
+
```
nickfarrow, line 1: unused?

frostsnapp/native/src/coordinator.rs
```diff
@@ -108,25 +125,14 @@ impl FfiCoordinator {
                 }
             }
             // to give time for the other threads to get a lock
-            std::thread::sleep(Duration::from_millis(10));
+            std::thread::sleep(Duration::from_millis(100));
```
nickfarrow, line 136: Trying to reduce CPU on desktop?

nickfarrow:
Reviewed rust code changes and looked over most of the dart, only minor questions above. Thanks for fixing the orientation and getting rid of those buggy Extends(..).

Tested on linux and it's awesome! Only suggestions (can wait for a later PR) is to bring back the old device list style(?), and make the "Press if all devices show..." a separate button appear to underneath the list.

Did we decide that we would always show the threshold slider on keygen page? Even when there are no devices or for 1-of-1?

frostsnap_core/src/key_id.rs
LLFourn: It's an extension trait for adding a method to get a key_id -- extension traits are just a way of adding a method to a foreign type. The traits themselves have no meaning they are just a syntax manipulation tool (so I can use method call syntax rather than function call).

frostsnapp/native/src/coordinator.rs
LLFourn: Yes. Every 100ms seems to be fast enough for ui now. Will take some work to investigate why so much CPU is being used.

frostsnapp/lib/device_listener.dart
LLFourn: Yes

frostsnap_core/notes.md
```diff
@@ -0,0 +1,22 @@
...
+
+Need to remember used vs unused. If we share them around, so long as one device refuses to reuse a nonce then it will be secure. Don't need to choose the devices upfront.
+Coordinator chooses a nonce that hasnt been used before.
+Goes around to each device and recreate a device
+
+Protect joint key instead of their own share from nonce reuse.
+
+## Review
+
+Signing parties are not an option
```
LLFourn, line 22: @nickfarrow was this intended to be committed? I'm guessing not.

> Need to remember used vs unused. If we share them around, so long as one device refuses to reuse a nonce then it will be secure. Don't need to choose the devices upfront.

FWIW this bit doesn't seem to be correct. All devices must refuse to use nonces they've already used. No other device can protect them.

LLFourn:
I just added the device list to the name confirmation screen. The idea behind it is to always show all the devices when you are asking the user to do something with one of them.


> Reviewed rust code changes and looked over most of the dart, only minor questions above. Thanks for fixing the orientation and getting rid of those buggy Extends(..).

What's `Extends`?

> 
> Tested on linux and it's awesome! Only suggestions (can wait for a later PR) is to bring back the old device list style(?), and make the "Press if all devices show..." a separate button appear to underneath the list.

I put the color back to how it was.
 
> Did we decide that we would always show the threshold slider on keygen page? Even when there are no devices or for 1-of-1?

There are no decisions being made here. At the moment we are just trying to learn. I have made some interesting UI elements that maybe aren't strictly necessary but again this is just me trying to learn how this whole system works. The most important thing is to get every workflow step that needs to be there present in the UI and to make the workflows robust. There's still a few things missing here even from keygen and naming.

So I guess try to avoid making UI tweaks. Do radical changes. Do experiments that we can learn something from. I predict almost nothing we do right now will exist in 6 months time so there is really no useful gains to be had from small improvements yet.

Of course if something is really bothering you because and detracts from a demo feel etc just go ahead and change it in another PR (no need to ask permission or discuss).

PR #97, opened 2023-11-21
Fix master and upgrade esp32-hal

LLFourn:
I accidentally merged a PR that broke master. Hopefully we can't do this anymore after tweaking github. I fixed that problem by refactoring our testing approach in 898c86ea32d14357a553fa893426bb1674a060e8. We now test a lot more as well and they are much easier to develop and read.

I then included #95 on top of that and then on top of that I made a fix in 3758c253c338413e3dee3e4fe953e5e24eb8aa1f

@musdom can you confirm how you determined the heap size in b3ef7d68f6a0153f00d53e8da2fc5a1f7a864296?


musdom:
We used to allocate 320kB, but this new allocator only lets us use max 314kB.

LLFourn:
We may want to keep the heap small just so we notice when it fills up come to think of it?

nickfarrow:
I think `run_until` will be very useful in the future.

Nice work fixing the hacky XOnlyPublicKey conversion.

Esp-hal changes are great too, looks much cleaner.

I left some small comments while going through but looks like I accidentally added them to individual commits and aren't showing here. I'll repost them.

frostsnap_core/src/message.rs
```diff
@@ -263,3 +278,39 @@ pub struct SignItem {
...
+            .to_scalar();
+            xpub.key().clone().into_xonly_with_tweak(
+                Scalar::<Public, _>::from_bytes_mod_order(tweak.to_be_bytes())
+                    .non_zero()
+                    .expect("computationally unreachable"),
+            )
+        } else {
+            xpub.key().clone().into_xonly()
+        }
+    }
```
nickfarrow, line 301: I realize it was like this before, but it's hard to follow the reasoning for these steps.

We take our root Frostsnap key and derive an extended public key by adding tweaks according to the BIP32 path.

Then, if the key is tweaked, we get the TapTweakHash from this extended public key. A "cumulative tweak" from the derived key?

Then using this tweak, we add take our derived extended public key again and finally convert it to an xonly key under this tweak.

The code makes it seem like we are adding the tweak in multiple times.

Should self.tap_tweak always be true if self.bip32_path is not empty?

device/src/io.rs
```diff
@@ -177,18 +173,28 @@ impl<'a, U> SerialIo<'a, U> {
...
-                .write_bytes(words)
-                .map_err(|_| SerialInterfaceError::JtagError),
-            SerialIo::Uart(uart) => uart
-                .write_bytes(words)
-                .map_err(SerialInterfaceError::UartWriteError),
+                .write_bytes(bytes)
+                .map_err(|_| SerialInterfaceError::JtagError)?,
+            SerialIo::Uart(uart) => {
+                for byte in bytes {
+                    while let Err(e) = uart.write(*byte) {
```
nickfarrow, line 186: > you can end up writing less than the bytes and won't notice it

Wouldn't this have previously returned an error coming from `uart.write_bytes`?

The improvement I see here is not erroring on a `WouldBlock` and instead trying again.

musdom:
> We may want to keep the heap small just so we notice when it fills up come to think of it?

I tried leaving it small and it panics when doing a 3of4 keygen

frostsnap_core/src/message.rs
LLFourn: > I realize it was like this before, but it's hard to follow the reasoning for these steps.
> 
> We take our root Frostsnap key and derive an extended public key by adding tweaks according to the BIP32 path.
> 
> Then, if the key is tweaked, we get the TapTweakHash from this extended public key. A "cumulative tweak" from the derived key?

There is no cumulative tweak. The TapTweakHash is not from the extended public key. Just from the xonly key. I've pushed a commit to make this clearer as they were being conflated before: 45e2dc816abd5482a63d65fd21face69240ac53d

> 
> Then using this tweak, we add take our derived extended public key again and finally convert it to an xonly key under this tweak.
> 
> The code makes it seem like we are adding the tweak in multiple times.

we are.

> 
> Should `self.tap_tweak` always be true if `self.bip32_path` is not empty?

They're orthogonal. You can do a BIP32 derivation for a nostr key for example.

device/src/io.rs
LLFourn: My guess is this is what the previous implementation of `write_bytes` did: Keep writing and trying again if you get `WouldBlock` -- now it just stops on `WouldBlock` and returns the number of bytes written.

frostsnap_core/src/message.rs
nickfarrow: Ah got it, thanks!

nickfarrow (approved):

nickfarrow:
This patch can be used to debug memory issues onscreen, highest i *saw* was ~14000b in a 3-of-3
<details>
<summary>patch</summary>

```
diff --git a/device/src/bin/frostypede.rs b/device/src/bin/frostypede.rs
index 1e8f14f..39d6f10 100644
--- a/device/src/bin/frostypede.rs
+++ b/device/src/bin/frostypede.rs
@@ -391,6 +391,9 @@ where
                 ConnectionState::Established => Rgb565::GREEN,
             });

+        self.display
+            .set_mem_debug(ALLOCATOR.used(), ALLOCATOR.free());
+
         self.display.flush().unwrap();
     }
 }
diff --git a/device/src/st7735.rs b/device/src/st7735.rs
index 2eebba9..9881ecd 100644
--- a/device/src/st7735.rs
+++ b/device/src/st7735.rs
@@ -310,4 +310,26 @@ where

         Ok(())
     }
+
+    pub fn set_mem_debug(&mut self, used: usize, free: usize) {
+        Rectangle::new(Point::new(80, 60), Size::new(80, 20))
+            .into_styled(
+                PrimitiveStyleBuilder::new()
+                    .fill_color(Rgb565::BLACK)
+                    .build(),
+            )
+            .draw(&mut self.framebuf)
+            .unwrap();
+
+        TextBox::with_textbox_style(
+            &format!("{}/{}", used, free),
+            Rectangle::new(Point::new(80, 60), Size::new(80, 20)),
+            MonoTextStyle::new(&FONT_7X14, Rgb565::GREEN),
+            TextBoxStyleBuilder::new()
+                .alignment(HorizontalAlignment::Right)
+                .build(),
+        )
+        .draw(&mut self.framebuf)
+        .unwrap();
+    }
 }

```

</details>

Issue #98, opened 2023-11-22
Display memory usage on screen during development with feature flag


LLFourn:
We should do this on the next boards where we're likely to have more screen real estate.

PR #99, opened 2023-11-22
Onscreen memory feature flag

nickfarrow:
Implements https://github.com/frostsnap/frostsnap/issues/98


nickfarrow:
As Lloyd has pointed out, for some reason `used + free` varies and `!= size`

PR #100, opened 2023-11-22
[app] Model device list in rust

LLFourn:
Rather than do it in dart because it is easier to model things in rust. We still translate the device `UsbSerialManager` events into actions on a single "device list" we just do it in rust since.


LLFourn:
Attached another commit to this one to fix a bug with device naming.

nickfarrow:
These two commits were included in https://github.com/frostsnap/frostsnap/pull/101

PR #101, opened 2023-11-23
App signing

LLFourn:


coordinator-cli/src/signer.rs
```diff
@@ -147,18 +132,13 @@ impl<'a, 'b> Signer<'a, 'b> {
             // because often a big bunch of devices will register at similar times if they are daisy
             // chained together.
             loop {
+                std::thread::sleep(Duration::from_millis(100));
```
nickfarrow, line 135: What's the motivation behind this sleep?

frostsnap_coordinator/src/signing.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::collections::HashSet;
+
+pub struct SigningDispatcher {
+    request: SignRequest,
+    targets: HashSet<DeviceId>,
+    got_signatures: HashSet<DeviceId>,
+    need_to_send_to: HashSet<DeviceId>,
+}
+
+impl SigningDispatcher {
```
nickfarrow, line 17: LGTM

nickfarrow:
Looked over recent changes, going to test now.

One question I have is whether this `state_counter` pattern is going to come up elsewhere? E.g. could we have similar race conditions for a `wallet transaction history` display syncing with the blockchain?

nickfarrow (approved):

coordinator-cli/src/signer.rs
LLFourn: This loop hoses CPU otherwise. It should have a comment.

LLFourn:
> Looked over recent changes, going to test now.
> 
> One question I have is whether this `state_counter` pattern is going to come up elsewhere? E.g. could we have similar race conditions for a `wallet transaction history` display syncing with the blockchain?

I think in theory you can have these race conditions when you start streams in this way but they are just very unlikely. Doing things this way seems like good practice as it helps you recover from panics that occur during processing the event (this was the actual problem) by rebuilding the list.

PR #104, opened 2023-12-27
Small share indicies

nickfarrow:
Currently we use a hash of the DeviceId to determine the scalar party index.
But we want to use small simple indices (`1,2,3,...`) to achieve significantly smaller backups (1 bech32m character vs 52 characters).

This PR is WIP exploration of having the coordinators decide these party indexes during keygen.

Signers will need to remember their share index (inside `FrostsnapKey`).

Coordinators will want to remember `BTreeMap<DeviceId, ShareIndex>`

### DIFF
--- a/frostsnap_coordinator/src/signing.rs
+++ b/frostsnap_coordinator/src/signing.rs
@@ -32,9 +32,10 @@ impl SigningDispatcher {
             .iter()
             .enumerate()
             .find_map(|(i, m)| match m {
-                CoordinatorSend::ToDevice(CoordinatorToDeviceMessage::RequestSign(request)) => {
-                    Some((i, request.clone()))
-                }
+                CoordinatorSend::ToDevice {
+                    message: CoordinatorToDeviceMessage::RequestSign(request),
+                    ..
+                } => Some((i, request.clone())),
                 _ => None,
             })
             .expect("must have a sign request");
@@ -86,10 +87,8 @@ impl SigningDispatcher {
 
     pub fn resend_sign_request(&mut self) -> Option<CoordinatorSendMessage> {
         if !self.need_to_send_to.is_empty() {
-            let target_destinations = Destination::from(self.need_to_send_to.clone());
-            self.need_to_send_to = BTreeSet::new();
             return Some(CoordinatorSendMessage {
-                target_destinations,
+                target_destinations: Destination::from(core::mem::take(&mut self.need_to_send_to)),
                 message_body: CoordinatorSendBody::Core(CoordinatorToDeviceMessage::RequestSign(
                     self.request.clone(),
                 )),
--- a/frostsnap_coordinator/src/signing.rs
+++ b/frostsnap_coordinator/src/signing.rs
@@ -6,25 +6,28 @@ use frostsnap_core::{
     },
     DeviceId,
 };
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 
 /// Keeps track of when
 #[derive(Debug)]
 pub struct SigningDispatcher {
-    need_to_send_to: HashSet<DeviceId>,
+    need_to_send_to: BTreeSet<DeviceId>,
     signing_state_changed: bool,
     // FIXME: make accessors
     pub request: SignRequest,
     pub finished_signatures: Option<Vec<EncodedSignature>>,
-    pub targets: HashSet<DeviceId>,
-    pub got_signatures: HashSet<DeviceId>,
+    pub targets: BTreeSet<DeviceId>,
+    pub got_signatures: BTreeSet<DeviceId>,
 }
 
 impl SigningDispatcher {
     /// Takes in the messages from `start_sign` and extracts the signing request to handle separately.
     ///
     /// We need to do this because we want to only send out the message to the devices that are connected.
-    pub fn from_filter_out_start_sign(start_sign_messages: &mut Vec<CoordinatorSend>) -> Self {
+    pub fn from_filter_out_start_sign(
+        start_sign_messages: &mut Vec<CoordinatorSend>,
+        targets: BTreeSet<DeviceId>,
+    ) -> Self {
         let (i, request) = start_sign_messages
             .iter()
             .enumerate()
@@ -37,11 +40,10 @@ impl SigningDispatcher {
             .expect("must have a sign request");
 
         start_sign_messages.remove(i);
-        Self::new_from_request(request)
+        Self::new_from_request(request, targets)
     }
 
-    pub fn new_from_request(request: SignRequest) -> Self {
-        let targets = request.devices().collect::<HashSet<_>>();
+    pub fn new_from_request(request: SignRequest, targets: BTreeSet<DeviceId>) -> Self {
         Self {
             request,
             targets,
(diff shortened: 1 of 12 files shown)


frostsnap_core/src/lib.rs
```diff
@@ -103,90 +140,92 @@ impl FrostCoordinator {
                                 .map(|(device_id, shares)| Some((device_id, shares?)))
                                 .collect::<Option<BTreeMap<_, _>>>();
 
-                            match all_responded {
-                                Some(responses) => {
-                                    let point_polys = responses
+                            if let Some(responses) = all_responded {
```
nickfarrow, line 143: clippy was complaining about this empty match

nickfarrow:
Coordinators store the party indexes in their `CoordinatorFrostKeyState`.

Signers store party indexes in their `SignerState::FrostKey`.

For signers they store this `BTreeMap<DeviceId, ShareIndex>`, and they know their own DeviceId (Keypair stored **alongside** the signer state) so can lookup their own index.
I think this design be reasonably extendable to recovery (new device id points to same share index) and controlling multiple shares (vec of indexes)

LLFourn (changes requested):

frostsnap_core/src/lib.rs
```diff
@@ -48,6 +50,40 @@ pub const NONCE_BATCH_SIZE: usize = 10;
...
+        Self {
+            index: Scalar::from_non_zero_u32(index).public(),
+        }
+    }
+
+    pub fn new_from_scalar(index: Scalar<Public>) -> Self {
+        Self { index }
+    }
+
+    pub fn index(&self) -> Scalar<Public> {
```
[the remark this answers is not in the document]
nickfarrow: Yeah unnecessary. Not sure what I was thinking, maybe justifying it from api perspective. Removed

We also have `schnorr_fun::frost::PartyIndex`, and being private only really helps with code legibility.

frostsnapp/lib/bridge_generated.web.dart
```diff
@@ -1,5 +1,5 @@
 // AUTO GENERATED FILE, DO NOT EDIT.
-// Generated by `flutter_rust_bridge`@ 1.82.6.
+// Generated by `flutter_rust_bridge`@ 1.82.1.
```
nickfarrow, line 2: reminder to self to upgrade

frostsnap_core/src/coordinator.rs
```diff
@@ -78,11 +78,13 @@ impl SigningSessionState {
 #[derive(Clone, Debug)]
 pub enum KeyGenState {
     WaitingForResponses {
+        devices: BTreeMap<DeviceId, Scalar<Public, NonZero>>,
```
LLFourn, line 81: Should be named, `device_to_share_index`.

frostsnap_core/src/device.rs
```diff
@@ -455,17 +472,19 @@ pub enum SignerState {
...
-        devices: BTreeSet<DeviceId>,
+        devices: BTreeMap<DeviceId, Scalar<Public, NonZero>>,
         threshold: usize,
         aux_rand: [u8; 32],
     },
     AwaitingSignAck {
         key: FrostsnapKey,
+        devices: BTreeMap<DeviceId, Scalar<Public, NonZero>>,
         sign_task: SignTask,
         nonces: BTreeMap<DeviceId, (Vec<Nonce>, usize, usize)>,
```
LLFourn, line 483: Do we need to know the device id's of the other devices after key generation? I think I just need to know my index.
We shouldn't have the `DeviceId` in the nonces either, it should just be the signer index.

Same is true for the `FrostKey` state. Just need to know my index.

frostsnapp/lib/bridge_generated.io.dart
```diff
@@ -669,9 +669,9 @@ class NativeWire implements FlutterRustBridgeWireBase {
   }
 
   late final _wire_get_devicePtr = _lookup<
-          ffi
-          .NativeFunction<WireSyncReturn Function(ffi.Pointer<wire_DeviceId>)>>(
-      'wire_get_device');
+      ffi.NativeFunction<
+          WireSyncReturn Function(
+              ffi.Pointer<wire_DeviceId>)>>('wire_get_device');
```
LLFourn, line 674: can you trash these bridge_definitions changes?

frostsnap_coordinator/src/signing.rs
```diff
@@ -84,8 +86,10 @@ impl SigningDispatcher {
 
     pub fn resend_sign_request(&mut self) -> Option<CoordinatorSendMessage> {
         if !self.need_to_send_to.is_empty() {
+            let target_destinations = Destination::from(self.need_to_send_to.clone());
+            self.need_to_send_to = BTreeSet::new();
             return Some(CoordinatorSendMessage {
-                target_destinations: Destination::from(self.need_to_send_to.drain()),
```
nickfarrow, line 88: I have swapped the the HashSet types to BTreeSet to better align with `frostsnap_core`. This drain was the only significant HashSet feature.

frostsnap_core/src/device.rs
nickfarrow: We're using sets of `DeviceId` in each `CoordinatorToDevice` variant to get `default_destinations()`, so i added `targets` on `CoordinatorToDevice::SignRequest`.

frostsnap_core/src/message.rs
```diff
@@ -52,16 +53,17 @@ pub enum CoordinatorToDeviceMessage {
 pub struct SignRequest {
     // TODO: explain these `usize` and create a nicely documented struct which explains the
     // mechanism
-    pub nonces: BTreeMap<DeviceId, (Vec<Nonce>, usize, usize)>,
+    pub nonces: BTreeMap<Scalar<Public, NonZero>, (Vec<Nonce>, usize, usize)>,
     pub sign_task: SignTask,
+    pub targets: BTreeSet<DeviceId>,
```
LLFourn, line 58: I don't think it makes sense to put the target destinations here. The devices simply don't need to know this so why send them over the wire here. The devices figure out if it's meant for them in the `frostsnap_comms::CoordinatorSendMessage.target_destinations` field.

I do see the problem you're facing. We need to make who the targets are explicit because of this change. The code that generates the message needs to set the targets. I think we can do:

```rust
#[derive(Clone, Debug)]
pub enum CoordinatorSend {
    ToDevice { message: CoordinatorToDeviceMessage, destinations: BTreeSet<DeviceId>},
    ToUser(CoordinatorToUserMessage),
    ToStorage(CoordinatorToStorageMessage),
}
```

so that `start_sign` or whatever method that returns a CoordinatorSend can explicitly set the destinations.

frostsnap_coordinator/src/signing.rs
```diff
@@ -6,25 +6,28 @@ use frostsnap_core::{
...
 }
 
 impl SigningDispatcher {
     /// Takes in the messages from `start_sign` and extracts the signing request to handle separately.
     ///
     /// We need to do this because we want to only send out the message to the devices that are connected.
-    pub fn from_filter_out_start_sign(start_sign_messages: &mut Vec<CoordinatorSend>) -> Self {
+    pub fn from_filter_out_start_sign(
+        start_sign_messages: &mut Vec<CoordinatorSend>,
+        targets: BTreeSet<DeviceId>,
```
LLFourn, line 29: So with my suggestion above I'm saying that this wouldn't be passed in here -- it'll be in the `ToDevice` variant below.

frostsnap_coordinator/src/signing.rs
LLFourn: Changing to BTreeSet is fine. I think you want to use `core::mem::take` here rather than `clone` and then set it to `new`.

LLFourn (approved):
LGTM

PR #106, opened 2024-01-19
Persist data in the phone app

LLFourn:
This uses my new experimental embedded db [`llsdb`](https://github.com/LLFourn/llsdb).

The usage can be seen in `native/src/coordinator.rs` and in `persist_core.rs`. The API will change quite a bit as we develop I imagine but the core principles are there. With llsdb you have typed structs that behave almost as if they were using in memory storage but they're actually persisting things to disk. 

I also split up frostsnap_core into `coordinator.rs` and `device.rs`. This will make the changes to `frostsnap_core` harder to review. The significant changes I made here were:

1. Nonces are consumed in the device after the user has confirmed they want to sign and the signature. Doing this too early caused problems with restarting the signing session.
2. We now emit a a new storage message to the coordinator when ever a signing session actually makes progress (because you should probably save it).

On the app side, we can now restore a signing session by pressing "continue signing". To make this work I refactored the events we publish to the app through the signing stream to be the full signing state. 

Sorry for the one giant commit.


frostsnapp/native/src/coordinator.rs
```diff
@@ -31,35 +36,41 @@ pub struct FfiCoordinator {
...
-        Self {
-            coordinator,
+        let (core_persist, device_names, coordinator) = db
+            .execute(|tx| {
+                let persist = PersistCore::new(tx)?;
+                let (handle, api) = tx.store_and_take_index(persist);
+                let coordinator = api.core_coordinator()?;
+                let device_names_list = tx.take_list("device_names")?;
+                let device_names_btree = llsdb::index::BTreeMap::new(device_names_list, &tx)?;
+                let device_names_btree_handle = tx.store_index(device_names_btree);
```
nickfarrow, line 57: Curious as to why we add the `device_names` entry here and not in `PersistCore`, is the distinction that we're more willing to forget it?

frostsnapp/native/src/coordinator.rs
```diff
@@ -285,36 +342,99 @@ impl FfiCoordinator {
...
             .unwrap()
-            .replace(SigningState { stream, session })
-        {
-            old_signing_state.stream.close();
+            .execute(|tx| tx.take_index(self.core_persist).persisted_signing())?;
+
+        let signing_session_state =
+            signing_session.ok_or(anyhow!("no signing session to restore"))?;
+        let mut coordinator = self.coordinator.lock().unwrap();
+        coordinator.restore_sign_session(signing_session_state.clone());
```
nickfarrow, line 411: Nice!

nickfarrow:
## frostsnap_core changes

> Nonces are consumed in the device after the user has confirmed they want to sign and the signature. Doing this too early caused problems with restarting the signing session. 

I'm now confused by what we were doing originally.. We were essentially expending the nonce upon receiving a `SignRequest` by updating `self.nonce_counter`, before we were emitting `ToStorage::ExpendNonce` (sent after `sign_ack`). Maybe when we introduced sign_ack we didn't move the nonce expenditure.

I think you mentioned this also: the two usizes in the SignRequest feel odd. I suppose because the device isn't looking up it's own indexes from a list of generated public nonces, the coordinator is telling the device where to look. Noting that signers will learn the nonce indexes of other signers for no reason.

> We now emit a a new storage message to the coordinator when ever a signing session actually makes progress 

`CoordinatorToStorageMessage::StoreSigningState` and the functions to recover a signing session look great, will test today/tomorrow.

Thinking for future: since we can "pause" and recover a signing session the next step in this direction would be to manage different signing sessions in parallel. Handling this `frostsnap_core` level would we instead stor something like a `BTreeMap<SessionId, SigningSessionState>` in the db rather than the single `SigningSessionState`?




## llsdb

I don't have any knowledge on databases but this looks extremely impressive.
I have a rough understanding of what's going on from a brief look, tempting to look deeper. 

It was a little tricky to see what was going on at first glance with the `take_list`s, almost too magical :
```
tx.take_list("frostsnap/signing")?;
```
could be turbofished but not necessary.

Questions:
* What does lls stand for?
* Do you think we should also use `llsdb` on the device?
* Did you mention that you'd like to use this in/alongside BDK? Is there any discussion anywhere?
* Do other database libraries come with many other features not implemented here? Would these features ever be useful to us? Would they be easy to implement in `llsdb`?
    - I like the idea that we'll be able to easily extend this database for our own needs, e.g. error correction.
* How performant would this db be? (though i cant think why we'd need high performance)
* In the event that we were ever convinced to revert to another db library, would this be relatively simple or significant?
    - My experience with sqlx: the `tx` and `execute`, this looks roughly the same. but I recall lots of pain passing Connections and Transactions everywhere

frostsnapp/native/src/coordinator.rs
LLFourn: Good question. I didn't leave a comment answering it because I felt it's something that will change quickly. But since you asked, there are two reasons:

1. PersistCore stores `frostsnap_core` stuff which I felt should be separate.
2. There is a problem with doing this because rust only includes [`RefMut::map_split`](https://doc.rust-lang.org/std/cell/struct.RefMut.html#method.map_split) which only allows you to split a `RefMut` into 2! But adding a third field I'd need to split it into three to keep it in line with the design of llsdb. I found out about this limitation while trying to do this so I didn't try and fix it just now (it is fixable with `unsafe` code in llsdb I think).

The long term vision of llsdb is to just have a single struct that you put a derive on and everything happens. It's the next thing I'll work on in my off hours.

LLFourn:
> ## frostsnap_core changes
> 
> > Nonces are consumed in the device after the user has confirmed they want to sign and the signature. Doing this too early caused problems with restarting the signing session.
> 
> Somewhat confused by what we were doing originally.. Updating `self.nonce_counter` upon receiving a `SignRequest`, before we were emitting `ToStorage::ExpendNonce` (sent after `sign_ack`). Maybe when we introduced sign_ack we didn't move the nonce expenditure.

Yep right. We didn't consume the thing in the same place we emitted the message.

> 
> I think you mentioned this also: the two usizes in the SignRequest feel odd. The first usize tells the device the index to generate nonces from, rather than device indexing nonces itself from a list of public nonces. Noting that signers will learn the nonce indexes of other signers for no reason (seems fine). The second usize (`index_of_last_nonce`) can instead be calculated from the request len. Later PR
> 
> > We now emit a a new storage message to the coordinator when ever a signing session actually makes progress
> 
> `CoordinatorToStorageMessage::StoreSigningState` and the functions to recover a signing session look great, will test today/tomorrow.
> 
> Thinking for future: since we can "pause" and recover a signing session the next step in this direction would be to manage different signing sessions in parallel. Handling this `frostsnap_core` level would we instead stor something like a `BTreeMap<SessionId, SigningSessionState>` in the db rather than the single `SigningSessionState`?

Yes but my relatively recent thinking is we shouldn't do this. `frostsnap_core` is actually fine as it is. It manages protocol state. If you want to restart a particular singing session then you feed in the state of that signing session and get it to restart the protocol from where it left off. It *doesn't* manage application state i.e. it doesn't manage how many signing sessions are currently running and which ones have been canceled.

There is a concern with taking this position though: the nonce counter. Signing sessions have to be restarted in-order in this design so you can start a new one while another is running but the devices must sign them in order. I think this is fine for now and we can wait and see if this restriction actually causes an issue (I can't see how it could).

> ## llsdb
> 
> I don't have any knowledge on databases but this looks extremely impressive. After looking at some parts of the code I have a rough understanding of what's going on but would take some time to understand better.
> 
> It was a little tricky to see what was going on at first glance with the `take_list`s, almost too magical :
> 
> ```
> tx.take_list("frostsnap/signing")?;
> ```

It's retrieving a list called "frostsnap/signing" and creating it if it doesn't exist. All the lists are registered in the database.

> 
> because of the inferred typing. Could be turbofished but not necessary.
> 
> A few questions:
> 
>     * What does lls stand for?

linked lists

> 
>     * Do you think we should use `llsdb` on the device?

Probably there is more of a justification for using it on the device since well developed database that run on NOR flash don't seem to exist. Not that `llsdb` works well on NOR flash yet.

> 
>     * Did you mention that you'd like to use this in/alongside BDK? Is there any discussion anywhere?

Yes BDK would just mean creating a new list and storing BDK stuff in there.

> 
>     * Do other database libraries come with many other features not implemented here? Would these features ever be useful to us? Would they be easy to implement in `llsdb`?

The main thing I wanted with this project is to have an interface to the data that reflected how I like to think about data in terms of `Vec<_>`, `BTreeMap`, `BtreeSet` etc. I also don't care about indexing the data on disk because there is so little data it's always fine to read it in at startup in one go. So this means you don't need what most databases give you. I think you could make such an interface to any existing database or storage engine it's just that 90% of the features of the storage engine won't be used.

The best way to think about this is that by choosing `llsdb` we are choosing *not* to use a database yet. Before `llsdb` we had a file where we just dumped the serialized frostsnap core types. We read them in and wrote them out again whenever it changed. `llsdb` is a more structured way of continuing on with this minimal approach. So the principle I suggest we follow is to avoid (proper) databases for as long as possible since they enforce a certain way of thinking about things.

>       * I like the idea that we'll be able to easily extend this database for our own needs, e.g. error correction.
> 
>     * How performant would this db be? (though i cant think why we'd need high performance)

It's the worst possible startup time and memory usage with the best possible read and write speed. Since we have little data none of this matters. My main concern about llsdb is about time investment to make it work in the first place -- if there was something that I could find that would use less of my time but would work over the project I'd like to know about it but I never found anything satisfactory that'd work both on the coordinator and the device.

> 
>     * In the event that we were ever convinced to revert to another db library, would this be relatively simple or significant?

I think the main cost of moving to a database is the cost of moving to it and not really what you had there before. If llsdb has a lot of unexpected ongoing costs then that would be a good reason to dump it in favor of something better.
      
>       * My experience with sqlx: the `tx` and `execute`, this looks roughly the same. but I recall lots of pain passing Connections and Transactions into function types everywhere (maybe my user error)

Yes it is the little things that add continuous friction going forward. `llsdb` uses `bincode` and `anyhow` so you don't need to spend time here implementing traits etc.

nickfarrow:
> Not that llsdb works well on NOR flash yet.

In recent reading, interesting that ECC apparently isn't often used with NOR flash
```
The modern ECCs are not applicable to the NOR Flash memory
devices, because the efficiency of the ECC increases with
amount of the data written at once.
```
https://arxiv.org/pdf/1306.5350.pdf


> The best way to think about this is that by choosing llsdb we are choosing not to use a database yet.

I like this framing very much.

LLFourn:
> > Not that llsdb works well on NOR flash yet.
> 
> In recent reading, interesting that ECC apparently isn't often used with NOR flash
> 
> ```
> The modern ECCs are not applicable to the NOR Flash memory
> devices, because the efficiency of the ECC increases with
> amount of the data written at once.
> ```
> 
> https://arxiv.org/pdf/1306.5350.pdf

Skimming the paper I think they're talking about using ECCs in the actual hardware. This would make sense since you can write a single byte -- this would cause you to need to then rewrite a whole chunk of ECC for a single byte written! I think when you are writing an entry though it would be very helpful to use ECC or at least error detecting codes.

This PDF is really great: https://www.renesas.com/us/en/document/apn/an500nor-flash-memory-erase-operation

This is why I'd use error detecting codes around each write:

image

nickfarrow:
There is a bug on android startup where `coord.switchToHostHandlesSerial();` is called before initializing `coord`.

Because of this bug I also experienced a screen overflow with the error. The error textbox can be made scrollable:

<details>
<summary>patch (expand)</summary>

```
diff --git a/frostsnapp/lib/main.dart b/frostsnapp/lib/main.dart
index dd96c14..dec20d7 100644
--- a/frostsnapp/lib/main.dart
+++ b/frostsnapp/lib/main.dart
@@ -12,10 +12,14 @@ void main() async {
   // certain positions/sizes
   debugPaintSizeEnabled = false;
 
+  WidgetsFlutterBinding.ensureInitialized();
+
   try {
     final appDir = await getApplicationSupportDirectory();
     final dbFile = '${appDir.path}/frostsnap.db';
 
+    coord = await api.newCoordinator(dbFile: dbFile);
+
     if (Platform.isAndroid) {
       api.turnLogcatLoggingOn(level: Level.Debug);
       coord.switchToHostHandlesSerial();
@@ -23,7 +27,6 @@ void main() async {
       api.turnStderrLoggingOn(level: Level.Debug);
     }
 
-    coord = await api.newCoordinator(dbFile: dbFile);
     coord.startThread();
 
     runApp(MyApp());
@@ -96,59 +99,64 @@ class StartupErrorWidget extends StatelessWidget {
       appBar: AppBar(
         title: Text('Startup Error'),
       ),
-      body: Padding(
-        padding: EdgeInsets.all(16.0),
-        child: Center(
-          child: Column(
-            mainAxisAlignment: MainAxisAlignment.center,
-            crossAxisAlignment: CrossAxisAlignment.center,
-            children: <Widget>[
-              Text(
-                'ERROR',
-                style: TextStyle(
-                  fontSize: 24.0,
-                  fontWeight: FontWeight.bold,
-                  color: Colors.red,
-                ),
-              ),
-              SizedBox(height: 8),
-              Text(
-                'Please report this to the frostsnap team',
-                style: TextStyle(
-                  fontSize: 16.0,
-                  color: Colors.black54,
-                ),
-              ),
-              SizedBox(height: 20),
-              Container(
-                padding: EdgeInsets.all(8.0),
-                decoration: BoxDecoration(
-                  color: Colors.grey[200],
-                  borderRadius: BorderRadius.circular(4.0),
-                  border: Border.all(color: Colors.grey[400]!),
+      body: SingleChildScrollView(
+        child: Padding(
+          padding: EdgeInsets.all(16.0),
+          child: Center(
+            child: Column(
+              mainAxisAlignment: MainAxisAlignment.center,
+              crossAxisAlignment: CrossAxisAlignment.center,
+              children: <Widget>[
+                Text(
+                  'ERROR',
+                  style: TextStyle(
+                    fontSize: 24.0,
+                    fontWeight: FontWeight.bold,
+                    color: Colors.red,
+                  ),
                 ),
-                child: SelectableText(
-                  error,
+                SizedBox(height: 8),
+                Text(
+                  'Please report this to the frostsnap team',
                   style: TextStyle(
-                    fontFamily: 'Courier', // Monospaced font
-                    color: Colors.black,
+                    fontSize: 16.0,
+                    color: Colors.black54,
                   ),
                 ),
-              ),
-              SizedBox(height: 20),
-              IconButton(
-                icon: Icon(Icons.content_copy),
-                onPressed: () {
-                  Clipboard.setData(ClipboardData(text: error));
-                  ScaffoldMessenger.of(context).showSnackBar(
-                    SnackBar(
-                      content: Text('Error message copied to clipboard!'),
+                SizedBox(height: 20),
+                Container(
+                  padding: EdgeInsets.all(8.0),
+                  decoration: BoxDecoration(
+                    color: Colors.grey[200],
+                    borderRadius: BorderRadius.circular(4.0),
+                    border: Border.all(color: Colors.grey[400]!),
+                  ),
+                  child: SingleChildScrollView(
+                    scrollDirection: Axis.vertical,
+                    child: SelectableText(
+                      error,
+                      style: TextStyle(
+                        fontFamily: 'Courier', // Monospaced font
+                        color: Colors.black,
+                      ),
                     ),
-                  );
-                },
-                tooltip: 'Copy to Clipboard',
-              ),
-            ],
+                  ),
+                ),
+                SizedBox(height: 20),
+                IconButton(
+                  icon: Icon(Icons.content_copy),
+                  onPressed: () {
+                    Clipboard.setData(ClipboardData(text: error));
+                    ScaffoldMessenger.of(context).showSnackBar(
+                      SnackBar(
+                        content: Text('Error message copied to clipboard!'),
+                      ),
+                    );
+                  },
+                  tooltip: 'Copy to Clipboard',
+                ),
+              ],
+            ),
           ),
         ),
       ),
```
</details>

The saving and restoring key works great on linux, but i'm having some issues on android when signing. The app doesn't seem to want to register devices that're being plugged in, will investigate tomorrow.

LLFourn:
Thanks @nickfarrow I'll take a look at this.

LLFourn:
So I came across freezing in the phone app a lot which took a bit of refactoring to fix:

> [app] Fix freezing
> 
> App freezing was being caused by lock race over the UsbSerialManager. If
> a call from flutter tried to access it while it was already locked by
> the coordinator loop and then the coordinator loop made a serial request
> back out you'd have a deadlock.

The solution was to never ever lock the usb serial manager from flutter calls and get stuff like device names from elsewhere. Right now we get them from the global device list but this is a bit of a mess. I need to sort out who is in charge of naming devices and avoid all these redundant copies around. This is good enough for now though.

Please test!

nickfarrow:
Panic buffer looks good.

589da68f152d731061620f42ef601df0a4ab7b91 has fixed all my freezes except for one, I can reliably get the app to freeze if i plug in a device before I click "continue signing".

If I create a new key, it still shows "continue signing" on that same card afterwards. Goes away when I close the app. So maybe the clear doesnt make it through to the UI? Might be something more easily resolved when we have multiple keys & cards

nickfarrow (approved):
Tested on desktop and android

PR #108, opened 2024-01-24
[core] remove SignRequest unnecessary usize

nickfarrow:


frostsnap_core/src/device.rs
```diff
@@ -416,7 +415,7 @@ impl FrostSigner {
                 }
 
                 let replenish_nonces = self
-                    .generate_nonces(key.aux_rand, *my_replenish_index, sign_items.len())
+                    .generate_nonces(key.aux_rand, *my_nonce_index, sign_items.len())
```
LLFourn, line 418: I don't get it -- this is replenishing from the current index. But the coordinator will most likely already have these nonces. We should be giving it nonces it doesn't already have. This I suppose was the point of the extra usize?

nickfarrow: Yeah that explains the extra usize, we could instead use the size of our nonce cache to replenish nonces (latest commit). But this doesn't seem worthwhile, if this value were to change you'd have to inform all coordinators (maybe giving them nonces) to stay in sync with their cache length.

nickfarrow: Let me know if you think this update PR change is worth making.. It perhaps simplifies things for now before we make further decisions for nonce design.

LLFourn:
Closed in favor of discussing first: https://github.com/frostsnap/frostsnap/issues/112

PR #109, opened 2024-01-24
[core] non-deterministic keygen

nickfarrow:
A `Hal::Rng` `seed_rng` seeds a `ChaCha20Rng` which is passed into `recv_coordinator_message()`.

We also use this rng to seed the `aux_rand`, stored and used for nonce generation (note: we discussed we might later change this seed to some pubkey/HMAC from coordinator).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -965,6 +965,7 @@ dependencies = [
  "fugit",
  "mipidsi",
  "nb 1.1.0",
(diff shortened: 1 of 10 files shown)


LLFourn:
I pushed a commit to make it so `RngCore` is passed to `esp32_run` instead.

LLFourn (approved):
LGTM

PR #111, opened 2024-02-05
[device] VecDequeue instead of ringbuffer

nickfarrow:


device/src/io.rs
```diff
@@ -74,7 +74,7 @@ where
             return false;
         }
         let (progress, found) = frostsnap_comms::make_progress_on_magic_bytes::<D>(
-            self.ring_buffer.drain(),
+            self.ring_buffer.drain(..),
```
LLFourn, line 77: I don't think this is correct:

[image](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.drain)

We need `drain` not to drop the entire buffer. We only want it to drop things from the buffer if they are iterated. It doesn't look like rust provides this built in which is weird. I think you can just do something like:

```rust
core::iter::from_fn(|| self.ring_buffer.pop_front())
```

LLFourn (changes requested):
Need to look carefully at `drain`

device/src/io.rs
nickfarrow: Ah great spot, I didn't think to check if drain would behave differently.

So with this `drain`, if a device ingested magic bytes quickly followed by another message, the magic bytes would get read but the message in the buffer would be dropped.

This fix looks good!

nickfarrow:
Took your fix for `drain`, also removed one of the drains by just using `.into::<Vec<u8>>()`

LLFourn (approved):
LGTM. @nickfarrow feel free to merge if you tested this a bit.

PR #113, opened 2024-02-08
[app] Add WIP wallet

LLFourn:
That works on signet. Give it a try.


frostsnapp/native/src/wallet.rs
```diff
@@ -212,18 +216,16 @@ impl _Wallet {
         // reveal_next_spk will panic if the key doesn't exist
         self.ensure_key_exists(key_id)?;
```
nickfarrow, line 217: Is there a case where this is called before creating a key? is this a pattern we should follow elsewhere (checking inputs are valid)?

frostsnapp/native/src/wallet.rs
```diff
@@ -303,4 +305,227 @@ impl _Wallet {
...
+                    Some(
+                        self.graph.graph().calculate_fee(&tx).ok()? as f32
+                            / tx.weight().to_wu() as f32,
+                    )
+                })
+                .collect::<Vec<_>>();
+
+            let mut average = feerates.iter().sum::<f32>() / feerates.len() as f32;
+
+            if !average.is_normal() {
```
nickfarrow, line 373: Awesome to see coin selection and the change policy in action! 
`f32::is_normal` new terminology to me & useful method.

frostsnap_core/src/coordinator.rs
```diff
@@ -483,6 +483,7 @@ impl FrostCoordinator {
 
     pub fn start_sign(
         &mut self,
+        key_id: KeyId,
```
nickfarrow, line 486: These KeyId changes make sense to me, but I am wondering about how far the scope of frostsnap_core should extend.

Is it actually important for Frostsnap core to know about various keys? Though I can't think of a better place to handle it currently. Maybe you've thought about this with upcoming multiple keys PR

LLFourn: It's a good point. It could be removed for now. It got added in place of something more complicated I was doing which was making the sign task have the key in it.

frostsnapp/native/src/wallet.rs
LLFourn: I tried to clear up this call to make it more self evident: e102d341cd3cac18f50c20afd93166c9d54cc359

In general you should avoid designing methods that check inputs and valid (and that's not what this method does). You try and make APIs correct by construction by defining types with no invalid instances. This allows the caller to try and convert their data to the type and deal with the errors there before it reaches your code.

PR #114, opened 2024-02-13
Display backups on the device

nickfarrow:
Ready for review.

Devices automatically display their backup after confirming keygen.

There's now a "Show Devices" button underneath the key list.

On this page each device has a "Settings" button.

On the device settings page you can rename the device, and see a list of keys for which the device belongs. You can request the device display the backup for a particular key.

The device will ask you to confirm whether you wish to display this backup.

The backup is slightly too long for the screen of our current prototypes.


frostsnapp/lib/keygen.dart
```diff
@@ -183,37 +183,32 @@ class _DoKeyGenScreenState extends State<DoKeyGenScreen> {
 
   @override
   Widget build(BuildContext context) {
-    return PopScope(
-      onPopInvoked: (didPop) async {
-        coord.cancelAll();
```
LLFourn, line 188: We do need to cancel somehow when the back button is pressed though. The problem is that `popScope` also gets triggered when it closes successfully. I had the same problem in #117. I've got an idea I'll try and push a commit.

LLFourn: see: 11597f6074b851d7c7e827d9d83cfc2cf0c39a4e

frostsnapp/lib/key_settings.dart
```diff
@@ -0,0 +1,82 @@
...
+        body: Center(
+            child: Padding(
+          padding: EdgeInsets.all(8.0),
+          child: KeySettingsContent(frostKey: frostKey),
+        )));
+  }
+}
+
+class KeySettingsContent extends StatelessWidget {
+  final FrostKey frostKey;
```
LLFourn, line 34: Having a key settings page is a fine idea but not really the point of this PR. It would make sense to see a list of devices that had key shares for example. But the `frostKey` parameter isn't being used here.

I think the simpler approach is just to have a `Show devices` button next to `New Key` on the key list which brings up the device list and has a `⚙` on each card where you can do the renaming workflow and show backups.

LLFourn:
Cool. I've made a few comments and pushed a commit to (maybe) fix one thing. Needs a big rebase and also to upgrade `schnorr_fun` to `0.10`.

>     * Include the `SignItem` in `SignSessionProgress` so we have the tweaks for signature verification, pass/store an `EncodedFrostKey` instead of `FrostKey<Normal>`.

I don't see this change I guess it was trashed.

frostsnapp/lib/key_settings.dart
nickfarrow: Half fixed, still labeled as deviceSettings in code, need to add `⚙` and popup.

nickfarrow:
Rebased onto master, will rebase again after https://github.com/frostsnap/frostsnap/pull/117 which is more important with an easier rebase order

nickfarrow:
Now `IndividualDeviceSettingsPage` is a StatefullWidget which listens to `deviceListChangeStream` for renaming events, displaying the refreshed name after a renaming.

LLFourn:
I've pushed a commit around simplifying and unifying the device settings with the device list. Whenever you have the device list you can now edit settings or init a new device. Before you could only set up a new device on the "New key" page but could only access settings on the "Show devices" page.

In addition you can now rename the device from the settings page without having to go to a new page.

I think this PR still needs some work. The displaying of the backup doesn't work for me from the settings page. It asks me to confirm but then doesn't display the backup. I think we have to drive the display backup stuff from the phone. The phone should know when the backup workflow has been finished on the device. It should use our design language of "phone is in dialog when you should be doing something with the device".

> Now `IndividualDeviceSettingsPage` is a StatefullWidget which listens to `deviceListChangeStream` for renaming events, displaying the refreshed name after a renaming.

The commit re-did this. The approach here should have been to use a StreamBuilder with a StatelessWidget. This keeps it simpler. I've made the approach to pass in  a `DeviceId` by itself rather than a `Device` since the name can be fetched from the state. Also if you disconnect a device on the settings page it will say "device disconnected" now. This was a natural thing to do since you can only get the name from the state stream and if you can't get it you no it's no longer plugged in.

frostsnap_core/src/device.rs
```diff
@@ -490,27 +502,44 @@ impl FrostSigner {
...
-    pub fn key_backup(&mut self, key_id: KeyId) -> Result<String, ActionError> {
-        match self.keys.get(&key_id) {
-            Some(key) => {
+    pub fn display_backup_ack(&mut self) -> Result<Vec<DeviceSend>, ActionError> {
+        match self.action_state {
+            Some(SignerState::DisplayBackup {
+                key_id,
+                awaiting_ack: true,
+            }) => {
+                let key = self.keys.get(&key_id).expect("key must exist");
```
nickfarrow, line 511: Yep we now error earlier, in `recv_coordinator_message`, if a coordinator requests a backup of an unknown key

frostsnap_core/src/device.rs
```diff
@@ -536,6 +569,7 @@ impl SignerState {
             SignerState::KeyGen { .. } => "KeyGen",
             SignerState::KeyGenAck { .. } => "KeyGenAck",
             SignerState::AwaitingSignAck { .. } => "AwaitingSignAck",
+            SignerState::DisplayBackup { .. } => "DisplayBackup",
```
nickfarrow, line 572: Something i noticed,  `DisplayBackup` is now a `SignerState` inside `frostsnap_core`. At some point we may want to separate the extra device functionality above a more general `frostsnap_core::Signer`. 

This would be useful for implementing other signers for our own purposes (on coordinator or on server). With this separation we could document a set of fundamental `CoordinatorToSignerMessage`s that can be implemented by compatible coordinators/hwws without the bells and whistles .

nickfarrow:
Nice! tested. The showDeviceActionDialog is great.

There is one unrelated flow after keygen where we can progress the device screen before reaching the app dialogue screen.

Somewhere down the track we should change "Cancel" to "Done" or "Back" for workflows like this.

Added a commit so that the backup interpolation tests with a threshold number of backups, and still checks all N of them also. Might be slight overkill

frostsnap_core/src/device.rs
LLFourn: Is displaying the backup an "extra functionality" though? It seems like core functionality. Remember that the concrete device impl can decide what to do when the coordinator instructs the device to display backup -- it can always deny the request for example.

frostsnap_core/tests/endtoend.rs
```diff
@@ -348,11 +344,19 @@ fn test_display_backup() {
             (decoded.share_index, decoded.secret_share)
         })
         .collect::<Vec<_>>();
-    let interpolated_joint_secret =
-        schnorr_fun::fun::poly::scalar::interpolate_and_eval_poly_at_0(decoded_backups);
 
+    let threshold_interpolated_joint_secret =
+        schnorr_fun::fun::poly::scalar::interpolate_and_eval_poly_at_0(
+            decoded_backups[..(threshold as usize)].to_vec(),
```
LLFourn, line 350: :bulb: Can you make the test just randomly choose two signers and makes them display the backup and then interpolate. This will be better than fixed indices.

frostsnap_core/tests/endtoend.rs
```diff
@@ -348,11 +344,19 @@ fn test_display_backup() {
...
+        );
+    let all_interpolated_joint_secret =
```
LLFourn, line 352: not sure what `all_` is doing here.

LLFourn (approved):
Approved. I think there's a small improvement you can make to the test that you fixed.

nickfarrow:
> Is displaying the backup an "extra functionality" though? It seems like core functionality. 

Core to a device, definitely. Core to a "signer", i'm not so sure. I was thinking we could distinguish fundamental FROST protocol messages. But good point about denying requests, can just leave impls blank.

I've made the other suggested changes.

PR #117, opened 2024-02-26
Separate nonces from key generation

LLFourn:
1. Nonces are separated from key generation. The coordinator requests nonces from a device after it is first named.
2. Coordinator and device now store "to storage" core messages in an event log type architecture and re-apply them on startup

This allows devices and coordinator to have multiple keys.

## Notes

I ran into a lot of low level problems while doing this PR.

1. Messages sent from the device to the coordinator were not fully received by the coordinator (they were truncated). This was fixed after a long investigation by changing how JTAG writing works on the device side. We write each byte manually now (similar to uart) and then flush. With JTAG calling flush actually seems to do something at the hardware level. This magic "thing" was also being done by the `.write_bytes` call we were making before but perhaps because of the frequency with which we were calling it (several times for a single message), it didn't work.
2. Messages sent from the coordinator to the first device over JTAG were also being truncated. In particular we were sending a superfluous cancel message when the naming device dialog closed successfully. This cancel message wasn't fully received and panic'd the device. This in theory should be fine since you'd expect a short message like this to just end up in the hardware buffer until the device was ready to deal with it  -- but maybe the JTAG serial buffer is smaller than the uart one? Anyway I fixed the bug by not sending the  cancel message.
3. Having more data in the device's flash memory meant that it was taking longer to start up which throws off the delicate timings of JTAG switching. I've simply delayed the JTAG switching. Hopefully we can get rid of this timing soon with upstream detection.
4. Panic messages not being displayed -- just the led goes red. After a lot of work I found that upgrading esp-hal fixed this. The upgrade will come in a follow up PR.

I think there might still be some difficult to reproduce panics around but I they'll be difficult to debug until we upgrrade esp-hal.

Note: A lot of lines changed in core were due to reorganizing match statements not actual logic changes.


device/src/storage.rs
```diff
@@ -1,73 +1,122 @@
...
     enc::write::Writer,
     error::{DecodeError, EncodeError},
 };
 use embedded_storage::{ReadStorage, Storage};
 use esp_storage::{FlashStorage, FlashStorageError};
+use frostsnap_core::{message::DeviceToStorageMessage, schnorr_fun::fun::Scalar};
 
-pub const NVS_PARTITION_START: u32 = 0x9000;
-pub const NVS_PARTITION_SIZE: usize = 0x6000;
+const NVS_PARTITION_START: u32 = 0x9000;
```
nickfarrow, line 12: I think you asked @musdom this elsewhere, could we add a comment where these come from?

device/src/storage.rs
```diff
@@ -1,73 +1,122 @@
...
-    flash: FlashStorage,
-    start_pos: u32,
+#[derive(Debug, Clone, bincode::Encode, bincode::Decode)]
+pub enum Change {
+    Core(DeviceToStorageMessage),
+    Name(String),
+}
+
+#[derive(Debug, Clone, bincode::Encode, bincode::Decode)]
+pub struct Header {
```
nickfarrow, line 31: I want to understand the Header better: is this for data which will never change? What else might end up going in here?

frostsnap_core/src/coordinator.rs
```diff
@@ -3,485 +3,412 @@ use crate::{
...
+            action_state: None,
+            device_nonces: Default::default(),
         }
     }
 
-    pub fn from_stored_key(key: CoordinatorFrostKeyState) -> Self {
-        Self {
-            state: CoordinatorState::FrostKey { key },
+    pub fn apply_change(&mut self, change: CoordinatorToStorageMessage) {
+        use CoordinatorToStorageMessage::*;
```
nickfarrow, line 39: Just for shorthanding the enum variants?

frostsnap_core/src/coordinator.rs
```diff
@@ -517,134 +448,242 @@ impl FrostCoordinator {
...
-            }),
+        Ok(outgoing)
+    }
+
+    pub fn maybe_request_nonce_replenishment(
+        &self,
+        device_id: DeviceId,
+    ) -> Option<CoordinatorSend> {
+        let needs_replenishment = match self.device_nonces.get(&device_id) {
+            Some(device_nonces) => device_nonces.nonces.len() < MIN_NONCES_BEFORE_REQUEST,
```
nickfarrow, line 570: Nice! I like it

frostsnap_core/src/device.rs
```diff
@@ -69,36 +63,49 @@ impl FrostSigner {
         DeviceId::new(self.keypair().public_key())
     }
 
-    pub fn generate_nonces(
-        &mut self,
-        aux_rand: [u8; 32],
-        start: usize,
-        n: usize,
+    fn generate_nonces(
+        // this is always the device key for now but because of lifetimes issues it has to be passed in
```
nickfarrow, line 67: Is this still the case?

I tried changing to 
```
 fn generate_nonces(
        &self,
        start: u64,
)
```
and everything seems to compile, maybe i'm missing something!

frostsnap_core/src/device.rs
```diff
@@ -254,129 +243,181 @@ impl FrostSigner {
...
-                let (my_nonces, my_nonce_index, _) = match nonces.get(&key.share_index) {
-                    Some(nonce) => nonce,
-                    None => return Ok(Vec::new()),
-                };
+                let key = self.keys.get(&key_id).ok_or_else(|| {
+                    Error::signer_invalid_message(
+                        &message,
+                        // we could instead send back a message saying we don't have this key but I
+                        // think this will never happen in practice unless we have a way for one
+                        // coordinator to delete a key from a device without the other coordinator
```
nickfarrow, line 271: I could see this being a possibility in the future, but yes this is good for now!

frostsnap_core/src/device.rs
```diff
@@ -254,129 +243,181 @@ impl FrostSigner {
...
+                        "this device was asked to sign but no nonces
+                were provided",
+                    )
+                })?;
 
                 let n_signatures_requested = sign_task.sign_items().len();
-                if my_nonces.len() != n_signatures_requested {
-                    return Err(Error::signer_invalid_message(&message, format!( "Number of nonces ({}) was not the same as the number of signatures we were asked for {}", my_nonces.len(), n_signatures_requested)));
+                if my_nonces.nonces.len() != n_signatures_requested {
+                    return Err(Error::signer_invalid_message(&message, format!("Number of nonces ({}) was not the same as the number of signatures we were asked for {}", my_nonces.nonces.len(), n_signatures_requested)));
```
nickfarrow, line 288: I'm surprised this passes clippy fmt, but I also have had trouble formatting multiline strings in rust 😅

frostsnap_core/src/device.rs
```diff
@@ -254,129 +243,181 @@ impl FrostSigner {
...
+                    //
+                    self.nonce_counter = my_session_nonces
+                        .start
+                        .saturating_add(sign_items.len() as u64);
+
+                    // This calculates the index after the last nonce the coordinator had. This is
+                    // where we want to start providing new nonces.
+                    let replenish_start = self.nonce_counter + my_session_nonces.nonces_remaining;
+                    // How many nonces we should give them from that point
+                    let replenish_amount =
```
nickfarrow, line 386: So the coordinator always replenishes up to `NONCE_BATCH_SIZE`.

It feels a little odd that the coordinator is keeping track of whether it needs to request more nonces, while also the device is calculating this on their side also.

What goes wrong if we only replenish the number of nonces used in this signing session? The coordinator could begin multiple sign sessions that the device never sees and run out of nonces. But couldn't the coordinator request more nonces before starting the next session?

frostsnap_core/src/device.rs
```diff
@@ -403,105 +444,97 @@ impl FrostSigner {
...
+}
+
+impl FrostsnapSecretKey {
+    pub fn key_id(&self) -> KeyId {
+        self.frost_key.key_id()
+    }
 }
 
 #[derive(Debug, Clone)]
 pub enum StartSignError {
```
nickfarrow, line 533: I think this should be in `coordinator.rs`? Moved in previous PR

Realised because `StartSignError::DeviceNotPartOfKey` doesn't 't make sense from the perspective of the device, previously we discussed that devices do not need to have knowledge of other device ids.

frostsnap_core/src/device.rs
```diff
@@ -69,36 +63,49 @@ impl FrostSigner {
...
+        // this is always the device key for now but because of lifetimes issues it has to be passed in
+        secret_key: &Scalar,
+        start: u64,
     ) -> impl Iterator<Item = NonceKeyPair> + '_ {
         let mut nonce_rng = derive_nonce_rng! {
             nonce_gen => nonce::Deterministic::<Sha256>::default().tag(b"frostsnap/nonces"),
-            secret => self.keypair.secret_key(),
-            public => [aux_rand],
+            secret => secret_key,
+            public => [b""],
```
nickfarrow, line 74: Why did we drop `aux_rand` here? It certainly was annoying, but are we now deriving the same nonces for every key? I guess this is fine with a global device nonce_counter?

device/src/storage.rs
LLFourn: Yes. Never change unless it's factory reset.

frostsnap_core/src/coordinator.rs
LLFourn: Yes.

frostsnap_core/src/device.rs
LLFourn: cargo fmt ignores macros because it's nt sure how changing the contents will change the semantics I guess.

frostsnap_core/src/device.rs
LLFourn: > So the coordinator always replenishes up to `NONCE_BATCH_SIZE`.
> 
> It feels a little odd that the coordinator is keeping track of whether it needs to request more nonces, while also the device is calculating this on their side also.
> 
> What goes wrong if we only replenish the number of nonces used in this signing session? The coordinator could begin multiple sign sessions that the device never sees and run out of nonces. But couldn't the coordinator request more nonces before starting the next session?

Do you mean send another message sent right after signing to get more nonces?
We could do that *instead* of sending nonces here but I'm not sure what the motivation is. The motivation for slipping nonces into the signature response is to avoid spontaneously sending messages while processing could be happening (and the message gets dropped because we only have one core).

frostsnap_core/src/device.rs
LLFourn: good catch!

frostsnap_core/src/device.rs
LLFourn: `aux_rand` used to be there so we got a different set of nonces per key we generated. Now we don't nonces per key so it no longer makes sense. We will have different nocnes per coordinator but it was premature to introduce this before supporting a concept of coordinator identity.

frostsnap_core/src/device.rs
LLFourn: Good catch. Thanks!

device/src/storage.rs
nickfarrow: https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-guides/partition-tables.html

device/src/storage.rs
nickfarrow: Maybe silly question, but in this case do we need a header at all? Or maybe in future we'll purge some of change set?

frostsnap_core/src/device.rs
nickfarrow: The motivation was to avoid passing the number of remaining nonces to the device. But this seems to make a bigger and more fragile mess.

The idea was to replenish nonces alongside signing as we are currently, but only replenish `sign_items.len()`.

The coordinator would `maybe_replenish_nonces` based on the size of `NONCE_BATCH_SIZE.saturating_sub(device_nonces.nonces.len()`.

I got it almost fully working, but the last step would be for the coordinator to call this replenish nonces somewhere **prior** to signing (so it works in the event that nonces were not replenished).

It also reintroduces the `NONCE_BATCH_SIZE` onto the coordinator, which as discussed could cause issues if this value were to change.

Can take a look here if curious 5ef1984f7b3b2d6bd65fb5263d3f9f358ebaa997

nickfarrow (approved):

PR #118, opened 2024-02-27
Upgrade esp-hal v0.16.1

LLFourn:
This allows us to drop the nightly toolchain. I deleted the toolchain.toml file.

One peculiar thing I had to do was lower the heap size to avoid some allocation failures. It would be really good to understand why this was necessary before merging.

Main benefit of this is that it seems to fix the problem of panic messages not being displayed in the panic handler. Note i'm not sure if that's to do with lowering the heap size or just the dependency upgrade.

This is on top of #117

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
           tar -zxf riscv32-elf-ubuntu-22.04-nightly-2023.05.24-nightly.tar.gz
           echo "PATH=$PATH:${GITHUB_WORKSPACE}/device/riscv/bin" >> $GITHUB_ENV
           ls ${GITHUB_WORKSPACE}/device/riscv/bin
-      - uses: dtolnay/rust-toolchain@nightly
+      - uses: dtolnay/rust-toolchain@stable
         with:
           targets: riscv32imc-unknown-none-elf
           components: clippy
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,6 +108,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bare-metal"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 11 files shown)


LLFourn:
I just pushed a commit to clean up display errors in `st7735.rs`. We need `error_print` to be infallible so we don't panic inside the panic handler.

LLFourn:
@musdom do you have any idea why the heap size change was necessary?

musdom:
Not sure. I tried flashing this 0.15 template https://github.com/esp-rs/esp-template and was able to give it 300kB no problem.

nickfarrow:
Nice so many unwraps removed :broom:

Rebased and upgraded `hal` to `0.16.0` which notably makes `esp32c3` a feature rather than its own package.

Tested and reviewed. 

---

I could not figure out the cause of the memory allocation error.

I looked through the `hal` changelogs but nothing really stood out, couldnt quite understand https://github.com/esp-rs/esp-hal/pull/1289.

Tried finding which line the error was occurring on using panics. 
When using a 300kb heap size on a device that already has a key, the panic seems to be happening in this line `match flash.read_header()`, but I don't think there's anything of interest happening in here? Memory probably being exhausted elsewhere, i can keep looking

LLFourn:
I'm pretty sure @musdom already discovered that it was changes to the display driver.

nickfarrow:
Confirmed it seems to be SPI/display related. When using a 166+ byte heap, I am unable to allocate anything after intializing the display.

Haven't managed to figure out much more..

There are some things in the changelog which look interesting,
https://github.com/esp-rs/esp-hal/blob/main/esp-hal/CHANGELOG.md
enabling `place-spi-driver-in-ram` seemed to have no effect.

I began to try the unreleased `esp-hal`, but didn't quite manage to build for our device, not sure if this avenue is worth exploring.

nickfarrow (approved):

Issue #116, opened 2024-02-21
Remove dependency on ring


LLFourn:
I think if rustls in electrum client were upgraded we wouldn't have this dependency.
---

Works great!

Ack changes and tested on desktop.

I need to fix my environment issues so i can compile to android https://github.com/briansmith/ring/issues/1488

_Originally posted by @nickfarrow in https://github.com/frostsnap/frostsnap/pull/113#pullrequestreview-1890852703_

PR #120, opened 2024-04-17
Update BDK to 0.15 with patches

nickfarrow:
* Upgrades bdk to version 0.12 with minor api changes.
 * Patch keeps the keychain_txout_more_range changes from
      https://github.com/bitcoindevkit/bdk/pull/1324,
 * This bdk rev uses an updated electrum-client with upgraded rustls.
 * Frostsnapp no longer depends on ring.

https://github.com/bitcoindevkit/rust-electrum-client/pull/132

Fixes: https://github.com/frostsnap/frostsnap/issues/116

### DIFF
--- a/frostsnapp/native/src/api.rs
+++ b/frostsnapp/native/src/api.rs
@@ -381,7 +381,8 @@ fn _load(db_file: String, usb_serial_manager: UsbSerialManager) -> Result<(Coord
         db.clone(),
         bitcoin::Network::Signet,
         persist_core_handle,
-    )?;
+    )
+    .with_context(|| format!("loading wallet from data in {db_file}"))?;
     let coordinator = Coordinator(RustOpaque::new(coordinator));
     let chain_sync = ChainSync::new(wallet.network)?;
     let wallet = Wallet::new(wallet, chain_sync);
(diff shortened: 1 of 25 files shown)


nickfarrow:
Got rid of ring!!!
And replaced it with `aws-lc-sys`.......... also not buidling with some architectures

nickfarrow:
TODO: since important changes got merged do not use patch, use BDK version

frostsnapp/native/src/api.rs
```diff
@@ -647,16 +649,18 @@ impl Wallet {
...
-        let mut sync_request = {
+        let sync_request = {
+            let inspect_stream = stream.clone();
             let wallet = self.inner.lock().unwrap();
-            wallet.start_sync(key_id)
+            let sync_req = wallet.start_sync(key_id);
+            let total = sync_req.spks.len();
+            let mut i = 0;
+            sync_req.inspect_spks(move |_spk| {
+                inspect_stream.add((i as f64) / (total - 1) as f64);
```
nickfarrow, line 659: On this progress of syncing spks, there can never be just one spk?

frostsnapp/native/src/wallet.rs
```diff
@@ -519,8 +557,8 @@ impl _Wallet {
                     .api(&tx)
                     .push(&bincode::serde::Compat(changes.graph))?;
             }
-            tx.take_index(self.keychain_handle)
-                .extend(changes.indexer.0)?;
+            // it's unlikely to have any changes but just in case
+            self.consume_spk_revelation_change(tx, changes.indexer)?;
```
nickfarrow, line 616: Could you help me understand this one?

frostsnapp/native/src/wallet.rs
```diff
@@ -181,22 +192,48 @@ impl _Wallet {
...
 
-        let tr = Tr::new(key, None).expect("infallible since it's None");
-        Descriptor::Tr(tr)
+        let (app_key, chaincode) =
+            frost_key.app_tweak_and_expand(frostsnap_core::tweak::AppTweakKind::Bitcoin);
+        let root_bitcoin_xpub = *frostsnap_core::tweak::Xpub::new(app_key, chaincode).xpub();
+
+        [Keychain::External, Keychain::Internal]
+            .into_iter()
+            .map(move |keychain| {
```
nickfarrow, line 206: I'm not sure this requires `move` since we're already taking ownership through `into_iter` and nothing in the closure is getting mutated?

frostsnapp/native/src/api.rs
nickfarrow: This does indeed cause a /0 error if there's only one receive address in the graph.

frostsnapp/native/src/wallet.rs
LLFourn: ill try and rewrite this so the comment goes away by having a method that applies a changeset. It's unlikely that there are any changes related to spk revelation in the preceding operations.

frostsnapp/native/src/api.rs
LLFourn: Yes it should be +1 to the numerator rather than -1 from the denominator.

LLFourn: 961f801cc76f98af994df0135e8bf952c6767607

LLFourn (approved):

PR #121, opened 2024-04-19
Exploring Nostr

nickfarrow:
So far is post only, and vendored openssl isn't building for me with android targets. If I can get that building I'd like to try using it.

Is there a way we can avoid building an entire nostr client ourselves, while still providing a useful experience with our devices?
Can we hook into other clients?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1377,6 +1377,7 @@ dependencies = [
  "frostsnap_coordinator",
  "lazy_static",
  "llsdb",
+ "openssl",
  "tracing",
  "tracing-android",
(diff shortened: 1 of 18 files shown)


midizzz:
I think that this will be cool but isn't needed before launch.. Frost bitcoin is cool by itself ;)

nickfarrow:
> I think that this will be cool but isn't needed before launch.. Frost bitcoin is cool by itself ;)

This is what we also decided!

LLFourn:
going to close this since this has bitrotted and there may be other approaches available to us if we did this now.

PR #122, opened 2024-04-22
Upgrade esp-hal to 0.18.0

nickfarrow:
Compiles using the cst816s fork @musdom found.

Also upgrades to embedded_hal 1.0.

Currently the screen display is blank.

* The new `mipsi::Builder` wants to receive an `SpiDevice` and not the `SpiBus`, so we need to choose a CS pin? Not sure what to put here.
* Is `Orientation::Portrait(false)` equal to `Rotation:Deg90`?
* I'm also setting the reset pin to high (https://github.com/almindor/mipidsi/blob/master/docs/TROUBLESHOOTING.md)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,15 @@ resolver = "2"
 
 
 [workspace.dependencies]
-bincode = {  version = "2.0.0-rc.3", features = ["serde", "alloc", "derive"], default-features = false }
-serde = { version = "1", features = ["derive","alloc"], default-features = false }
+bincode = { version = "2.0.0-rc.3", features = [
+    "serde",
+    "alloc",
+    "derive",
+], default-features = false }
+serde = { version = "1", features = [
+    "derive",
+    "alloc",
+], default-features = false }
 frostsnap_core = { path = "frostsnap_core", default-features = false }
 frostsnap_comms = { path = "frostsnap_comms", default-features = false }
 frostsnap_coordinator = { path = "frostsnap_coordinator" }
@@ -26,3 +33,18 @@ sha2 = { version = "0.10", default-features = false }
 # bdk_chain = { path = "../bdk/crates/chain" }
 bdk_chain = { git = "https://github.com/LLFourn/bdk.git", rev = "165f06b8f78bd24f57dfbfd66da1faa189c67457" }
 bdk_electrum = { git = "https://github.com/LLFourn/bdk.git", rev = "165f06b8f78bd24f57dfbfd66da1faa189c67457" }
+
+[profile.dev]
+# Rust debug is too slow.
+# For debug builds always builds with some optimization
+opt-level = "s"
+
+[profile.release]
+# LLVM can perform better optimizations using a single thread
+codegen-units = 1
+debug = 2
+debug-assertions = false
+incremental = false
+lto = 'fat'
+opt-level = 's'
+overflow-checks = false
--- a/device/Cargo.toml
+++ b/device/Cargo.toml
@@ -16,22 +16,24 @@ mem_debug = []
 [dependencies]
 frostsnap_core = { workspace = true }
 frostsnap_comms = { workspace = true }
-esp-hal = { package = "esp-hal", version = "0.16.1", features = ["esp32c3"] }
+esp-hal = { package = "esp-hal", version = "0.17.0", features = ["esp32c3"] }
 esp-alloc = "0.3.0"
 nb = "1"
+embedded-hal-nb = "1"
+embedded-hal-bus = "0.1.0"
 embedded-storage = "0.3.1"
 esp-storage = { version = "0.3.0", features = ["esp32c3"] }
 bincode = { workspace = true }
-display-interface = { version = "0.4.1", optional = true }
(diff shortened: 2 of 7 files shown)


musdom:
Display works but now top devices upstream go in uart mode

musdom:
Todo: remove tick constant in esp32_run.rs and io.rs

musdom:
Rebased over #120

Cargo.toml
```diff
@@ -11,8 +11,15 @@ resolver = "2"
...
-serde = { version = "1", features = ["derive","alloc"], default-features = false }
+bincode = { version = "2.0.0-rc.3", features = [
+    "serde",
+    "alloc",
+    "derive",
+], default-features = false }
+serde = { version = "1", features = [
+    "derive",
+    "alloc",
+], default-features = false }
```
LLFourn, line 22: Can we not have this weird formatting please? I'm guessing this was done automatically without noticiing?

Cargo.toml
```diff
@@ -26,3 +33,18 @@ sha2 = { version = "0.10", default-features = false }
 # bdk_chain = { path = "../bdk/crates/chain" }
 bdk_chain = { git = "https://github.com/LLFourn/bdk.git", rev = "165f06b8f78bd24f57dfbfd66da1faa189c67457" }
 bdk_electrum = { git = "https://github.com/LLFourn/bdk.git", rev = "165f06b8f78bd24f57dfbfd66da1faa189c67457" }
+
+[profile.dev]
+# Rust debug is too slow.
+# For debug builds always builds with some optimization
+opt-level = "s"
```
LLFourn, line 40: I think this is a good idea. I've been just doing --release even when testing because of this.

Cargo.toml
musdom: github CI formatted it because it wasn't like this in my local repo

Cargo.toml
```diff
@@ -26,3 +33,18 @@ sha2 = { version = "0.10", default-features = false }
...
+
+[profile.release]
+# LLVM can perform better optimizations using a single thread
+codegen-units = 1
+debug = 2
+debug-assertions = false
+incremental = false
+lto = 'fat'
+opt-level = 's'
```
LLFourn, line 49: Why are we doing this in the workspace Cargo.toml and not the individual crates. Surely we'll want a different profile for device than everything else.

device/Cargo.toml
```diff
@@ -16,23 +16,26 @@ mem_debug = []
...
+display-interface-spi = { version = "0.5.0", optional = true }
+mipidsi = { version = "0.8.0", features = ["batch"], optional = true }
 embedded-graphics-framebuf = { version = "0.5.0", optional = true }
 embedded-graphics = "0.8.1"
 embedded-text = "0.7.0"
 u8g2-fonts = { version = "0.3.0", features = ["embedded_graphics_textstyle"] }
 rand_chacha = { workspace = true }
-cst816s = "0.1.4"
-embedded-hal = { version = "0.2.7" }
+cst816s = { git = "https://github.com/fxweidinger/cst816s", rev = "82f88186e94a94126b962d30b4c0129fd3c1c5f8" }
```
LLFourn, line 35: We should probably vendor this inside the workspace at some point if we're going to use it in production.

Cargo.toml
musdom: Cargo complains profiles should be in the root Cargo.toml

LLFourn: Apparently not possible so I guess we'll just suck it up for now but at some point we may want to split out device into its own repo.

device/src/bin/v2.rs
```diff
@@ -109,13 +112,14 @@ fn main() -> ! {
...
         .with_mosi(io.pins.gpio7);
-    let di = SPIInterfaceNoCS::new(spi, io.pins.gpio9.into_push_pull_output());
-    let display = Builder::st7789(di)
-        .with_display_size(240, 280)
-        .with_window_offset_handler(|_| (0, 20)) // 240*280 panel
-        .with_invert_colors(ColorInversion::Inverted)
-        .with_orientation(Orientation::Portrait(false))
-        .init(&mut delay, Some(io.pins.gpio6.into_push_pull_output())) // RES
+    let spi_device = embedded_hal_bus::spi::ExclusiveDevice::new_no_delay(spi, NoCs);
+    let di = SPIInterface::new(spi_device, Output::new(io.pins.gpio9, Level::Low));
```
LLFourn, line 116: It'd be interesting to know what the `embedded_hal_bus` crate is doing that the original Spi type was not. It sounds like from the documentation it helps different devices share the bus. Are we doing that?

musdom: Nope, it's just for the display

LLFourn: Right so why do we need the `embedded_hal_bus` crate. What goes wrong without it?

LLFourn (approved):
Working nicely for me and LGTM.

device/src/bin/v2.rs
musdom: display_interface_spi expects that SpiDevice embedded-hal-bus type now

Issue #124, opened 2024-04-23
Fix nonce replenishing after cancelled sign requests


nickfarrow:
If you cancel a sign request, the devices are trying to replenish more nonces than the coordinator is expecting

```
 ERROR native::coordinator: Failed to process message: Invalid message of kind SignatureShare: 
Signer did not replenish the correct number of nonces. Expected 1, got 3, from: 
"0265a62436444c83d70d0536ad3dee76dcaaf45869a294c4d4327a37855f6b5268"
    at frostsnapp/native/src/coordinator.rs:196
```

PR #125, opened 2024-04-23
Allow over-replenishment of nonces

nickfarrow:
Fixes https://github.com/frostsnap/frostsnap/issues/124

### DIFF
--- a/frostsnap_core/tests/endtoend.rs
+++ b/frostsnap_core/tests/endtoend.rs
@@ -364,3 +364,101 @@ fn test_display_backup() {
         coord_frost_key.frost_key().public_key()
     );
 }
+
+#[test]
+fn fully_replenish_nonces() {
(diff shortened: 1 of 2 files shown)


frostsnap_core/tests/endtoend.rs
```diff
@@ -364,3 +364,96 @@ fn test_display_backup() {
...
+    };
+
+    let used_sign_request = run
+        .coordinator
+        .start_sign(key_id, completing_sign_task, device_set)
+        .unwrap()
+        .clone();
+
+    run.extend(used_sign_request);
+    run.run_until_finished(&mut env, &mut test_rng);
```
LLFourn, line 441: worth saying here that the reason for this test is just to test that this runs without erroring and the asserts around are just sanity checks.

LLFourn (approved):

PR #127, opened 2024-04-26
Fix dropped tweak when combining signature shares

nickfarrow:
* With the EncodedFrostKey PR we accidentally dropped the tweaks because we use a too-convenient into method.
* This meant we combined the signature shares without the tweak, causing invalid schnorr signatures in our bitcoin wallet.
* We now verify the signature after combining the signature shares.

### DIFF
--- a/frostsnap_core/src/coordinator.rs
+++ b/frostsnap_core/src/coordinator.rs
@@ -6,9 +6,9 @@ use alloc::{
     vec::Vec,
 };
 use schnorr_fun::{
-    frost::{self, EncodedFrostKey, FrostKey, SignSession},
+    frost::{self, EncodedFrostKey, Frost, FrostKey, SignSession},
     fun::{marker::*, Scalar},
-    Message, Schnorr,
+    Message, Schnorr, Signature,
 };
 use sha2::Sha256;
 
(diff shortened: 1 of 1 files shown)


frostsnap_core/src/coordinator.rs
```diff
@@ -351,18 +350,25 @@ impl FrostCoordinator {
...
                                 session_progress
                                     .signature_shares
                                     .iter()
                                     .map(|(_, &share)| share)
                                     .collect(),
-                            )
+                            );
+
+                            assert!(session_progress.sign_item.verify(
+                                &Schnorr::verify_only(),
```
LLFourn, line 366: I think we can just use the `frost.schnorr()` instance here.

frostsnap_core/src/coordinator.rs
```diff
@@ -351,18 +350,25 @@ impl FrostCoordinator {
...
+                                session_progress.key.into_frost_key().public_key(),
```
LLFourn, line 367: There's something odd here. Why do we call `session_progress.sign_item.verify` and then pass in another `session_progress` thing into the call. It looks like `.verify` should just be on `session_progress`.

LLFourn:
I pushed a commit addressing my comments and cleaning things up a bit. I think we should change the name of `EncodedFrostKey` to `RootFrostKey` and use it whenever we want a tweakless key.

LLFourn (approved):

frostsnap_core/src/coordinator.rs
```diff
@@ -670,13 +664,44 @@ pub struct SignSessionProgress {
 }
 
 impl SignSessionProgress {
+    pub fn new<NG>(
+        frost: &Frost<sha2::Sha256, NG>,
+        root_key: EncodedFrostKey,
+        sign_item: SignItem,
+        nonces: BTreeMap<frost::PartyIndex, frost::Nonce>,
+    ) -> Self {
+        // FIXME: This shouldn't be raw -- plain messages should do domain separation
```
nickfarrow, line 673: 👍

will need to look at specs

PR #129, opened 2024-05-10
PSBT Signing in App

nickfarrow:
This PR:
* Adds the ability to load PSBTs and sign them. 
    - Load button can be found on wallet send page.
    - PSBTs can be broadcast or saved back to a file.
* Scan PSBT QR codes (Universal Resources)
    - Handles static and fountain codes.
    -  QR recognition and UR decoding is done on the rust side.
    - Introduces some borrowed image conversion code which *should* also work on IOS.
    - Scan button is hidden on desktop (linux not supported by camera library)
    - Progress bar based on old metric used in seedsigner (` detected / (1.75 * sequence length)`)  which is pretty good for our purposes. Required a commit on `ur-rs` to expose the sequence length in api.
* Displays the wallet descriptor on the receive page of the wallet, tpub for signet.

Currently we aren't using a change keychain in the wallet, spending change from `1/0` returns invalid signature share (wrong tweak).


frostsnapp/native/src/wallet.rs
```diff
@@ -411,7 +416,7 @@ impl _Wallet {
         for ((_, i), selected_utxo) in selected_utxos {
             prevouts.push(frostsnap_core::message::TxInput {
                 prevout: selected_utxo.txout.clone(),
-                bip32_path: Some(vec![*i]),
+                bip32_path: Some(vec![0, *i]), // TODO [0,i]?
```
nickfarrow, line 419: Needs to be modified for change keychains

frostsnapp/native/src/api.rs
```diff
@@ -823,16 +951,31 @@ impl Wallet {
 
 pub struct SignedTx {
     pub inner: RustOpaque<RTransaction>,
+    pub psbt: Option<RustOpaque<bitcoin::psbt::PartiallySignedTransaction>>,
 }
```
nickfarrow, line 955: One of the more significant changes: decided to add an Option of a PSBT to `UnsignedTx` and `SignedTx`, this was easiest, keeps APIs mostly the same.

frostsnapp/native/src/api.rs
```diff
@@ -823,16 +951,31 @@ impl Wallet {
...
 
 impl SignedTx {
     pub fn tx(&self) -> SyncReturn<RustOpaque<RTransaction>> {
         SyncReturn(self.inner.clone())
     }
+
+    pub fn save_psbt_to_file(&self, filename: String) -> Result<SyncReturn<()>> {
```
nickfarrow, line 962: It was easier to save the file using rust than return unknown length lists to flutter

nickfarrow:
There's a panic when we load a PSBT that requires signatures from other parties `missing `TxOut` for one or more of the inputs of the tx`. It also happens if you scan a PSBT that requires no signatures from you. I think the workflow moves to broadcasting without checking whether it has all the signatures. 

In this scenario, we can save back to PSBT file?

nickfarrow:
A spinning wheel should display while the PSBT is being broadcast

frostsnapp/native/src/api.rs
nickfarrow: Separated Psbt into its own type in 0b00f4f29ff4adefa43eb4d78da4034d08dd9541

nickfarrow:
LGTM! (can't approve my own PR, and feel free to try find that flutter panic again)

LLFourn:
I couldn't reproduce whatever issue I was having anymore.

I made a few more commits to clean things up a bit. More importantly I got rid of the janky descriptor text thing and put it into a "export descriptor" dialog that gives you a QR code to scan it. I haven't done PSBT animation QRs to export a signed PSBT back to sparrow yet.

LLFourn (approved):

LLFourn:
@nickfarrow needs some warning fix?

PR #130, opened 2024-05-15
App stylization

nickfarrow:
Basic global colours and theme. Feel free to experiment with different colours when testing!


LLFourn (approved):
Thanks. Let's roll we this.

PR #131, opened 2024-05-24
Concise and formatted share backups

nickfarrow:
Uses schnorr_fun patch https://github.com/LLFourn/secp256kfun/pull/182

Also re-adds displaying backups after keygen, with backup instructions in-app.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1530,6 +1530,7 @@ dependencies = [
  "image",
  "lazy_static",
  "rusqlite",
+ "sha2",
  "tracing",
  "tracing-android",
  "tracing-subscriber",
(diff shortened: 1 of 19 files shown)


nickfarrow:
Rebasing onto OTA, will rebase onto master once that is merged.

I guess there is no need to modify the existing backup UiProtocol? Since we're just emitting a ShowBackup message in frostsnap_core for the initial backup prompt.

LLFourn (approved):
I added a commit to make the display of the backup in columns of three. I also made the "frost" the same color so they know to back it up too.

nickfarrow:
Looks great, will revisit number of columns when entering backups - I think Adam preferred 4

PR #132, opened 2024-05-28
OTA updates

LLFourn:
Assorted changes:

1. All devices plugged in are put into update mode and then updates are sent down the cable. All devices update in tandem. The device connected to the coordinator will only request the next sector when itself and all the downstream devices are ready.
2. Protocols that use a `StreamSink` to communicate to the flutter ui have put put into a trait: `UiProtocol`. These get notified of certain messages and output things to their internal `StreamSink`. Main thing is that you can only have one at a time.
3. The firmware for the upgrade is embedded in the native rust crate via `build.rs`. Right now this is done via the `espflash` cli utility. In the future it should be done via using espflash as a library.
4. App now no longer goes to sleep if there is a device plugged in.
5. We use the Sha peripheral to do the firmware hashing.


device/src/lib.rs
```diff
@@ -6,17 +6,27 @@ extern crate alloc;
...
+    pub is_device: bool,
+    pub state: UpstreamConnectionState,
+}
+
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum UpstreamConnectionState {
-    Disconnected,
-    Connected { is_device: bool },
-    Established { is_device: bool },
+    /// We're always at least connected since how did we get power unless we were connected!
```
nickfarrow, line 24: :grin:

device/src/ota.rs
```diff
@@ -0,0 +1,521 @@
...
+                } else {
+                    seq + 1
+                }
+            }
+            None => 1,
+        };
+
+        let cs = CRC.checksum(&next_seq.to_le_bytes());
+        let mut bytes = [0xffu8; ESP32_OTADATA_SIZE as usize + FS_PARTITION_METADATA_SIZE as usize];
+        bytes[0..4].copy_from_slice(&next_seq.to_le_bytes());
```
nickfarrow, line 229: The purpose of this checksum is to guarantee that we never load into an old partition if these 'sequence' bytes were corrupted?

device/src/ota.rs
```diff
@@ -0,0 +1,521 @@
...
+                                ui::Prompt::ConfirmFirmwareUpgrade {
+                                    firmware_digest: *expected_digest,
+                                    size: *size,
+                                },
+                            )),
+                        )
+                    }
+                    State::Erase { seq } => {
+                        let mut finished = false;
+                        for _ in 0..32 {
```
nickfarrow, line 333: Is this some maximum number of sectors per image?

LLFourn: Good question. No it was just to make it go faster by updating the screen less. I'll see if it's still worthwhile.

device/src/ota.rs
LLFourn: Yes. This is just what the default bootloader wants so that's what it gets.

LLFourn:
I've updated this PR to increase the baudrate during the upgrade which makes it much faster.

frostsnap_coordinator/src/usb_serial_manager.rs
```diff
@@ -536,6 +554,92 @@ impl UsbSerialManager {
...
+    }
+
+    pub fn upgrade_bin(&self) -> FirmwareBin {
+        self.firmware_bin
+    }
+
+    pub fn run_firmware_upgrade(&mut self) -> impl Iterator<Item = anyhow::Result<f32>> + '_ {
+        let n_chunks = ceil_div(
+            self.firmware_bin.bin.len() as u32,
+            FIRMWARE_UPGRADE_CHUNK_LEN,
```
nickfarrow, line 572: On the device we're checking the partition size is divisible by this chunk length.
If i recall correctly, our firmware size is going to be equal to the partition size.
So would this binary also be exactly divisible by this chunk length, rather than ceil_div?

frostsnap_coordinator/src/usb_serial_manager.rs
```diff
@@ -536,6 +554,92 @@ impl UsbSerialManager {
...
+                        error = e.to_string(),
+                        "writing firmware failed"
+                    );
+                    return Some(Err(e.into()));
+                }
+                let mut byte = [0u8; 1];
+
+                match io.raw_read(&mut byte[..]) {
+                    Ok(_) => {
+                        if byte[0] != FIRMWARE_NEXT_CHUNK_READY_SIGNAL {
```
nickfarrow, line 621: Do we ever get anything else written upstream other than this signal?

nickfarrow:
Any reason in particular to force users to update device firmware upon new keygens?

We do want people maximally up-to-date early on as we rapidly change things.
I am not sure forced updates * feel * right, since we want to emphasise the independence between coordinator and devices, but I don't think this presents any security risk given the devices are going to be verifying firmware signatures.

nickfarrow:
If devices send messages at the same time, the upstream device will fail to read the downstream message, disconnecting.

Maybe this is what you encountered when signing failed? Confirming on both devices at the same time

frostsnap_coordinator/src/usb_serial_manager.rs
LLFourn: Yes. We don't pad it to partition length on the wire.

frostsnap_coordinator/src/usb_serial_manager.rs
LLFourn: not while updating of course.

nickfarrow (approved):

LLFourn:
See: afe1bd8877b1884b0c655f859bde613f1cf9af94 instead of [d57ae70](https://github.com/frostsnap/frostsnap/pull/132/commits/d57ae70209f00a958245da6476269e37d39c8c98)

We shouldn't be modifying the UI based on receiving/sending coord messages. We should do it only ToUser messages. We don't follow that rule right now with the busy task stuff but we probably should. I just put it back to the way it was and fixed the bug. This also brings back the "busy" screen which I don't think we were seeing before (but this was probably fine!).

PR #133, opened 2024-05-28
Give keys names

nickfarrow:
- When initiating keygen, the coordinator sends over a `CoordinatorSendBody::KeyName(key_name)` to the device
- Currently the names are stored inside each `CoordinatorFrostKey`
- Devices receive this name and store it as a change after confirming keygen
- The device panics if it receives a new `Change::KeyName` message for a key.

Still not super happy with the mix of where key names are involved. With recent changes i've put it more inside frostsnap_core, no longer carrying around some `pending_key_name` inside the coordinator state.

I think we should either:
* Go all out into making a separate sql table for key names, removing it as much as possible from frostsnap_core
* Go all out into putting key names into keygen messages (might be the nicer solution). FinishKeyGen would tell the device the name, rather than storing some pending one while awaiting keygen.

### DIFF
--- a/device/src/bin/v2.rs
+++ b/device/src/bin/v2.rs
@@ -387,11 +387,20 @@ where
                             self.display.print(format!("Sign nostr: {message}"))
                         }
                     },
-                    Prompt::KeyGen(session_hash) => self.display.show_keygen_check(&format!(
-                        "{} {}",
-                        hex::encode(&session_hash[0..2]),
-                        hex::encode(&session_hash[2..4])
-                    )),
+                    Prompt::KeyGen {
+                        session_hash,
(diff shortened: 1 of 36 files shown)


nickfarrow:
I think this would be better if the device was informed of a pending keyname prior to keygen, name can be shown during keygen.

LLFourn:
- I simplified the key generation in core so we don't add any new messages 
- Changed the UI so that key generation is a multiple step process.
- Tried to improve things here and there.

nickfarrow (commented):
Works great, ignoring minor UI nits for now.

Please approve since I can't on my own PR

LLFourn (approved):

PR #136, opened 2024-06-25
Device Keyboard

nickfarrow:
Keyboard
- Built on @musdom 's jog keyboard.
- `Keyboard` lives inside `FrostyUi`.
- Has separate render and poll functions to fit into UI workflows.
- Indicates validity on the device after 59 backup characters.
- [ ] I havent quite managed to get highlighted keys working upon initial touch

Check backup
- Minimal workflow to make the keyboard useful.
- Coordinator suggests hrp index for the backup.
- Deeply tied to `frostsnap_core` as the basis for a restoration protocol.
- Device sends `g!(secret_share * G)` to the coordinator, who checks that the secret share belongs to this key.


nickfarrow:
Bunch of temp commits here for testing & exploration. Will drop them all except for keyboard changes once we're happy.

You can find "Restore" under the device settings.

device/src/keyboard.rs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        // skip the first rows to only show the end 12 chunks
+        let rows_to_skip = if chunked_backup.len() <= 12 {
+            0
+        } else {
+            (chunked_backup.len() - 1) / 3 - 3
+        };
+
+        for (i, chunk) in chunked_backup[(rows_to_skip * 3)..].into_iter().enumerate() {
```
musdom, line 180: Instead of drawing a textbox for each chunk, you can just add a space character after every 4 chars and pass it to one textbox

https://github.com/frostsnap/esp32-bsc/blob/cf6159c4dcebe067672d2b051e3cc29ef047bf53/frostsnap-graphics/src/lib.rs#L18

LLFourn:
I took a test run of this. I haven't reviewed the app side. The device code of the device side looks well arranged. Here's my UX comments trying to record and enter a few backups:

1. The keyboard state should be in the workflow.
2. It seems like it would be much more efficient to have a larger 4 row keyboard. I don't need to see several rows of what I've already entered. The last four characters would be enough in a single row at the top. This would allow for four rows of characters.
3. The share index should be entered separately and not assumed.
4. The `frost[X]` shouldn't be displayed while entering the backup.

A general comment is that I think we should use capital letters, not start with the `1` which is confusing because you can also have lowercase `l`. The `1` can be added by us internally before parsing.

device/src/keyboard.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+}
+
+#[derive(Default)]
+pub struct Keyboard {
+    buffer: Vec<char>,
+    hrp: Option<String>,
+    last_touch: Option<Instant<u64, 1, 1_000_000>>,
+    touched_key: Option<KeyboardKey>,
+    keyboard_keys: Vec<Vec<KeyboardKey>>,
```
LLFourn, line 70: I don't think we want this in the state of the keyboard. It seems like it could be a const. Furthermore I think you can figure out the rectangle and character just from the key's index or even just from the character itself in a functional way.

midizzz:
might be a bit late but I have been using the meshtastic keyboard on their t-watch (which I believe is the same screen setup) it looks like this, it has haptic and it seems like a good layout compromise for their purpose, though it is still to small in my opinion for frequent messaging. I hope this is helpful
IMG_3098

nickfarrow:
We could show the `FROST[x]` inside the framebuffer once on the entering share stage.

We used to have the ability to return to the entering HRP page by pressing backspace with an empty framebuffer, I kind of liked this feature but maybe it is excessive functionality.

@midizzz we're going to looking into some other keyboard designs in a subsequent PR. Our keyboard is likely too small for a full qwerty keyboard but we have some other ideas like nokia style or expanding grouped keys.

LLFourn:
> We could show the `FROST[x]` inside the framebuffer once on the entering share stage.

I don't think so. There's no space for that?
 
> We used to have the ability to return to the entering HRP page by pressing backspace with an empty framebuffer, I kind of liked this feature but maybe it is excessive functionality.

I think we shouldn't even enter it on the device since it's public. You never want to edit this. You do want it displayed on the final confirm page on the device. We should look at this again when we make a more step by step workflow for restoring the backup. This along with printing out backup cards. We might decide not to have `FROST[..]` anywhere at all.

device/src/graphics/widgets/enter_share_screen.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub fn is_finished(&self) -> bool {
+        self.backup_input_preview.is_finished()
+    }
+
+    pub fn try_create_share(&self, share_index: u16) -> Result<SecretShare, String> {
+        assert!(self.is_finished(), "must be finished to take share");
+        let characters = self.backup_input_preview.get_input();
+        let backup_string = format!("frost[{}]1{}", share_index, characters.to_lowercase());
+
+        SecretShare::from_bech32_backup(&backup_string).map_err(|_e| backup_string)
```
nickfarrow, line 102: More useful to map the error to_string for debugging

device/src/graphics/widgets/enter_share_screen.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    pub fn is_share_valid(&self) -> bool {
+        if !self.is_finished() {
+            return false;
+        }
+        self.try_create_share(42 /* XXX: doesn't matter to check */)
```
nickfarrow, line 109: I think this does matter for the checksum, if I step through it slowly an error is thrown here after typing in the correct share. One time I think I saw a flash of red characters after entering a valid share, the other call of `try_create_share` with the correct share index means the workflow progresses.

device/src/graphics/widgets/enter_share_screen.rs
LLFourn: I don't think so. We're not debugging anything here. We're interacting with the user. The only explanation is the user did it wrong at this point.

device/src/graphics/widgets/enter_share_screen.rs
LLFourn: We definitely should not be doing this hack but before removing it we owe it to ourselves to figure out why it works at all :rofl:

LLFourn: in summary: @nickfarrow was spot on.

nickfarrow:
Reviewed and tested!

Do you think it is worth taking the time to make our existing UI elements widget style? Or only as needed?

Please approve (I can't approve own pr) after checking 5a36deff4437dc0532c6dec371ac68ca3a7c48a6 makes sense. The coordinator loop thread finishing keygen UI protocol was racing against the backup protocol beginning.

LLFourn (approved):

PR #138, opened 2024-07-17
Migrate to Sqlite

LLFourn:
Changes the database backend to sqlite. Along with this:

1. Re-artchitect where device names are stored.
2. Create some common interfaces for persisting data. This is a bit half baked but it's better than how things were.
3. Re-architect frostsnap_core internal coordinator mutations to achieve atomicity between in-memory changes and persisted changes (fixed a nonce bug).

on top of #129


frostsnap_coordinator/src/frostsnap_persist.rs
```diff
@@ -0,0 +1,210 @@
...
+#[derive(Default)]
+pub struct DeviceNames {
+    names: HashMap<DeviceId, String>,
+    mutations: VecDeque<(DeviceId, String)>,
+}
+
+impl DeviceNames {
+    pub fn insert(&mut self, device_id: DeviceId, name: String) {
+        self.names.insert(device_id, name.clone());
+        self.mutations.push_back((device_id, name));
```
nickfarrow, line 146: Should only add a mutation if the name has changed

LLFourn: fixed: 5976023668cc920a94e36a3ac9afc2d04aebb10a

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -487,22 +349,21 @@ impl FrostsnapWallet {
...
+                Ok(tx_graph
+                    .index
+                    .next_unused_spk(&(key_id, AppAccountKeychain::internal()))
+                    .expect(
+                        "this should have been initialzed by now since we are spending from it",
+                    ))
+            })?;
+
+            self.tx_graph
+                .MUTATE_NO_PERSIST()
```
nickfarrow, line 362: Curious why MUTATE_NO_PERSIST here?

Because we don't want to mark as used until later (tx broadcasted / chain synced)?

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -330,37 +210,19 @@ impl FrostsnapWallet {
...
-        }
-
-        self.consume_spk_revelation_change(tx, changeset.indexer)?;
-
-        Ok(())
+        let mut db = self.db.lock().unwrap();
+
+        let changed =
+            self.tx_graph
+                .multi(&mut self.chain)
```
nickfarrow, line 216: These `Multi` types are so we can mutate tuple tx_graph and chain within the same db call?

frostsnap_coordinator/src/bitcoin/wallet.rs
LLFourn: good question. It demonstrates why this `MUTATE_NO_PERSIST` has to be fixed long term. The answer is that `mark_used` doesn't persist anything but how are you meant to know that.

frostsnap_coordinator/src/bitcoin/wallet.rs
LLFourn: Yes. This will be gone in the future hopefully.

frostsnapp/native/src/coordinator.rs
```diff
@@ -9,97 +8,81 @@ use frostsnap_coordinator::firmware_upgrade::{
...
-            .context("initializing db")?;
+        let mut db_ = db.lock().unwrap();
+
+        let coordinator = Persisted::<FrostCoordinator>::new(&mut db_, ())?;
+        let device_names = Persisted::<DeviceNames>::new(&mut db_, ())?;
+        let signing_session = Persisted::<Option<SigningSessionState>>::new(&mut db_, ())?;
 
         let usb_sender = usb_manager.usb_sender();
 
         // HACK: if the global device list depends on db state then it shouldn't be global! The
```
nickfarrow, line 62: nit: This is solved now? Can remove this comment

LLFourn: Thanks removed.

nickfarrow (approved):

PR #141, opened 2024-08-05
Alpha device UI

musdom:
- Simple black header
- Custom color palette instead of Rgb565 default colors
- Just confirm button, no borders
- Confirm bar back on top
- Faster display flush 150ms > 50ms


device/src/lib.rs
```diff
@@ -5,11 +5,11 @@ extern crate alloc;
 
 pub mod device_config;
 pub mod esp32_run;
+#[cfg(feature = "v2")]
+pub mod graphics;
```
LLFourn, line 9: :pick: I actually liked it being `st7789` since the graphics is tightly bound to the screen dimensions. If we were to make a second edition with a different screen size etc where would we put it?

LLFourn (approved):
approved and merged despite nit.

device/src/lib.rs
musdom: We could pass the mipidsi Model and implement Graphics for a different display model

PR #142, opened 2024-08-14
Chilldkg

LLFourn:
This uses my branch over here: https://github.com/LLFourn/secp256kfun/pull/192

As well as shifting to a new DKG I also changed the keygen check to a 4 byte check rather than the full 32 since I am confident I know how to get rid of that in the future. So the UI etc can be built around this.

EDIT the small check was moved into its own PR https://github.com/frostsnap/frostsnap/pull/143


nickfarrow (approved):
Reviewed and tested. API looks great and satisfying to see it fit nicely into our existing code!

PR #146, opened 2024-08-23
App Welcome Screen and Feedback Form

nickfarrow:
Small welcome screen upon app launch.

Feedback menu item on Key List page (tmp).

Feedback page form submits to https://feedback.frostsnap.com

See
https://github.com/frostsnap/feedback-server


frostsnapp/lib/feedback.dart
```diff
@@ -0,0 +1,136 @@
...
+      });
+      return;
+    }
+
+    setState(() {
+      _isLoading = true;
+      _resultMessage = '';
+    });
+
+    final token = api.randomSecpPointForToken();
```
LLFourn, line 49: what is this token doing and why does it need to be a point?

frostsnapp/lib/main.dart
```diff
@@ -196,7 +197,23 @@ class _MyHomePageState extends State<MyHomePage> {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-        appBar: AppBar(title: Text(widget.title)),
+        appBar: AppBar(title: Text(widget.title), actions: [
```
LLFourn, line 200: This should not just be on the home page. It should be on every page. Make a custom appbar widget which always has the action set and passes the widget title to the feedback form submission as well so we know where they were when they clicked it.

frostsnapp/lib/feedback.dart
```diff
@@ -0,0 +1,136 @@
...
+    final url = Uri.parse(
+        'https://feedback.frostsnap.com/submit?feedback=$feedback&token=$token');
```
LLFourn, line 51: What is feedback.frostnsp.com?

frostsnapp/lib/feedback.dart
```diff
@@ -0,0 +1,136 @@
...
+          style: TextStyle(fontSize: 16),
+        ),
+        const SizedBox(height: 8),
+        Divider(),
+        const SizedBox(height: 8),
+        Text('If you could have any feature right now, what would it be?'),
+        const SizedBox(height: 8),
+        Text('What is frustrating you about the app or devices?'),
+        const SizedBox(height: 8),
+        Text('Other thoughts/ideas/suggestions?'),
```
LLFourn, line 96: Name and email or some way to contact them?

It would also be good if they could attach a screenshot. It should be straightforward to check the clipboard for an image and attach it if they press "attach screenshot".

frostsnapp/lib/feedback.dart
```diff
@@ -0,0 +1,136 @@
...
+    setState(() {
+      _isLoading = false;
+      if (response.statusCode == 200) {
+        _resultMessage = 'Thank you!!';
+        _resultSuccess = true;
+        _feedbackController.clear();
+      } else {
+        _resultSuccess = false;
+        _resultMessage =
+            'Failed to submit feedback. Please try again.\n${response.body}';
```
LLFourn, line 63: Please try again "later".

frostsnapp/lib/main.dart
```diff
@@ -196,7 +197,23 @@ class _MyHomePageState extends State<MyHomePage> {
...
+          PopupMenuButton<String>(
+            onSelected: (String result) {
+              if (result == 'feedback') {
+                Navigator.push(context, MaterialPageRoute(builder: (context) {
+                  return FeedbackPage();
```
LLFourn, line 205: I wonder if an alert dialog would make more sense here. Once you submit you can dismiss it. With the page you can do this too but it currently does it and it's less idiomatic.

LLFourn (commented):
Great idea to implement this. My weakly held feeling is that the welcome alert should be a screen like:

image

(of course you need to decide how this looks if they already have a wallet etc).

And then the feedback page should actually be an alert.

frostsnapp/lib/feedback.dart
```diff
@@ -0,0 +1,136 @@
...
+        ),
+        SizedBox(height: 20),
+        Text(
+          'Thanks for taking part in the future of Bitcoin self-custody.',
+          style: TextStyle(fontSize: 16),
+        ),
+        const SizedBox(height: 8),
+        Divider(),
+        const SizedBox(height: 8),
+        Text('If you could have any feature right now, what would it be?'),
```
LLFourn, line 92: Someone who is on a feedback form already knows why they are there so I feel the text is cluttering up things. It would be good to prompt them for feedback right after they've done their first key generation, with a small dialog that explaining where the feedback form is and taking them here if they're ready.

midizzz:
This form is missing fields under each of the questions at the moment.

PR #147, opened 2024-08-27
[app] Fix device SVG dimensions

LLFourn:
Adam made the SVG more realistic and then merged the PR but this broke a bunch of things.
This uses the same SVG fixes most of the broken things.


nickfarrow (approved):

nickfarrow:
Tested. Approved

PR #148, opened 2024-08-27
Create appbar shadow

LLFourn:
Makes it easier to tell where the content is.
Also make dividers actually show.


nickfarrow (approved):
Tested

nickfarrow (approved):

PR #149, opened 2024-08-27
Improve device action dialogs

LLFourn:
- Stop using a DraggableScrollableSheet since this is more complicated
than what we need and gets in the way of some gestures.
- Create DialogHeader which makes the formatting of the dialogs more consistent
- Device lists are always scrollable now and they show the scrollbar if they're long enough always.

### DIFF
--- a/frostsnapp/lib/theme.dart
+++ b/frostsnapp/lib/theme.dart
@@ -6,6 +6,7 @@ const Color successColor = Color.fromARGB(255, 21, 255, 0);
 const Color awaitingColor = Color.fromARGB(255, 207, 124, 15);
 const Color uninterestedColor = Color.fromARGB(255, 167, 160, 160);
 const Color errorColor = Color.fromARGB(255, 172, 23, 23);
+const Color shadowColor = Colors.black26;
 const Color backgroundPrimaryColor = Color.fromARGB(255, 60, 107, 134);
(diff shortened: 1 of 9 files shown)


LLFourn:
Added 6be2a228d1ba3a14b747c534648f79abe19faee8 to clear up some confusion I was having in later work.

frostsnapp/lib/theme.dart
```diff
@@ -39,15 +41,15 @@ final ThemeData frostsnappTheme = ThemeData(
   iconTheme: IconThemeData(color: textSecondaryColor),
   checkboxTheme: CheckboxThemeData(
     checkColor:
-        MaterialStateProperty.resolveWith<Color>((Set<MaterialState> states) {
-      if (states.contains(MaterialState.selected)) {
+        WidgetStateProperty.resolveWith<Color>((Set<WidgetState> states) {
```
nickfarrow, line 44: Note: upgrade to recent flutter version.

After `flutter upgrade` some of the background colours stopped rendering so I fixed them in 383703cb0a0a8ed5faece81e76cb4b503dcc4823

nickfarrow (dismissed):

nickfarrow (dismissed):

nickfarrow (approved):

PR #153, opened 2024-09-11
Add touchgrid to debug touch screen

LLFourn:
Not sure if we want to merge this but here's the cst debugger screen I made.

To use:

```rust
just erase otadata;
cargo run --release --bin touchgrid
```


LLFourn:
Note this is not intended to be merged. I'll leave the branch on the repo though.

PR #154, opened 2024-09-11
Verify Address on Device

nickfarrow:


midizzz (commented):
I scrolled the whole thing... Looks tidy, especially the Rust. Inspiring me to get into it, C, Dart, not so much. There is a long way to go before I understand any of it though. Nice work!

LLFourn:
1. I don't think the address QR code needs to be in a dialog. I'd put them in a dialog only when they click verify or something. 
2. The QR code should have a white background for max contrast. I think you can do that as part of the QR code style.
3. I don't see the point of confirming on the device here. Confirming on the device is to make sure the coordinator is not able to instruct the device to do something the user doesn't want. But here it's the user who is taking action not the device. I think the device should just show the address.
4. The device should show the derivation path I think.
5. The device should highlight 3 chunks not just 1 I think.
6. It's nice to show the list of devices you could plug in but if it's in a dialog I guess it has to be scrollable.

I'm still thinking about the best design here.

LLFourn:
I rebased this on current master.

This is looking really good. I've got a list of small improvements:

1. The copy paste snackbar covers the verify button
2. I'm not sure if `done` should go all the way back after using this design. After you click verify you may realize that you can't copy paste the address on this screen and click done to go back. Maybe only show done if they've plugged in an device and otherwise a cancel button is available which just brings you back.
3. The address is testnet for me on the device XD
4. Maybe instead of a copy paste button below everything by itself have a show QR code button -- most people don't use QR codes for transmitting bitcoin addresses.
5. If you're using sparrow you might need a way to verify an address by inputting the derivation index from sparrow somewhere.

frostsnap_coordinator/src/verify_address.rs
```diff
@@ -0,0 +1,131 @@
...
+    key_id: KeyId,
+    derivation_index: u32,
+    is_complete: Option<Completion>,
+    need_to_send_to: BTreeSet<DeviceId>,
+    sink: Box<dyn Sink<VerifyAddressProtocolState>>,
+}
+
+impl VerifyAddressProtocol {
+    pub fn new(
+        verify_address_messages: &mut Vec<CoordinatorSend>,
```
LLFourn, line 29: The modern way to do this is to pass a `&FrostCoordinator` in here and call `.verify_address` on it directly. To improve on this it would be more ergonomic to return some kind of `VerifyAddress` which impls `Into<Vec<CoordinatorSend>>` or something like that. This way yo can have an easy time getting hold of the details for this but can also turn it into a `CoordiantorSend` form. This improvement could also be done to `start_sign` which looks where you got this approach from.

frostsnap_coordinator/src/verify_address.rs
```diff
@@ -0,0 +1,131 @@
...
+            target_destinations: Destination::Particular(core::mem::take(
+                &mut self.need_to_send_to,
+            )),
+            message_body: CoordinatorSendBody::Core(CoordinatorToDeviceMessage::VerifyAddress {
+                key_id: self.key_id,
+                derivation_index: self.derivation_index,
+            }),
+        };
+
+        self.emit_state();
```
LLFourn, line 120: why emit state every time you're polled?

frostsnap_coordinator/src/verify_address.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    fn is_complete(&self) -> Option<Completion> {
+        self.is_complete.clone()
+    }
+
+    fn connected(&mut self, id: frostsnap_core::DeviceId) {
+        if self.state.target_devices.contains(&id) {
+            self.need_to_send_to.insert(id);
+        }
+        self.emit_state()
```
LLFourn, line 83: shouldn't you only emit state when something relevant is connected?

frostsnap_coordinator/src/verify_address.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    fn disconnected(&mut self, device_id: frostsnap_core::DeviceId) {
+        self.need_to_send_to.remove(&device_id);
```
LLFourn, line 87: shouldn't you emit state here when something relevant is disconnected?

LLFourn (commented):
Did some code review.

I think we want to merge #157 first which will necessitate some refactoring on this one.

device/src/esp32_run.rs
```diff
@@ -587,6 +587,22 @@ where
                                     ui::Prompt::ConfirmLoadBackup(share_backup),
                                 ));
                             }
+                            DeviceToUserMessage::VerifyAddress { spk, bip32_path } => {
+                                let address =
+                                    bitcoin::Address::from_script(&spk.spk(), BITCOIN_NETWORK)
```
nickfarrow, line 592: Little  hacky passing the spk around instead of the address so we can use the const easily here.

frostsnap_coordinator/src/verify_address.rs
nickfarrow: Instead of passing in the coordinator I decided to just return a `VerifyAddress` from `coordinator.verify_address(...)`, which implements `IntoIterator` since we're always calling `into_iter` anyway.

I was also able to make `coordinator.verify_address` just take `&self`, not mutating anything.

Check it out and lmk what you think.

frostsnap_core/src/coordinator.rs
```diff
@@ -872,6 +875,42 @@ impl FrostCoordinator {
         }])
     }
 
+    pub fn verify_address(
+        &self,
+        access_structure_ref: AccessStructureRef,
+        derivation_index: u32,
+        encryption_key: SymmetricKey,
```
LLFourn, line 882: We shouldn't need the decryption key to verify an address. You should send the master app key over, the device can verify it's correct with the `KeyId`, and derive the address from the `MasterAppkey`.

LLFourn (changes requested):
Nice work. The address searching bonus feature is especially awesome. We need to make it more discoverable. There's a low level change you need to make here. You shouldn't need to put in your pin to get an address.

LLFourn (approved):
ACK 5a1e67860d93e413e964974b23b7ddcb3bb9a626

PR #155, opened 2024-09-16
Bundle App for Desktop and Upload in CI

nickfarrow:
App build matrix:
- Build AppImage for linux
    - runs on arch linux 
    - runs on ubuntu-18 in a docker container (no graphical display but looked good)
- APK and app bundle (needed for android stores)

These artifacts get uploaded during the action.

Builds upon https://github.com/frostsnap/frostsnap/pull/165


.github/workflows/test.yml
```diff
@@ -78,9 +113,12 @@ jobs:
...
+      - run: sudo apt-get update && sudo apt install -y libudev-dev
       - run: just frostsnapp/install-rust-deps
-      - run: just frostsnapp/build linux
-      - run: just frostsnapp/build appbundle
+      - run: just frostsnapp/build appbundle --release
+      - name: Upload APK
+        uses: actions/upload-artifact@v4
+        with:
+          name: frostsnapp-${{ github.sha }}.apk
+          path: target/riscv32imc-unknown-none-elf/release/v2
```
LLFourn, line 123: This is called `Upload APK` but it uploads an `riscv32imc` elf binary?

LLFourn:
@LLFourn to check out whether MacOS works.

LLFourn:
Let's be slightly less ambitious and just publish android and linux for now. We can try to iron out windows, mac and iPhone later.

nickfarrow:
TODOs:
- [ ] fix appimage size
- [ ] check without appimage
- [ ] check if we need canberra-gtk

.github/workflows/test.yml
```diff
@@ -375,11 +373,12 @@ jobs:
         with:
           path: ~/.pub-cache/
           key: ${{ runner.os }}-pub-cache-${{ hashFiles('frostsnapp/pubspec.lock') }}-v1
+          restore-keys: |
+            ${{ runner.os }}-pub-cache-
 
       - uses: subosito/flutter-action@v2
         with:
           channel: stable
-          cache: true
```
nickfarrow, line 318: i think this looks for **/pubspec.lock, which doesn't work? Will revisit tomorrow
https://github.com/frostsnap/frostsnap/actions/runs/11666583441/job/32481995066

.github/workflows/test.yml
```diff
@@ -259,35 +279,55 @@ jobs:
...
+            libcairo2-dev \
+            python3-pip \
+            git \
+            wget \
+            libudev-dev \
+            qtbase5-dev \
+            qt5-qmake \
+            qtdeclarative5-dev \
+            cmake \
+            libarchive-tools
```
LLFourn, line 329: Why do we have all these dependencies now? Also wget and get? I think they are already installed. This seems to be something to do with app image stuff? (Although I don't really understand why even app images would need all these qt and libcairo etc). If so it should be done in another step. The app image just needs the flutter linux build artifact right? We should build that in one step and package the app image in a different workflow IMO.


Also can we have a `just` command to build the app image and have the CI call the just command.

.github/workflows/test.yml
LLFourn: we can just commit `pubspec.lock`. It would be better to use the flutter action caching approach.

nickfarrow: Was already commited. I just undid this change and it magically worked. I also deleted all the caches, ran again, and it still worked.

LLFourn (approved):
self-ACK. @nickfarrow do the honors of merging if you're good with this.

PR #156, opened 2024-09-18
View logs in app

nickfarrow:


frostsnapp/native/src/api.rs
```diff
@@ -264,9 +291,11 @@ pub fn turn_stderr_logging_on(level: LogLevel) {
...
-        .finish();
+        .finish()
+        .with(StreamLogsToDart);
+
     let _ = tracing::subscriber::set_global_default(subscriber);
-    event!(Level::INFO, "logging to stderr");
+    event!(Level::INFO, "logging to stderr and Dart logger");
 }
 
 pub fn turn_logcat_logging_on(_level: LogLevel) {
```
LLFourn, line 301: I think this should return the Stream of log rather than there being a separate API for this.

LLFourn:
I pushed a commit.

b4bce7e33bc1ed799c56991da1e1711e75af2715

I want to extend it so that the copy paste log appears also in the error startup screen.

LLFourn:
Take a look at the new commits. I've changed a bit:

1. Publush the logs on the STARTUP ERROR page
2. You can log from dart now
3. Make it just a Stream<String> rather than Stream<LogEntry> so we centralize all logging logic in rust.

nickfarrow:
Works great!

Only suggestion would be to leave the time `2024-10-04T08:09:53.041097Z` as just unix time, or shorten as something else?

LLFourn (approved):
I removed the local timezone from the logs so we don't learn the user's timezone if they send us there logs to report a bug.

PR #165, opened 2024-10-24
Do not commit bridge generated files

nickfarrow:
Completes https://github.com/frostsnap/frostsnap/issues/159


.github/workflows/test.yml
```diff
@@ -29,45 +28,93 @@ jobs:
...
+      - uses: actions/download-artifact@v4
+        with:
+          name: frostsnap-${{ github.sha }}.elf
+          path: target/riscv32imc-unknown-none-elf/release/
+      - uses: subosito/flutter-action@v2
+        with:
+          channel: "stable"
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
```
LLFourn, line 55: don't think we need clippy here.

.github/workflows/test.yml
```diff
@@ -29,45 +28,93 @@ jobs:
...
+      - name: Restore generated files
+        uses: actions/cache/restore@v3
+        with:
+          path: ./frostsnapp
+          key: bridge-files-${{ github.sha }}
+          fail-on-cache-miss: true
+      - uses: actions/download-artifact@v4
         with:
-          name: frostsnap-v2-${{ github.run_id }}.elf
+          name: frostsnap-${{ github.sha }}.elf
```
LLFourn, line 112: why is this changing and what does it mean?

.github/workflows/test.yml
```diff
@@ -79,41 +126,30 @@ jobs:
...
-      - name: Show riscv release directory
-        run: ls -R target/riscv32imc-unknown-none-elf/release/
       - run: sudo apt-get update && sudo apt install -y ninja-build libstdc++-12-dev libgtk-3-0 libgtk-3-dev libudev-dev
-      - name: Cache Flutter dependencies
-        uses: actions/cache@v3
-        with:
-          path: ~/.pub-cache
-          key: ${{ runner.os }}-flutter-${{ hashFiles('**/pubspec.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-flutter-
```
LLFourn, line 91: why remove this?

.github/workflows/test.yml
```diff
@@ -79,41 +126,30 @@ jobs:
...
-      - name: Install flutter dependencies
-        run: flutter pub get
+      - run: sudo apt-get update && sudo apt-get install -y pkg-config libudev-dev
```
LLFourn, line 151: why do we need these now when we didn't need them before?

.github/workflows/test.yml
nickfarrow: Discussed, the commit sha is more useful for distinguishing artifacts than runner IDs.

.github/workflows/test.yml
nickfarrow: Oops, added this back and did some more crazy caching for the rust dependencies.

.github/workflows/test.yml
```diff
@@ -7,7 +7,43 @@ on:
...
+          restore-keys: |
+            ${{ runner.os }}-cargo-registry-
+
+      # Restore target cache
+      - uses: actions/cache@v4
+        with:
+          path: target/
+          key: ${{ runner.os }}-target-build-rust-deps-${{ hashFiles('./Cargo.lock') }}-v1
+          restore-keys: |
+            ${{ runner.os }}-target-build-rust-deps-
```
LLFourn, line 30: we're not building anything to `target/` here though as far as I can tell.

.github/workflows/test.yml
```diff
@@ -16,104 +52,237 @@ jobs:
...
+            autoconf automake autotools-dev curl python3 libmpc-dev \
+            libmpfr-dev libgmp-dev gawk build-essential bison flex \
+            texinfo gperf libtool patchutils bc zlib1g-dev \
+            libexpat-dev ninja-build
+
       - name: Setup | RISCV GCC
+        if: steps.cache-riscv.outputs.cache-hit != 'true'
         run: |
           wget -nv https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/2023.05.24/riscv32-elf-ubuntu-22.04-nightly-2023.05.24-nightly.tar.gz
           tar -zxf riscv32-elf-ubuntu-22.04-nightly-2023.05.24-nightly.tar.gz
```
LLFourn, line 78: PS why are we using some random nightly riscv toolchian?

.github/workflows/test.yml
```diff
@@ -16,104 +52,237 @@ jobs:
...
+      - uses: taiki-e/install-action@just
+      - run: just frostsnapp/install-rust-deps
+      - run: just frostsnapp/install-flutter-deps
+      - run: just gen
+
+      - name: Upload Bridge Generated Files
+        uses: actions/upload-artifact@v4
+        with:
+          name: bridge-generated-files
+          path: ./frostsnapp
```
LLFourn, line 153: What's the thinking behind making the bridge files an artifact -- also why is the whole frostsnapp directory here rather than just the `bridge_` files. I think you can just use the cache again and use this glob:

```
frostsnapp/lib/bridge_*.dart
frostsnapp/native/src/bridge_*.rs
```

and cache it on the contents of `frostsnap/native/src/*`

frostsnapp/justfile
```diff
@@ -29,11 +28,26 @@ flutter_rust_bridge_version:
...
 install-rust-deps:
     #!/bin/sh
     flutter_rust_bridge_version=$(just flutter_rust_bridge_version)
     if test "$flutter_rust_bridge_version"; then
-        cargo install "flutter_rust_bridge_codegen@$flutter_rust_bridge_version" cargo-ndk espflash
+        # Check if binaries are already cached, skipping installation if they exist
+        if [ ! -x ~/.cargo/bin/flutter_rust_bridge_codegen ] || \
+           [ ! -x ~/.cargo/bin/cargo-ndk ] || \
+           [ ! -x ~/.cargo/bin/espflash ] || \
+           [ ! -x ~/.cargo/bin/cargo-expand ]; then
```
LLFourn, line 43: let's just let cargo figure this out!

LLFourn (approved):

PR #171, opened 2024-11-13
Use BDK-maintained SQLite

evanlinjin:
This allows us to persist floating txouts and has backwards compatibility built in.


evanlinjin:
[Persist floating txouts.](https://trello.com/c/ZzmYGvq0/290-persist-floating-txouts)

LLFourn (approved):
Deletes code so ACK. Hopefully we can do https://github.com/bitcoindevkit/bdk/issues/1695 to make this less bitter.

PR #172, opened 2024-11-15
Permissions design

nickfarrow:


docs/permissions.md
```diff
@@ -0,0 +1,112 @@
...
+3. For a device to become useful with a new coordinator, the coordinator must become AWARE of a key of which the device holds a share.
+
+## Potential Annoyance/Pitfall
+
+One potential limitation with this design is that to create a new key across a set of devices with a new coordinator, the coordinator must first learn of a key belonging to **each** device.
+
+Example scenario: You have 5 Frostsnap devices and set up:
+
+- A 2-of-3 threshold with three devices
+- A 2-of-2 threshold with the other two devices
```
LLFourn, line 20: Not clear whether these both for the same key or different keys.

docs/permissions.md
```diff
@@ -0,0 +1,112 @@
...
+
+Key points:
+
+- Under trusted firmware, devices are less likely to collect information than phone coordinators
+- Keygen results in a new key which the coordinator is AWARE of
+- Once the coordinator reveals its awareness of keys, devices can disclose information about held shares:
+  - For AWARE keys: disclose index and share image
+  - For UNAWARE keys: provide a t-of-n key share and key fingerprint that progresses the coordinator towards becoming AWARE
+    - Coordinator must visit `t` devices to become AWARE
+    - Key names may be provided alongside shares
```
LLFourn, line 38: what is the different between a `t-of-n` key share and a `share image`.

docs/permissions.md
```diff
@@ -0,0 +1,112 @@
...
+| Device Configuration   | AWARE of ALL keys                   | Yes                    |
+
+## Action Details
+
+### Nonces
+
+Sharing nonces during coordinator recovery enables two-round signing.
+Though we do not want to allow unknown coordinators to request heaps of nonces under different identities to crash the devices.
+Is this limit on nonces during key restoration enough to prevent storage overflow?
+Perhaps the device could store any pending new-coordinator nonces in a separate location, only storing one at a time.
```
LLFourn, line 67: Why would the number of nonces requested matter? You don't store the nonces locally on the device.

I don't see why storage exhaustion would be a relevant attack since you can just destroy it.

docs/permissions.md
```diff
@@ -0,0 +1,112 @@
...
+This should require a long CPU countdown.
+Perhaps the timer to delete could scale with the number of shares held.
+
+## Device Configuration
+
+In the future, we might want a way to semi-permanently apply settings to a device, such as:
+
+- Disabled Actions
+- Speak to anyone mode
+- CPU timer multiplier
```
LLFourn, line 82: I'd guess that configuration is mostly to do with keys and not global. Other than name/appearance settings.

docs/permissions.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Other Ideas
+
+## Time Based Access
+
+If the devices have a concept of time (powered / unpowered) we could render a new coordinator trusted after N hours.
+
+## Sharable Coordinator Certificates
+
+Devices could sign certificates which get saved _elsewhere_ (cloud?), to be loaded into a new authenticator to make it trusted.
```
LLFourn, line 110: s/authenticator/coordinator/ ?

docs/permissions.md
```diff
@@ -0,0 +1,112 @@
...
+### Actions
+
+| Action                 | Access Requirement for Coordinator  | On-Device Confirmation |
+| ---------------------- | ----------------------------------- | ---------------------- |
+| KeyGen Start           | AWARE of any key (or no keys exist) | No                     |
+| KeyGen Finish          | AWARE (or no keys exist^^)          | Yes                    |
+| Share Public Nonces    | AWARE or UNAWARE (limited)          | No                     |
+| Signing                | AWARE of relevant key               | Yes                    |
+| Display Backup         | AWARE of relevant key               | Yes                    |
+| Check Backup           | AWARE                               | Yes (keyboard input)   |
```
LLFourn, line 51: of relevant key?

LLFourn (commented):
How would you demonstrate awareness?
Right now we follow this rule implicitly by for example, requiring the root key/decryption key contrib to be passed from the coordinator to device.

We could probably do something similar for some others. But what about `Rename device` for example. Of course using the `key_id` would be nice but we currently use that during recovery with `HeldShares` in #178. So I guess we'd have to come up with something else or take `key_id` out of there.

Overall it seems like to make sense. The rules seem to be:

1. Do anything with a particular key you must be aware of the key
2. To anything else you must be aware of any key
3. Unless you are destroying the device fully e.g. reset then you can just go ahead.

LLFourn:
I'm going to close this now as I think the rules here will be straightforward given what we are discussing about a 1-to-1 share to device mapping.

1. Any coordinator can generate a key share on an empty device
2. In order to do anything with the device after that you have to be "aware" of the key

@nickfarrow any thoughts?

PR #173, opened 2024-11-19
Fix firmware bundling again

LLFourn:


frostsnapp/lib/device_settings.dart
```diff
@@ -150,7 +150,9 @@ class _DeviceSettingsState extends State<DeviceSettings> {
                   fontSize: 16,
                 )),
             Expanded(
-                child: Text(coord.upgradeFirmwareDigest(),
+                child: Text(
+                    coord.upgradeFirmwareDigest() ??
+                        "<app compiled withotu firmware>",
```
evanlinjin, line 155:
```suggestion
                        "<app compiled without firmware>",
```

frostsnapp/native/src/api.rs
```diff
@@ -141,7 +141,7 @@ impl ConnectedDevice {
     }
 
     pub fn needs_firmware_upgrade(&self) -> SyncReturn<bool> {
-        SyncReturn(self.firmware_digest != self.latest_digest)
+        SyncReturn(Some(self.firmware_digest.as_str()) != self.latest_digest.as_deref())
```
evanlinjin, line 144:
```suggestion
        SyncReturn(match self.latest_digest {
            Some(latest_digest) => self.firmware_digest.as_str() != latest_digest.as_deref(),
            None => false,
        })
```

Wouldn't this make sense to always return `false` if we didn't compile with firmware?

LLFourn: Good question. I'll make a comment:

```
        // We still want to have this return true even when we don't have firmware in the app so we
        // know that the device needs a firmware upgrade (even if we can't give it to them).
```

evanlinjin (approved):
ACK e60c5ff7c96129761d702a83fa5355a50718937d

PR #176, opened 2024-11-21
Artifact upload action

LLFourn:
I dunno where these show up.


LLFourn:
This PR didn't really work but is worth merging anyway.

1. Cleans up the workflows to only run on important PR events
2. Uses `--locked` for most cargo things we can
3. Puts some prettier artifact download pages on the workflow page.

The next step is to upload the artifacts somewhere we control and make a comment on the PR so they can be downloaded by anyone without being logged in.

nickfarrow (approved):

PR #178, opened 2024-11-26
Basic wallet recovery

LLFourn:
1. Allows you to delete keys from the key list
2. Allows you to recover them by plugging in devices

UI is quite rough but

### DIFF
--- a/frostsnapp/native/src/coordinator.rs
+++ b/frostsnapp/native/src/coordinator.rs
@@ -342,86 +342,84 @@ impl FfiCoordinator {
                             usb_sender.send(send_message);
                         }
                         CoordinatorSend::ToUser(msg) => {
-                            if let Some(ui_protocol) = &mut *ui_protocol_loop {
-                                ui_protocol.process_to_user_message(msg);
-                            } else {
-                                match msg {
-                                    CoordinatorToUserMessage::PromptRecoverShare(recover_share) => {
-                                        let span = span!(
-                                            Level::INFO,
-                                            "recovering share",
-                                            from = recover_share.held_by.to_string(),
-                                            key_name = recover_share.held_share.key_name,
-                                            access_structure_ref = format!(
-                                                "{:?}",
-                                                recover_share.held_share.access_structure_ref
-                                            ),
+                            match msg {
+                                // there is no UI protocol for share recovery because it happens in the background.
+                                CoordinatorToUserMessage::PromptRecoverShare(recover_share) => {
+                                    let span = span!(
+                                        Level::INFO,
+                                        "recovering share",
+                                        from = recover_share.held_by.to_string(),
+                                        key_name = recover_share.held_share.key_name,
+                                        access_structure_ref = format!(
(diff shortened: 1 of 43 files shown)


frostsnapp/native/src/coordinator.rs
```diff
@@ -161,28 +175,54 @@ impl FfiCoordinator {
...
                                 if let Some(protocol) = &mut *ui_protocol_loop {
-                                    protocol.connected(*id);
+                                    protocol.connected(id);
+                                }
+
+                                if let Some(connected_device) = device_list.get_device(id) {
+                                    // we only send some messages out if the device has up to date firmware
+                                    if !connected_device.needs_firmware_upgrade().0 {
+                                        coordinator_outbox.extend(
+                                            coordinator.maybe_request_nonce_replenishment(id),
```
nickfarrow, line 194: Does this belong here/in this PR? - nonce replenishment has been in the code for a while, or is it to handle per-coordinator nonces?

nickfarrow:
If I restore delete and restore a key without deleting the entire database, the app remembers the network somehow. If I delete the DB, the network is assumed to be mainnet. I guess we are loading old `Mutation::SetWalletNetwork`s, maybe we want to avoid this behaviour, and only load key-related mutations after a NewKey?

frostsnapp/native/src/coordinator.rs
```diff
@@ -711,11 +828,107 @@ impl FfiCoordinator {
...
+        key_state(
+            &self.recoverable_keys.lock().unwrap(),
+            &self.coordinator.lock().unwrap(),
+        )
+    }
+
+    pub fn start_recovery(&self, key_id: KeyId) -> Result<()> {
+        let mut recoverable_keys = self.recoverable_keys.lock().unwrap();
+        let recover_shares_by_as = recoverable_keys
+            .range(AccessStructureRef::range_for_key(key_id))
```
nickfarrow, line 842: What was the thinking here? Seems to filter for all access structures

frostsnapp/native/src/coordinator.rs
LLFourn: Yes we shouldn't request nonces from devices unless they have update firmware. We should only send the messages needed to updated firmware.

LLFourn:
@nickfarrow don't we know the network of the key now? I guess we should just transmit that.

LLFourn:
@nickfarrow we didn't have the full network. We just had what kind of network. I changed this to the full network . 

> We need to know it on the device so the coordinator can recover it. Therefore it should be frostsnap_core responsibility and therefore we don't need to store it in an sql table or have a "settings" thing since it can't change (for now).

LLFourn:
@nickfarrow alright there's another separate commit that fixes a bug found by @evanlinjin on master. This PR felt like the easiest place to fix it since some of the parts for the fix I'd already developed as part of this PR.

Please review the last two commit separately.

nickfarrow (approved):
Works great, reviewed and tested recent commits

PR #181, opened 2024-12-06
Wipe device data

nickfarrow:


nickfarrow:
Ready for review. There is some slightly weird behaviour if you reset a device in a daisy chain - it panics the upstream device, and doesn't get removed from the device list.

LLFourn:
I don't get the panic. For me the coordinator just never realizes the latter device has reset since it doesn't get error bytes written to it directly. I think you can always just write magic bytes upstream before resetting to fix this -- a better way might be to add an explicit reset signal.

 It would be good if we had a proper modal to indicate you should be looking at the device screen when you need to confirm the data wipe but we can do that later.

frostsnapp/lib/device_settings.dart
```diff
@@ -47,14 +47,32 @@ class _DeviceSettingsState extends State<DeviceSettings> {
...
+            {
+              /* nothing needs to be done */
+            }
+          case DeviceListChangeKind.Named:
+            {
+              /* nothing needs to be done */
+            }
+        }
+      }
+
```
LLFourn, line 72: I think you can just do:

```
deviceListSubject.firstWhere((change) => change.kind == DeviceListChangeKind.Removed);
```

Or something!

LLFourn (changes requested):
Small dart code optimization otherwise this looks good.

frostsnapp/lib/device_settings.dart
nickfarrow: Fixed!

frostsnapp/native/src/api.rs
```diff
@@ -1038,6 +1038,11 @@ impl Coordinator {
     pub fn get_connected_device(&self, id: DeviceId) -> SyncReturn<Option<ConnectedDevice>> {
         SyncReturn(self.0.get_connected_device(id))
     }
+
+    pub fn wipe_device_data(&self, device_id: DeviceId) -> SyncReturn<()> {
```
LLFourn, line 1042: I think `SyncReturn<()>` is unnecessary. Is this thing fallible? It should probably return a Stream. I'm guessing this is just keeping it simple for now which is fine as who knows how this turns out after we do some redesigning.

LLFourn (approved):

PR #183, opened 2024-12-12
Conch communication

LLFourn:
Most important commit:

```
[comms] You can only send messages when you have the conch 🐚

This system makes it unlikely that a device can be busy doing something
else when a downstream device sends a messages upstream and therefore
lose the message. Only the device with the conch can send a message and
by implication you only need to be ready to receive downstream messages
if you have sent the conch downstream.

This is a breaking change.
```

Also I added message encapsulation so it will be easier to make non-breaking changes in the future. The idea is that we put a length so we can see the length of a message even if we can't decode it. That way we can ignore it if we don't understand it.


LLFourn:
I made this backwards compatible by using the last byte of the magic bytes as a kind of version communicator. Also bumped the number of nonces to 100.

device/src/esp32_run.rs
```diff
@@ -411,36 +324,228 @@ where
...
+                            upgrade = Some(upgrade_);
+                        }
+                        CoordinatorUpgradeMessage::EnterUpgradeMode => {}
+                    },
                 }
             }
 
-            if let Some(ui_event) = ui.poll() {
+            {
+                let staged_mutations = signer.staged_mutations();
```
nickfarrow, line 448: You accidentally duplicated a lot of code in this file

frostsnap_comms/src/lib.rs
```diff
@@ -45,13 +45,24 @@ pub const BINCODE_CONFIG: bincode::config::Configuration<
 pub enum ReceiveSerial<D: Direction> {
     MagicBytes(MagicBytes<D>),
     Message(D::RecvType),
+    Unused9,
```
nickfarrow, line 55: Very cool enums let you do this so simply

nickfarrow:
Rebased, and deleted code duplication in 509e72be8fa74727d5e46c75bd2e3a6d600de253

Need to check I did the right thing with DeviceDataWipe encapsulation

device/src/esp32_run.rs
LLFourn: Thanks for catching this!

nickfarrow:
When my device has a large number of held shares, I get message timeout errors . Is this unrelated to the conch design?
```
  DEBUG frostsnap_coordinator::usb_serial_manager: queueing message, RequestHeldShares
    at frostsnap_coordinator/src/usb_serial_manager.rs:497
    in frostsnap_coordinator::usb_serial_manager::send on port with port: "/dev/ttyACM0", gist: "RequestHeldShares"
    in frostsnap_coordinator::usb_serial_manager:: with destinations: "0357ce82f30a48e076c3a72b41f0b3bdcf29095fe92fdf4c00fdafa533e3dee3ed"
    in frostsnap_coordinator::usb_serial_manager::poll_ports
    at frostsnap_coordinator/src/
ERROR frostsnap_coordinator::usb_serial_manager: failed to read message from port, port: "/dev/ttyACM0", error: "Io { inner: Custom { kind: TimedOut, error: \"Operation timed out\" }, additional: 282 }
```

nickfarrow (approved):

PR #184, opened 2024-12-13
[native] Remove pending_for_outbox

LLFourn:
pending_for_outbox was no longer needed for anything but was being used to send messages to devices. These messages were always duplicates. This caused a panic when you had a device plugged in already before you went to sign.


nickfarrow (approved):

nickfarrow:
Ran `flutter upgrade` and tested, thanks for fixing!

PR #190, opened 2024-12-19
Retheme wallet page

evanlinjin:


frostsnapp/android/app/src/main/AndroidManifest.xml
```diff
@@ -3,12 +3,14 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-sdk android:minSdkVersion="12" />
 
+    <!-- llfourn removed allowBackup. Seems like we wouldn't want it. -->
+    <!-- evanlinjin added enableOnBackInvokedCallback -->
```
LLFourn, line 7: why though?

nickfarrow:
This rebase took quite a long time, squashing made it do-able. Reviewed during the rebase process, and now I also have a better understanding other recent PRs (particularly `SuperWallet` and context `wrap`). I think in future we should try communicate scope better to avoid overlapping changes. Separating out wallet send & receive dart files should be an additional PR to this theming PR.

A few things todo:
- Some of the colours don't contrast well, like the key-card & buttons, the colour scheme is great but probably just need different selections on buttons etc.
- Love the dark. Maybe background hue is slightly too green. Could explore a blue tint over a green. 
- Address Icon looks like it's going to open a link not view info - prefer previous/other icons.
- Device list SVGs are hard to see against background.
- Key --> Wallet
- I need to test on mobile & find mempool.space link(?)

nickfarrow:
Think this is good to go - looking fantastic!

I found the mempool.space link, wasn't so intuitive on desktop but that's fine I think.

Please take a quick look @evanlinjin and we'll merge if there's nothing I broke or anything you want to adjust.

nickfarrow (approved):

nickfarrow:
note: test broadcasting transaction -> return to wallet home

LLFourn:
Can we merge this and fix problems in a follow up?

evanlinjin (commented):
ACK 14da2ebd835747ca123532b78b2d4050fcd10c15

PR #191, opened 2024-12-20
Update esp-hal to v0.22.0

musdom:
- Take gpio pins from peripherals
- Remove 'Blocking' and uart Instance type constraints
- No longer passing reference to clocks in drivers
- Change baudrate in apply_config
- SHA peripheral change in ota verifier

Tested OTA update, keygen and sign with 3 devices. It didn't panic when I sign right after keygen.


device/src/bin/v2.rs
```diff
@@ -650,17 +639,23 @@ fn panic(info: &core::panic::PanicInfo) -> ! {
         None => write!(&mut panic_buf, "{}", info),
     };
 
-    let spi = Spi::new(peripherals.SPI2, 80u32.MHz(), SpiMode::Mode2, &clocks)
-        .with_sck(io.pins.gpio8)
-        .with_mosi(io.pins.gpio7);
+    let spi = Spi::new_with_config(
+        peripherals.SPI2,
+        spiConfig {
+            frequency: 40.MHz(),
```
LLFourn, line 645: Why did this change from 80 to 40. Did the semantics of the arguments change or are you changing the frequency for some reason?

LLFourn (approved):
LGTM. I asked one question.

device/src/bin/v2.rs
musdom: This should be 80 my bad

PR #193, opened 2025-01-07
Add app icon

nickfarrow:
If we leave the icon background transparent, it becomes white - doesn't look so nice. I quite like the sky blue but can try other colours!


LLFourn (approved):

LLFourn:
might as well yolo approve and merge this.

PR #198, opened 2025-01-09
[device] Fix panic handler

LLFourn:
Also remove some duplication between main and the panic handler


nickfarrow (approved):
Tested and reviewed

PR #199, opened 2025-01-10
[device] Use nor flash properly

LLFourn:
Based on #198.

Use nor flash properly. Never do erasure before writing. Just append
only in WORD_SIZE aligned chunks. Length prefix entries also.

### DIFF
--- a/device/src/storage.rs
+++ b/device/src/storage.rs
@@ -1,12 +1,13 @@
 extern crate alloc;
-use alloc::{format, string::String, vec::Vec};
+use alloc::{string::String, vec::Vec};
 use bincode::{
     de::read::Reader,
     enc::write::Writer,
     error::{DecodeError, EncodeError},
 };
-use embedded_storage::{nor_flash, ReadStorage, Storage};
-use esp_storage::{FlashStorage, FlashStorageError};
+use embedded_storage::nor_flash::{self, NorFlash};
+use esp_storage::FlashStorageError;
+use frostsnap_comms::BINCODE_CONFIG;
 use frostsnap_core::{device, schnorr_fun::fun::Scalar};
 
 const NVS_PARTITION_START: u32 = 0x3D0000;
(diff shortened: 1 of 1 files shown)


LLFourn:
Probably don't merge this yet -- it's a breaking change for storage so I'll throw in a few other breaking changes there to be merged with it.

device/src/storage.rs
```diff
@@ -79,55 +81,88 @@ impl DeviceStorage {
...
+
+            self.write_buffer[0..WORD_SIZE].copy_from_slice(
+                (word_length - 1/* don't include the length word */)
+                    .to_le_bytes()
+                    .as_slice(),
+            );
+            let byte_pos = (self.word_pos * WORD_SIZE) as u32;
+            nor_flash::NorFlash::write(&mut self.flash, byte_pos, &self.write_buffer[..])?;
+            self.word_pos += word_length as usize;
+            self.write_buffer.clear();
```
nickfarrow, line 123: Great - drew out on pen and paper quickly to validate

device/src/storage.rs
```diff
@@ -79,55 +81,88 @@ impl DeviceStorage {
...
         core::iter::from_fn(move || {
-            let pos_before_read = self.pos;
-            match bincode::decode_from_reader(&mut *self, bincode::config::standard()) {
-                Ok(change) => Some(change),
-                Err(_) => {
-                    self.pos = pos_before_read;
-                    None
-                }
+            let mut length_buf = [0u8; WORD_SIZE];
+            let byte_pos = self.word_pos * WORD_SIZE;
```
nickfarrow, line 132: Here we're only reading forward from `self.word_pos`, instead of from `DATA_START`.
Maybe intended since we only iter upon initial boot, making a note because it might be unexpected behaviour if we were to ever iter through the flash elsewhere in the code.

nickfarrow (approved):

device/src/storage.rs
```diff
@@ -79,55 +81,88 @@ impl DeviceStorage {
...
+            self.write_buffer
+                .extend(core::iter::repeat_n(0xff, padding_needed));
+            assert_eq!(self.write_buffer.len() % WORD_SIZE, 0);
+
+            let word_length: u32 = (self.write_buffer.len() / WORD_SIZE)
+                .try_into()
+                .expect("flash item too large");
+
+            self.write_buffer[0..WORD_SIZE].copy_from_slice(
+                (word_length - 1/* don't include the length word */)
```
LLFourn, line 116: I think this can underflow.

PR #202, opened 2025-01-21
Rework Send Bitcoin flow

evanlinjin:
* Bitcoin address QR code scanning.
* Add `calculate_avaliable_value` method
* Font simplifications (to be fixed later).
* Flutter-side state management for send flow.
* Feerate picker.
* Feerate ETA estimates from Electrum.
* Send all button.
* Change default font to Noto Sans and Noto Sans Mono. According to various studies, these fonts have greater appeal and readability.
* Use global variable for blur filter.


frostsnapp/lib/address_field.dart
evanlinjin, line 1: Okay I need to rename this to `send_models.dart` or something

evanlinjin:
PSBT-signing needs to be added back.

nickfarrow:
Looking really nice, some preliminary feedback/thoughts

- I think the send flow may feel nicer on its own page.
- Send max is a fantastic way to avoid the "send all" debate. Nice!
- I think the user should have to set the amount before feerate to avoid "Cannot calculate fee: no send amount"
- I think the user should have to choose the amount (and finalize everything) before we display a tick, maybe an -> arrow instead.
- Pull down doesn't work on desktop, is there an easy way to change the behavior for desktop? or should we avoid? Not sure what to do here

evanlinjin:
> I think the user should have to set the amount before feerate to avoid "Cannot calculate fee: no send amount".

Re-thinking about this. I think the first screen of the send flow should ask the user for tx priority. This will display the "available amount" and "max/min possible fee" on this screen.

Should we continue the discussion on this and tackle it in a separate PR?

evanlinjin:
Sometimes the feerates do not load...

nickfarrow:
Fantastic work @evanlinjin ! After taking another look I think the tx-priority up-front works, the row format before signing is very nice.

nickfarrow (approved):

PR #203, opened 2025-01-24
Nonce streams

LLFourn:
```
To allow multiple nonce lists between the same and different
coordinators.

1. Coordinators open nonce streams with a random 16 byte id. The device
then responds with nonces for the id.
2. The device and coordinator try and reconcile deviations in their
nonce set when the device is plugged in.
3. The device permanently deletes a nonce once it's use to create a
signature. This is guaranteed by the design as long as the
NonceStreamSlot trait is implemented to correctly erase the slot when
it's written.
```

TOOD: exfil protoection

### DIFF
--- a/frostsnap_core/src/device_nonces.rs
+++ b/frostsnap_core/src/device_nonces.rs
@@ -1,0 +1,379 @@
+use alloc::collections::VecDeque;
+use alloc::vec::Vec;
+use chacha20::{
+    cipher::{KeyIvInit, StreamCipher},
+    ChaCha20,
+};
+use rand_core::RngCore;
+use schnorr_fun::{
+    binonce,
+    frost::{NonceKeyPair, PairedSecretShare, PartySignSession, SignatureShare},
+    fun::prelude::*,
+};
+
+use crate::{
+    nonce_stream::{CoordNonceStreamState, NonceStreamId, NonceStreamSegment},
+    SignSessionId, NONCE_BATCH_SIZE,
+};
+
+type ChaChaSeed = [u8; 32];
+
+#[derive(Clone, Debug, PartialEq, bincode::Encode, bincode::Decode)]
+pub struct SecretNonceSlot {
+    //XXX: index must be first so we can decode it easily without decoding the whole thing
+    //XXX: index == u32::MAX is invalid because we use it to represent "uninitialized"
+    pub index: u32,
+    pub nonce_stream_id: NonceStreamId,
+    pub ratchet_prg_seed: ChaChaSeed,
+    pub signing_state: Option<SigningState>,
+}
+
+#[derive(Clone, Debug, PartialEq, bincode::Encode, bincode::Decode)]
+pub struct SigningState {
+    pub session_id: SignSessionId,
+    pub signature_shares: Vec<SignatureShare>,
+}
+
+pub trait NonceStreamSlot {
+    fn read_index(&mut self) -> Option<u32>;
+    fn read_slot(&mut self) -> Option<SecretNonceSlot>;
+    fn write_slot(&mut self, value: &SecretNonceSlot);
+}
+
+/// Manages two writable sectors of persistent storage such that we make sure the state of the system we're managing is never lost.
+/// The new state is first written, if that succeeds we finally write over the previous state.
+/// In this case we're managing the secret seed needed to produce nonces for signatures.
+#[derive(Clone, Debug, PartialEq)]
+pub struct ABSlot<S> {
+    slots: [S; 2],
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct ABSlots<S> {
+    slots: Vec<ABSlot<S>>,
+}
+
+impl<S: NonceStreamSlot> ABSlots<S> {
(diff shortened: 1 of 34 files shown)


LLFourn:
Exfil done here: #205

frostsnap_core/src/device_nonces.rs
```diff
@@ -0,0 +1,379 @@
...
+            i += 1;
+            if i >= self.slots.len() {
+                break None;
+            }
+        };
+
+        match found {
+            Some(i) => &mut self.slots[i],
+            None => {
+                let random_overwrite_idx = rng.next_u32() as usize % self.slots.len();
```
nickfarrow, line 91: What's the motivation for replacing a random slot?
Prevent attempts of gaming of some kind? Or to spread nonce-slot NOR wear across the sectors? Something else?

An alternative could be to replace the least recently used - but probably much more complex with some way to track.

Imagining if I have some key with devices on coordinator A, and I make a new key with a subset of those devices on a new coordinator B; if those devices have full slots, some of them may opaquely overwrite the nonce-slots belonging to coordinator A.

Then, next time I go to sign with coordinator A, I visit devices one by one, but maybe the last device no longer has knowledge of that nonce stream, and I have to start the signing session again?

frostsnap_core/src/device_nonces.rs
```diff
@@ -0,0 +1,379 @@
...
+    SignSessionId, NONCE_BATCH_SIZE,
+};
+
+type ChaChaSeed = [u8; 32];
+
+#[derive(Clone, Debug, PartialEq, bincode::Encode, bincode::Decode)]
+pub struct SecretNonceSlot {
+    //XXX: index must be first so we can decode it easily without decoding the whole thing
+    //XXX: index == u32::MAX is invalid because we use it to represent "uninitialized"
+    pub index: u32,
```
nickfarrow, line 25: Cool that you can do these partial reads with bincode!

frostsnap_core/src/device_nonces.rs
```diff
@@ -0,0 +1,379 @@
...
+        let found = loop {
+            let ab_slot = &mut self.slots[i];
+            match ab_slot.nonce_stream_id() {
+                Some(slot_nsid) => {
+                    if slot_nsid == stream_id {
+                        break Some(i);
+                    }
+                }
+                None => {
+                    ab_slot.initialize(stream_id, rng);
```
nickfarrow, line 78: Might be worth noting that this design allocates slots sequentially, we never delete them and overwrite them.
This means that searching & initializing like this is OK since we'll never have an empty slot before a non-empty one(s).

frostsnap_core/src/device_nonces.rs
```diff
@@ -0,0 +1,379 @@
...
+                let ab_slot = &mut self.slots[random_overwrite_idx];
+                ab_slot.initialize(stream_id, rng);
+                ab_slot
+            }
+        }
+    }
+
+    pub fn get(&mut self, stream_id: NonceStreamId) -> Option<&mut ABSlot<S>> {
+        //XXX: clippy is wrong about this
+        #[allow(clippy::manual_find)]
```
nickfarrow, line 101: I didn't believe, but I tried .find() and now I believe..! Reference mess

frostsnap_core/src/device_nonces.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+}
+
+impl<S: NonceStreamSlot> ABSlot<S> {
+    pub fn new(a: S, b: S) -> Self {
+        Self { slots: [a, b] }
+    }
+
+    pub fn initialize(&mut self, stream_id: NonceStreamId, rng: &mut impl RngCore) {
+        let mut ratchet_prg_seed = ChaChaSeed::default();
```
nickfarrow, line 123: An occurrence of automatic trait impl derivation that is perhaps risky, think this works

```
#[derive(Clone, Debug, PartialEq, bincode::Encode, bincode::Decode)]
#[repr(transparent)]
pub struct ChaChaSeed([u8; 32]);

impl ChaChaSeed {
    pub fn new(rng: &mut impl RngCore) -> Self {
        let mut bytes = [0; 32];
        rng.fill_bytes(&mut bytes);
        Self(bytes)
    }  
}
```

frostsnap_core/src/device_nonces.rs
LLFourn: Good point. I'm going to address this in another PR.

frostsnap_core/src/device_nonces.rs
LLFourn: Will add a note about this in the subsequent PR since this has #208 building on it

frostsnap_core/src/device_nonces.rs
```diff
@@ -0,0 +1,379 @@
...
+    fn current_slot(&mut self) -> usize {
+        if self.slots[1].read_index() > self.slots[0].read_index() {
+            1
+        } else {
+            0
+        }
+    }
+
+    pub fn current_index(&mut self) -> u32 {
+        self.slots[0]
```
nickfarrow, line 295: Should this use `self.current_slot()`?

nickfarrow (approved):

PR #204, opened 2025-01-27
Device symmetric key from hardware HMAC

musdom:
Use hardware HMAC to generate keys from a secret key in efuse. The HMAC peripheral has access to efuse keys that are otherwise inaccessible to firmware, if efuse_protect cargo feature is enabled.

If Key 0 is empty, devices will generate a random key and burn it in efuse Block 4. Then it will burn HMAC Upstream key purpose and write protect bits in Block 0. init_key can also burn keys for other uses.


device/src/efuse.rs
```diff
@@ -0,0 +1,198 @@
...
+        // We bundle every config in Block 0 to minimize write operations
+        // Todo: Write key purpose and rw flags for multiple keys
+        // Disable write to key block
+        let mut write_disable = 0x01_u32 << (23 + key_number);
+        // Disable write to key purpose
+        write_disable += 0x01_u32 << (8 + key_number);
+        buff[0..4].copy_from_slice(&write_disable.to_le_bytes());
+
+        #[cfg(feature = "efuse_protect")]
+        self.set_read_protect(key_number, &mut buff);
```
LLFourn, line 68: :wrench:  Have a bool to this his function which decides whether it should be read protected. The bool can be set in the main function by the feature flag. like:

```
let do_read_protect = #[cfg(feature = "efuse_protect")];
```

The name of the feature should be `read_protect_hmac_key`.

device/src/efuse.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        self.write_block(&buff, 0)
+    }
+
+    /// # Safety
+    /// For disabling reading of efuse keys after efuse_debug
+    pub unsafe fn set_read_protect(&self, key_number: u8, buff: &mut [u8; 32]) {
```
LLFourn, line 75: `pub` seems to be a mistake here.

device/src/efuse.rs
```diff
@@ -0,0 +1,198 @@
...
+pub struct EfuseController {
+    pub efuse: EFUSE,
+}
+
+impl EfuseController {
+    pub fn new(efuse: EFUSE) -> Self {
+        Self { efuse }
+    }
+
+    pub fn init_key(&self, key_number: u8, rng: &mut impl RngCore) -> Result<(), EfuseError> {
```
LLFourn, line 16: something like `init_hmac_key`?

device/src/efuse.rs
```diff
@@ -0,0 +1,198 @@
...
+            _ => return Err(EfuseError::EfuseError),
+        };
+        let key: u8 = Efuse::read_field_le(efuse_field);
+        // Check if there's an existing key so we don't accidentally overwrite it
+        if key == 0 {
+            let mut buff = [0x00_u8; 32];
+            rng.fill_bytes(&mut buff);
+
+            unsafe {
+                self.write_block(&buff, key_number + 4)?;
```
LLFourn, line 33: :wrench:  what's with the +4 can you put that in a descriptive constant somewhere.

device/src/efuse.rs
musdom: I should make it take a KeyPurpose too so you can use it for other kinds of keys

device/src/efuse.rs
musdom: Key 0 is in efuse block 4, and so on

device/src/efuse.rs
```diff
@@ -0,0 +1,198 @@
...
+        } else {
+            Ok(())
+        }
+    }
+
+    unsafe fn send_write_command(&self, block_number: u8) {
+        let efuse = &self.efuse;
+
+        // Send opcode, blknum and write command
+        efuse.conf().write(|w| w.op_code().bits(0x5A5A));
```
LLFourn, line 127: Weird to me that the PAC doesn't put this into a constant somewhere. Can we put that this is in a read command constant.

device/src/bin/v2.rs
```diff
@@ -194,11 +197,25 @@ fn main() -> ! {
...
+            efuse::KeyPurpose::HmacUpstream,
+            do_read_protect,
+            &mut hal_rng,
+        )
+        .expect("Efuse init error");
+
     let rng = {
         let mut chacha_seed = [0u8; 32];
         hal_rng.read(&mut chacha_seed);
         ChaCha20Rng::from_seed(chacha_seed)
```
LLFourn, line 215: bonus: Use the HMAC peripheral to HMAC what you read from the  trng to get the ChaCha20Rng seed. This makes it harder to physically reduce entropy by trying to physically interfere with the TRNG (because the initial entropy from the burnt key is mixed in).

musdom: I init Key 1 and use it to HMAC the seed

musdom:
Key 0: share encryption key
Key 1: ChaCha20rng seed HMAC key

LLFourn (approved):
Self-ACK. I cleaned up the API in 941a135f01021554c426fa97a1608c2715bc02dc

I also decided now is the time to try and extract more bytes from the TRNG to seed the initial rng.

PR #205, opened 2025-01-28
Anti exfil

LLFourn:
on top of #203


nickfarrow (approved):
Reviewed, tested, and looked through the relevant secpfun commit also https://github.com/LLFourn/secp256kfun/commit/821ffafdab9ba35a6903ca3d9a3796f62f7f3451

Issue #160, opened 2024-10-16
After KeyGen backup tasks that encourage geographic separation


nickfarrow:
Some ideas for encouraging users to make good backups.

- We'd like users to begin the backup process immediately after keygen.
- We also want users to geographically separate each backup.

We can help users make good decisions though a series of quests/tasks that form an enjoyable component to the Frostsnap setup experience.

If a user is completing their share backup tasks in succession, we can enact a ~15 minute time delay before showing the next backup in order to strongly encourage geographic separation. Do not display this timer countdown, but leave it as a mystery to the user so they actually feel encouraged to travel to their next secret location on their quest.

Something like:
---

 #### Next task: *Backup `device_name` - Travel to your second share citadel/guardian to create and conceal its backup.*
 - Tips: you'll want to find a place or person secure to protect this secret share.
 - Bring: `device_name`, and a paper & pen.
 - *Once you have arrived you can create and protect your this secret share*. 

** User starts **

 - Possible notification: *"Your secret share for device xxxx is now available for backup once you have arrived!"*
 - `[I'm Here]` button --> if too soon: *"You didn't travel very far! Are you sure this is a separate and secure location for your backup?"*
 - ** Backup Workflow **
 - *"Also hide device `device_name` under this citadel/guardian's protection."*
 - *"Optional: Add location hint(!): _________"*

#### Next task (hidden until prev. is complete): *Backup `device_name` - Travel to your third share citadel/guardian*....

and so on.

---

Could tasks+hidden timers add some chore/frustration..? maybe.. But I think we can make the setup experience enjoyable while encouraging extremely good practices for noobs securing their funds. We were considering the idea of pinging users with reminders to backup anyway so a "Current task: xxx" could be nice until they complete it, maybe lean into a magical experience if we can.

A pattern of  "quests" could also be useful elsewhere for encouraging good behaviors, e.g. *"New Task Available - Verify the Integrity of Backup #3"*. We would need to think carefully about what language we use for the feeling we want to convey. "Quest" is probably too unserious for some users. We can even reward users somehow for completing tasks, fits nicely into the idea of leveling up a frostsnap-pet 🐧.

To encourage users to maintain this geographic separation into the future, we could display a warning if they sign on a threshold number of devices within too short of a time-frame (e.g. if `signing_session_duration < threshold / 10mins`).

LLFourn:
Closing because we've done our best here for now

PR #206, opened 2025-01-29
Backup checklist

nickfarrow:
Introduces a backup workflow (https://github.com/frostsnap/frostsnap/issues/160) where users are instructed to backup each device and leave in different geographic locations. This process automatically begins at the end of key generation.

Users can pause and resume the process at any time, they can also go through the (greyed out) backup checklist items again if they so desire.

If a backup run has not been completed, a warning banner will display in the wallet.

A warning is displayed if a user tries to backup successive different devices too quickly (<5min), encouraging geographic separation.

This does not yet add much excitement to these tasks, still thinking on that one - what emotion and vibe we want the app to exude.

### DIFF
--- a/frostsnap_coordinator/src/backup_run.rs
+++ b/frostsnap_coordinator/src/backup_run.rs
@@ -1,0 +1,146 @@
+use crate::persist::Persist;
+use frostsnap_core::{DeviceId, KeyId};
+use rusqlite::params;
+use std::collections::BTreeMap;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+#[derive(Debug, Clone, PartialEq, bincode::Encode, bincode::Decode, Default)]
+pub struct BackupRun {
+    pub devices: Vec<(DeviceId, Option<u32>)>,
+}
+
+impl BackupRun {
+    pub fn new(devices: Vec<DeviceId>) -> Self {
+        Self {
+            devices: devices.into_iter().map(|d| (d, None)).collect(),
+        }
+    }
+
+    pub fn mark_device_complete(&mut self, device_id: DeviceId) {
+        if let Some(entry) = self.devices.iter_mut().find(|(d, _)| *d == device_id) {
+            entry.1 = Some(
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap()
+                    .as_secs() as u32,
+            );
+        }
+    }
+
+    pub fn is_run_complete(&self) -> bool {
+        self.devices
+            .iter()
+            .all(|(_, timestamp)| timestamp.is_some())
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct BackupState {
+    pub runs: BTreeMap<KeyId, BackupRun>,
+}
+
+#[derive(Debug, Clone)]
+pub enum BackupMutation {
+    StartBackup {
+        key_id: KeyId,
+        devices: Vec<DeviceId>,
+    },
+    MarkDeviceComplete {
+        key_id: KeyId,
+        device_id: DeviceId,
+        timestamp: u32,
+    },
+}
+
+impl Persist<rusqlite::Connection> for BackupState {
+    type Update = Vec<BackupMutation>;
+    type InitParams = ();
+
+    fn initialize(conn: &mut rusqlite::Connection, _: ()) -> anyhow::Result<Self> {
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS backup_runs (
+                key_id TEXT NOT NULL,
+                device_id TEXT NOT NULL,
+                timestamp INTEGER,
+                PRIMARY KEY (key_id, device_id)
+            )",
+            [],
+        )?;
+
+        let mut stmt = conn.prepare(
+            "SELECT key_id, device_id, timestamp 
+             FROM backup_runs 
+             ORDER BY key_id",
+        )?;
+
+        let rows = stmt.query_map([], |row| {
+            Ok((
+                row.get::<_, String>(0)?,      // key_id
+                row.get::<_, String>(1)?,      // device_id
+                row.get::<_, Option<u32>>(2)?, // timestamp
+            ))
+        })?;
+
+        // group by key_id and build BackupRun for each group
+        let runs = rows
+            .into_iter()
+            .collect::<Result<Vec<_>, _>>()?
+            .into_iter()
+            .fold(
+                BTreeMap::new(),
+                |mut map: BTreeMap<_, Vec<_>>, (key_id, device_id, timestamp)| {
+                    map.entry(key_id).or_default().push((device_id, timestamp));
+                    map
+                },
+            )
+            .into_iter()
+            .map(|(key_id, devices)| {
+                let devices = devices
+                    .into_iter()
+                    .map(|(device_id, timestamp)| Ok((device_id.parse()?, timestamp)))
+                    .collect::<Result<Vec<_>, anyhow::Error>>()?;
+
+                Ok((key_id.parse()?, BackupRun { devices }))
+            })
+            .collect::<Result<BTreeMap<_, _>, anyhow::Error>>()?;
+
+        Ok(BackupState { runs })
+    }
+
+    fn persist_update(
+        conn: &mut rusqlite::Connection,
+        update: Vec<BackupMutation>,
+    ) -> anyhow::Result<()> {
+        for mutation in update {
+            match mutation {
+                BackupMutation::StartBackup { key_id, devices } => {
+                    // Delete any previous backup run
+                    conn.execute(
+                        "DELETE FROM backup_runs WHERE key_id = ?1",
+                        params![key_id.to_string()],
+                    )?;
+
+                    for device_id in devices {
+                        conn.execute(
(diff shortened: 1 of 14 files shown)


frostsnapp/native/src/api.rs
```diff
@@ -695,20 +697,20 @@ impl SuperWallet {
     }
 }
 
-pub fn load(app_dir: String) -> anyhow::Result<(Coordinator, Settings)> {
+pub fn load(app_dir: String) -> anyhow::Result<(Coordinator, Settings, BackupManager)> {
```
LLFourn, line 700: I think we shouldn't have an ever growing tuple here for every bit of app state we need to keep on the rust side. Make a struct that has `Settings` and `BackupManger` inside it. It can also go in the same "context" on the dart side. Probably we should just merge all these top level "contexts" into one `FrostsnapContext`. Coordinator should also be in it rather than be global variable.

frostsnapp/lib/wallet.dart
```diff
@@ -727,3 +732,129 @@ Uri getBlockExplorer(BitcoinNetwork network) {
...
+                      context,
+                      MaterialPageRoute(
+                        builder: (context) => BackupStartPage(
+                          accessStructure:
+                              widget.frostKey.accessStructures()[0],
+                        ),
+                      ),
+                    );
+                    // force a rebuild after navigation
+                    setState(() {
```
LLFourn, line 822: why though?

frostsnapp/lib/wallet.dart
```diff
@@ -727,3 +732,129 @@ Uri getBlockExplorer(BitcoinNetwork network) {
...
+  const BackupWarningBanner({
+    required this.frostKey,
+    required this.backupManager,
+    Key? key,
+  }) : super(key: key);
+  @override
+  State<BackupWarningBanner> createState() => _BackupWarningBannerState();
+}
+
+class _BackupWarningBannerState extends State<BackupWarningBanner>
```
LLFourn, line 748: I don't understand why this has to be stateful. You have a Stream and you want to conditionally build the widget based on the value of the stream. That's what StreamBuilder is for.

frostsnapp/lib/backup_workflow.dart
```diff
@@ -0,0 +1,504 @@
...
+    );
+
+    return StreamBuilder<BackupRun>(
+      stream: backupStream,
+      builder: (context, snapshot) {
+        final allDevices = widget.accessStructure.devices();
+        final keyId = widget.accessStructure.accessStructureRef().keyId;
+        final manager = BackupContext.of(context)!.manager;
+        final pendingDevices = allDevices
+            .where((deviceId) => !manager.isDeviceComplete(
```
LLFourn, line 217: What's the point of this `StreamBuilder` when you aren't using the value of snapshot. That's the whole point!

The stream should be a stream of a data struct. Not an opaque type like it is now.

frostsnap_coordinator/src/backup_run.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn persist_update(
+        conn: &mut rusqlite::Connection,
+        update: Vec<BackupMutation>,
+    ) -> anyhow::Result<()> {
+        for mutation in update {
+            match mutation {
+                BackupMutation::StartBackup { key_id, devices } => {
+                    // Delete all except completion record
+                    conn.execute(
+                        "DELETE FROM backup_runs WHERE key_id = ?1 AND device_id != '_completed'",
```
LLFourn, line 125: It seems like you are trying to keep track of which backup has been completed. But why can't you just tell this by how many devices have been completed?

It is very weird to insert something into the `device_id` column that is not a `device_id`.

nickfarrow: The idea was to store some knowledge that a backup run has been completed before.

If a user starts a subsequent backup run, I was thinking that we should let them cancel it (maybe mid way though) and get rid of the warning banner. I haven't implemented this yet though.

Maybe a bit too hacky? Was avoiding another separate table to store a history of completed backup runs.

LLFourn: Do we need a history? As far as I understand there is a single backup task with each wallet. It can be in a state of `None` or `Some(BackupProgress)`. I don't think it needs to be any more complicated than that. We don't need to keep around whether they have completed one before this one. Just let them cancel it and set it back to `None`.

nickfarrow:
Thanks for great feedback! Much cleaner now.

Simplified by dropping the ability to restart the backup run after a backup run has been completed (also getting rid of the `_completed` item in device_id column).

Walking legs material icon isn't fantastic, but also not terrible. Couldn't find any decent license free animation and AI wasnt helping much.

One remaining bug is that the stream widget updates seems to fail in the case of a recovered key.

evanlinjin:
flutter_01

@nickfarrow is there a reason why we have a settings button top right corner in this screen?

Also, what do you think about making this screen a bottom sheet/dialog? Personally, I think this would make it more consistent and you can see which wallet you are backing up.

evanlinjin:
As mentioned in the meeting today, there is a bug when we close the "Connect device" modal bottom sheet by dragging, the Security Checklist page is also closed.

frostsnapp/lib/wallet.dart
```diff
@@ -716,3 +846,107 @@ Uri getBlockExplorer(BitcoinNetwork network) {
...
+
+  static BannerStateProvider? of(BuildContext context) {
+    return context.dependOnInheritedWidgetOfExactType<BannerStateProvider>();
+  }
+
+  @override
+  bool updateShouldNotify(BannerStateProvider oldWidget) {
+    return showBanner != oldWidget.showBanner;
+  }
+}
```
evanlinjin, line 867: This `InheritedWidget` is just here so that we can get the initial state of whether we should or should not show the banner?

However, we set the `showBanner` value when we construct `WalletHomeWithConfetti` and then build the widgets by "wrapping" it in this `BannerStateProvider`.

I think it makes more sense just to not construct the `BackupWarningBanner` if we don't have `showBanner`, and pass the `showBanner` into `WalletHome` when we construct it.

frostsnapp/lib/wallet.dart
```diff
@@ -716,3 +846,107 @@ Uri getBlockExplorer(BitcoinNetwork network) {
...
+  @override
+  Widget build(BuildContext context) {
+    final backupManager = FrostsnapContext.of(context)!.backupManager;
+    final walletCtx = WalletContext.of(context)!;
+    final backupStream = walletCtx.backupStream;
+
+    final bannerState = BannerStateProvider.of(context);
+    // Don't show if we're in new key flow and haven't dismissed dialog
+    if (bannerState != null && !bannerState.showBanner) {
+      return const SliverToBoxAdapter(child: SizedBox.shrink());
```
evanlinjin, line 888: Just don't construct `BackupWarningBanner` if showBanner is false.

frostsnapp/lib/wallet.dart
```diff
@@ -716,3 +846,107 @@ Uri getBlockExplorer(BitcoinNetwork network) {
...
+            !backupManager.isRunComplete(keyId: frostKey.keyId());
+
+        if (!showWarning) {
+          return const SliverToBoxAdapter(child: SizedBox.shrink());
+        }
+
+        return SliverToBoxAdapter(
+          child: Material(
+            color: theme.colorScheme.errorContainer,
+            child: InkWell(
```
evanlinjin, line 907: How would it look if we just use a `ListTile` here?

frostsnapp/lib/wallet.dart
evanlinjin: Okay I see why you wanted to use an `InheritedWidget`. It's because otherwise, we need to pass the `showBanner` variable through a few layers of widgets.

frostsnapp/lib/wallet.dart
nickfarrow: If I remember correctly this was to prevent:
1) Clicking begin backup,
2) Immediately backing out
3) User can't see banner, when we probably do want to show it (so they don't go and receive a million bitcoins right away)

Though your design sounds better, maybe see if you can avoid that behavior?

frostsnapp/lib/wallet.dart
evanlinjin: I think we should merge `TxList` and `WalletHome` as a follow up and just pass `showBanner` via the constructors for now.

frostsnap_coordinator/src/check_share.rs
```diff
@@ -67,7 +67,7 @@ impl CheckShareProtocol {
 
 impl UiProtocol for CheckShareProtocol {
     fn cancel(&mut self) {
-        self.abort("loading share cancelled".to_string())
+        self.abort("Loading share cancelled".to_string())
```
nickfarrow, line 70: I capitalized this because we're currently showing this error on the snackbar, probably will move to better error messaging here, but a quick style-fix for now.

frostsnapp/lib/wallet.dart
nickfarrow: Yeah very hacky sorry! I should've left a comment. Agree, passing via constructors does sound much better - worth trying in this PR?

LLFourn:
closed in favor of #221

PR #208, opened 2025-02-04
[device] Formalize flash partitions

LLFourn:
Create embedded_storage which contains systems that work over any
`embedded_storage::NorFlash` backend i.e. doesn't depend on esp32_*.

In particular I've `FlashPartition` which creates a sort of virtual
flash drive that you can pass to any sub system so it can exclusively
write to it (and cannot write outside of it).

In addition, I've moved the append only style database to
`embedded_storage` along with some of the logic around "A-B" slots.

Upgrading will wipe the device.


frostsnap_embedded/src/partition.rs
```diff
@@ -0,0 +1,324 @@
...
+
+            self.buf[self.buf_index] = bytes[i];
+            self.buf_index += 1;
+
+            if self.buf_index == BUFFER_SIZE {
+                self.flash
+                    .nor_write(self.word_pos * S::WRITE_SIZE as u32, &self.buf[..])
+                    .map_err(|e| bincode::error::EncodeError::OtherString(format!("{:?}", e)))?;
+                self.buf_index = 0;
+                self.word_pos += (BUFFER_SIZE / S::WRITE_SIZE) as u32;
```
nickfarrow, line 266: BUFFER_SIZE is always a multiple of WRITE_SIZE?

frostsnap_embedded/src/nor_flash_log.rs
```diff
@@ -0,0 +1,135 @@
...
+                return Some(Err(bincode::error::DecodeError::OtherString(format!(
+                    "failed to read length byte at {length_word_byte_pos} ({:?}) from {:?}",
+                    e, self.flash,
+                ))));
+            }
+            if length_buf == [0xff; WORD_SIZE as usize] {
+                return None;
+            }
+            self.word_pos += 1;
+            let word_length = u32::from_le_bytes(length_buf);
```
nickfarrow, line 63: In the extreme event of a word_length word being corrupted, I think this function could accidentally read beyond the bounds. We could add a check `expected_end > self.flash.size()`. BUT we already handle out-of-bounds errors in `FlashPartition`?

frostsnap_embedded/src/partition.rs
LLFourn: yes. 

image

frostsnap_embedded/src/sha.rs
```diff
@@ -0,0 +1 @@
+
```
LLFourn, line 1: accidental file commit

device/src/flash/share.rs
```diff
@@ -0,0 +1 @@
+
```
nickfarrow, line 1: Empty file now

device/src/flash_header.rs
```diff
@@ -0,0 +1 @@
+
```
nickfarrow, line 1: Empty file now

device/src/partitions.rs
```diff
@@ -0,0 +1,106 @@
...
+                "nvs" => {
+                    self_.nvs.set_offset_and_size(row.offset, row.size as u32);
+                }
+                _ => { /*ignore*/ }
+            }
+        }
+
+        for part in [
+            self_.nvs,
+            self_.ota.ota_1,
```
nickfarrow, line 74: Repeated, think this one should be `self_.ota.otadata`

device/src/esp32_run.rs
```diff
@@ -55,45 +54,58 @@ where
...
-        let active_firmware_digest = active_partition.digest(&mut sha256);
         ui.set_busy_task(ui::BusyTask::Loading);
+        let flash = RefCell::new(FlashStorage::new());
+        let mut partitions = crate::partitions::Partitions::load(&flash);
+        let header_flash = FlashHeader::new(partitions.nvs.split_off_front(2));
+        let header = match header_flash.read_header() {
+            Some(header) => header,
+            None => {
+                if !partitions.nvs.is_empty().expect("checking NVS is empty") {
+                    panic!("the device appears to be new but the NVS is not blank. Maybe you're a developer who needs to manually erase the device?");
```
nickfarrow, line 65: 🙏

nickfarrow:
Fantastic, just needs minor fixups with missing files and checking `otadata` is not empty during load. Testing now

device/src/partitions.rs
LLFourn: yes

frostsnap_embedded/src/nor_flash_log.rs
LLFourn: The flash partition does indeed protect against reading out of bounds.

nickfarrow (approved):
Empty files removed in forcepush

PR #209, opened 2025-02-05
Erase nonce streams based on which one is least recently used

LLFourn:
This is on top of #208

### DIFF
--- a/frostsnap_core/src/device_nonces.rs
+++ b/frostsnap_core/src/device_nonces.rs
@@ -13,18 +13,18 @@ use schnorr_fun::{
 
 use crate::{
     nonce_stream::{CoordNonceStreamState, NonceStreamId, NonceStreamSegment},
-    SignSessionId, NONCE_BATCH_SIZE,
+    SignSessionId, Versioned, NONCE_BATCH_SIZE,
 };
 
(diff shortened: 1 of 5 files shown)


frostsnap_core/src/device_nonces.rs
```diff
@@ -13,18 +13,18 @@ use schnorr_fun::{
...
-    SignSessionId, NONCE_BATCH_SIZE,
+    SignSessionId, Versioned, NONCE_BATCH_SIZE,
 };
 
 type ChaChaSeed = [u8; 32];
 
 #[derive(Clone, Debug, PartialEq, bincode::Encode, bincode::Decode)]
 pub struct SecretNonceSlot {
-    //XXX: index must be first so we can decode it easily without decoding the whole thing
-    //XXX: index == u32::MAX is invalid because we use it to represent "uninitialized"
```
nickfarrow, line 24: Is this comment no longer appropriate? I found it useful!

LLFourn: Yes this comment is no longer true. The index is no longer independently decoded or used to the which state is newer as part of the A/B write system.

nickfarrow (approved):

PR #211, opened 2025-02-05
Redesign send confirm workflow

evanlinjin:
Potential discussion points:

1. The UI for picking signers and actually signing are too similar. How can we make it more apparent to the user that they need to plug their devices in.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -187,4 +187,5 @@ jobs:
           cache: true
 
       - uses: taiki-e/install-action@just
+      - run: dart format --version
       - run: just lint-app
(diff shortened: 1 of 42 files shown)


LLFourn:
It could start with telling me which devices to plug in :). This is a 2-of-3 and I selected only 2 devices so why does tell me to plug in 3?

image

evanlinjin:
It's done bois.
* Fixed shown devices on "Sign Transaction" page.
* Added cancel button on "Sign Transaction" page.

LLFourn (approved):
LGTM I added a few commits

evanlinjin (commented):
ACK 650e586cdfc58781e3bf75923ebeff581f12ba2d

Thank you @LLFourn for the fixes

PR #216, opened 2025-02-14
Upload firmware binary with actions

nickfarrow:
Turns out you can't easily `espflash flash` a binary, see https://github.com/viamrobotics/micro-rdk/pull/289
Need to use `espflash write-bin 0x10000 ~/tmp/firmware.bin` with specific flash addresses

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,6 +21,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: taiki-e/install-action@just
       - uses: ./.github/actions/fetch-cargo-deps
+      - uses: "./.github/actions/install-cargo-bins"
 
       # RISC-V toolchain cache
       - uses: actions/cache@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -50,10 +51,15 @@ jobs:
 
       - run: just lint-device --release --locked
       - run: just build-device ${{ matrix.board }} --locked
+      - run: just save-image ${{ matrix.board }}
       - uses: "./.github/actions/upload-artifact"
         with:
           name: frostsnap-${{ github.sha }}.elf
           path: target/riscv32imc-unknown-none-elf/release/v2
```
LLFourn, line 58: :question: can we remove this artifact. I guess the fewer artifacts the better.

LLFourn (approved):

.github/workflows/test.yml
nickfarrow: Removed!

PR #218, opened 2025-02-20
[device] Increase baudrate to 115_200

LLFourn:
So that things happen faster during keygen. We should be able to handle this now that we use conch communication.


musdom (approved):
tACK

PR #220, opened 2025-02-26
[app] Use our electrum servers as default

LLFourn:
They are not using ssl until we can control ssl certs more closely.


LLFourn:
I also added a commit in this one to remove the useless patch statement.

musdom (approved):

PR #221, opened 2025-02-26
Backup workflow (2)

evanlinjin:
Replaces #206

* Remove `BannerStateProvider`. Use a `ValueNotifier` in `WalletContext` instead.
* Use `ListTile` where able (instead of something custom).
* Move non-stream-dependent widgets out of the `StreamBuilder::builder`.
* Have the backup workflow within a modal sheet (instead of a page) for consistency.

### DIFF
--- a/frostsnap_coordinator/src/backup_run.rs
+++ b/frostsnap_coordinator/src/backup_run.rs
@@ -1,0 +1,146 @@
+use crate::persist::Persist;
+use frostsnap_core::{DeviceId, KeyId};
+use rusqlite::params;
+use std::collections::BTreeMap;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+#[derive(Debug, Clone, PartialEq, bincode::Encode, bincode::Decode, Default)]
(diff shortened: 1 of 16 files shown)


nickfarrow (approved):
Awesome just tried it out, initially I was unsure of modal sheet (viewed desktop) but I think it is nicer on mobile.

One extra idea is that we could blur the background when the "Backup & Device Secured" dialog is open - that could reduce the overwhelming text?

One very minor regression is that the warning banner displays before the dialog pops up - but I don't think we need to perfect this minor detail.

evanlinjin:
@nickfarrow it would be simpler to do a blur and no need to manage state. 

In terms of bottom modal sheet, the thing is a dialog on desktop. Let me know what you think.

evanlinjin:
I just updated the "Wallet Created!" Dialog to have a backdrop blur and removed the `ValueNotifier` in the `WalletContext`.

LLFourn:
I merged this as is. I'll make a PR to change the text a bit later.

PR #222, opened 2025-02-27
[device] Busy tasks shouldn't interrupt main tasks

LLFourn:
distinguish between main workflows and busy tasks. Busy tasks shouldn't visuall interrupt main tasks.


nickfarrow (approved):
Tested

PR #225, opened 2025-03-03
[core] Make core API stateless

LLFourn:
- This makes the API of the core structs in `device.rs` and `coordinator.rs` stateless i.e. they are not in "signing mode" "keygen mode". They can always just respond to any message.

- To achieve this backup share images are verified by the device rather
than the coordinator.

- DisplayBackupConfirmed has been moved from core to comms


frostsnap_core/src/sign_task.rs
```diff
@@ -15,19 +15,52 @@ pub enum SignTask {
...
+                SignTask::Test { message }
+            }
+            WireSignTask::Nostr { event } => {
+                if !matches!(purpose, KeyPurpose::Nostr) {
+                    return Err(SignTaskError::WrongPurpose);
+                }
+                SignTask::Nostr { event }
+            }
+            WireSignTask::BitcoinTransaction(tx_template) => {
+                let network = match purpose {
```
nickfarrow, line 59: Not related to this PR, but will we eventually want to fix networks on key purposes?
I think there have been some HWW weaknesses in the past regarding a malicious coordinator / social engineering attack by tricking a user into signing "a testnet transaction" that is actually on mainnet.

frostsnap_coordinator/src/firmware_upgrade.rs
```diff
@@ -64,10 +69,20 @@ impl UiProtocol for FirmwareUpgradeProtocol {
...
+    fn process_comms_message(&mut self, from: DeviceId, message: CommsMisc) {
+        match message {
+            CommsMisc::AckUpgradeMode => {
+                if !self.state.confirmations.contains(&from) {
+                    self.state.confirmations.push(from);
+                    self.emit_state()
+                }
+            }
+            _msg => {
+                error!(
```
nickfarrow, line 81: Maybe overthinking but could upgrading be more robust to changes if we just ignore other messages here?

frostsnap_core/src/device.rs
```diff
@@ -75,19 +76,120 @@ pub struct AccessStructureData {
...
+    pub fn t_of_n(&self) -> (u16, u16) {
+        (
+            self.agg_input.shared_key().threshold().try_into().unwrap(),
+            self.agg_input.encryption_keys().count().try_into().unwrap(),
+        )
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct BackupDisplayPhase {
```
nickfarrow, line 160: What makes this a "phase", because it will be used in other workflows like checking shares?

nickfarrow: Ah "Often confirmations contain what we call a "phase" which is both the data that describes the action and what will be passed back"

device/src/graphics/widgets/backup_result_screen.rs
```diff
@@ -0,0 +1,115 @@
...
+use u8g2_fonts::U8g2TextStyle; // Assuming these exist in embedded_iconoir
+
+pub struct BackupResultScreen {
+    area: Size,
+    ok: bool,
+    // Only used for invalid backup (to offer a retry)
+    try_again_rect: Rectangle,
+}
+
+impl BackupResultScreen {
```
nickfarrow, line 16: Is this widget for a later PR? It isn't integrated yet

nickfarrow:
Tested a few workflows, works great! Only minor comments above, removed some old code in eab71e584cc8d575a5d2d1b5c39537fc9e0dce2b

nickfarrow (approved):

frostsnap_coordinator/src/firmware_upgrade.rs
LLFourn: isn't that what we are doing by just logging them?

frostsnap_core/src/sign_task.rs
LLFourn: In what way are they not fixed?

device/src/graphics/widgets/backup_result_screen.rs
LLFourn: Good catch. No this was some vibe coded thing I did at some point that got committed. I'll remove it.

frostsnap_coordinator/src/firmware_upgrade.rs
nickfarrow: woops! yes, i was thinking of device panics

frostsnap_core/src/sign_task.rs
nickfarrow: Ah yes they are fixed in KeyPurpose :+1:

PR #227, opened 2025-03-04
[app] Text bikesheading for backup workflow

LLFourn:
Ideas for improving the text of the workflow.

In other ideas, I think that we should use a stepper:

https://api.flutter.dev/flutter/material/Stepper-class.html

To do the:

1. Plug device in
2. Confirm to show backup on the device
3. Check the backup (skip-able)
4. unplug device so you can store it
5. done: reminder to keep it secure.


nickfarrow (approved):
Rebased, added minor wording and layout changes in c7874ff612b8a5236c7b467156fd354fec86f715

PR #228, opened 2025-03-07
Proptest state machine

LLFourn:
[test] Use proptest-state-machine test

This tests (most) valid state transitions on the devices and coordinator.
Testing invalid state transitions and checking an error is thrown is
left to later work.

It found several bugs:

## Erroneous `&mut`

https://github.com/frostsnap/frostsnap/blob/de85bfe2d69912e41f1a9275c59199933cf9f2cf/frostsnap_core/src/coord_nonces.rs#L42-L47

This method did not need `&mut self`. What's worse is the method actually did mutate `self` accidentally using the `entry` API while doing a lookup. This mutation created an empty entry in a `BtreeMap` which was harmless. It was caught because we always apply persisted mutations at the end of every test in the `Drop` implementation to check if we end up in the same spot.

## Overwriting nonces 1

https://github.com/frostsnap/frostsnap/blob/de85bfe2d69912e41f1a9275c59199933cf9f2cf/frostsnap_core/src/device.rs#L187-L198

There may be a difference in how many nonce streams the coordinator wants to open and how many the device has in total. So if the coordinator opened 8 and the device only had 4 slots the device would respond positively to all 8 requests but the last four would simply erase the first four. The device then would panic when the coordinator asked it to sign with one of the erased nonce streams. Instead we should just respond affirmatively to first four and ignore the rest. 

## Overwriting nonces 2

https://github.com/frostsnap/frostsnap/blob/de85bfe2d69912e41f1a9275c59199933cf9f2cf/frostsnap_core/src/coord_nonces.rs#L121-L148

Even after fixing bug 1 there is a problem with the combined logic on the device and the coordinator where the device responding affirmatively to the first four nonce streams would still break since the next time the coordinator sync'd nonce streams it would ask for the established nonce streams after throwing in new ones. Therefore the device responded affirmatively for the new ones which overwrote the old ones and then ignored the requests to re-open the old ones. The fix was just to get the device to prioritize responding affirmatively to those streams it already has opened.

## Nonce replenishing after a signing session has started brought a consumed nonce back to life

:warning: this was the bug that I was actually searching for

The coordinator should not let the device refresh nonces that it has already reserved for a signing session but it wasn't doing this properly. The logical problem was that we were consuming nonces too early (when the coordinator sent out the request) and then opening up the possibility of the device innocently adding them back. Instead we don't consume the nonces until the sign session is finished. A sign session freezes the state of each nonce stream it's using so there's no harm in waiting until it's over to delete the nonces it was using. It can also restore them if it never actually signed.

## Deleting a key didn't delete the associated sign sessions

https://github.com/frostsnap/frostsnap/blob/de85bfe2d69912e41f1a9275c59199933cf9f2cf/frostsnap_core/src/coordinator.rs#L309-L312

Here you should also delete signing sessions associated with they key.


frostsnap_core/tests/proptest.rs
```diff
@@ -0,0 +1,723 @@
...
+                Some(session) => {
+                    session.sent_req_to.contains(device_id)
+                        && session.key_index < state.finished_keygens.len()
+                        && !state.finished_keygens[session.key_index].deleted
+                        && !session.got_sigs_from.contains(device_id)
+                        && !session.canceled
+                }
+                None => false,
+            },
+            &Transition::CDeleteKey { key_index } => key_index < state.finished_keygens.len(),
```
nickfarrow, line 404: Could also check `!state.finished_keygens[key_index].deleted`, but I tested and it doesn't bring any failures

nickfarrow (approved):

PR #231, opened 2025-03-11
Transaction Details

evanlinjin:


nickfarrow:
Could we reuse the feerate times from the send workflow to display an ETA like `Receiving (ETA ~10m)` and also next to `Comfirmations None (ETA ~10m)`?

If the recipients are empty, maybe we could indicate internal transfer with text? 

#### Fee display:
For incoming payments, the fee amount isn't so useful and could even confuse people who are new to bitcoin (did they pay it?). We could label as "Sender fee".

The fee rate is more useful for unconfirmed payments (with RBF later on), but the nominal fee amount is more useful for confirmed transactions you have sent.

So I am thinking we might want to display the fee rate by default, and expand down to show the fee amount underneath upon user clicking?

Or alternatively: show the feerate for unconfirmed transactions, and the fee amount for confirmed transactions.

What do you think?


Tested with a many-recipient transaction, works great. Also rebroadcast button is nice - will be sweet to put RBF button there too once we get to it.

evanlinjin:
@nickfarrow Thank you for the review. I agree with all your points made. Would you be happy if I address them in a separate PR? Esp. the confirmation time estimates as we will need to rework some more modelling for that to work nicely!

evanlinjin:
Just did a rebase

LLFourn (approved):
ACK 186b3b114563e79ccd0a43d49856dd3f6d6624d5

LLFourn:
merging. Added an issue #234 to add @nickfarrow's improvements. Found #233 while reviewing.

PR #232, opened 2025-03-12
Signing dialog v2

evanlinjin:


LLFourn:
image

This dialog needs to inform the user that it has been successfully sent. The "Sending..." state should be at least be "Confirming...". For example from [here](https://bitcoin.design/guide/daily-spending-wallet/sending/):

image

LLFourn:
image
I don't think green and gold works well looking at it again. The yellowy orange sort of indicates a "warning" or "something isn't complete yet". The outgoing arrows should be reddish I think. Maybe I pinkish red like #FF6B6B

LLFourn:
image

From this view, compelted signatures should not disappear them from the list. They should just have a tick or something next to them.

LLFourn (changes requested):
This is a big improvement. Main remaining work is to do the blackout thing we discussed. Can be done in another PR.

nickfarrow:
Design:
* Glow is super nice.
* "Cannot calculate fee: no amount" looks like an error, I think better if we just hide in this state. OR "Fee preview will be visible after entering an amount to send"
* Could be nice to show devices that have already signed, rather than disappearing?
	* Nice progress to have alongside the circle.
	* Record of what has occurred so far. "Oh no i dont have access to this device, let me create a new txn with the same set of devices and one different."
* I don't think we should say "Disconnected" before connecting a device, it sounds like the verb rather than the adjective. I think blank there would be fine?
* Requesting signature is good. Could also be more explicit: "Verify and sign on device".
* Note: coin selection doesn't exclude "reserved UTXOs" with a pending sign task.
* Slick signing dialog isn't applied to PSBT signing or plain message signing (not a concern for now, unless it's easy to integrate there?). But PSBT txns do show up in the list! Need to force close app to quit dialog without cancelling. Later PR probably best.

- Agree w Lloyd with respect to gold colour on send arrows.
- Agree "Confirming..." or even "Pending..." is good for broadcasted sends. 

Dialog is incredible, and fantastic work with the pending sign tasks. One amazing thing that I've only just now appreciated is the ability to "queue up" a bunch of separate transactions, and then go sign them all at once. Even though this is possible with PSBTs today, this UX is next level.

Haven't reviewed code

frostsnapp/native/src/coordinator.rs
```diff
@@ -559,36 +566,204 @@ impl FfiCoordinator {
...
+        &self,
+        super_wallet: api::SuperWallet,
+        key_id: KeyId,
+    ) -> SyncReturn<Vec<api::SignedTxDetails>> {
+        let coord = self.coordinator.lock().unwrap();
+        let super_wallet = &*super_wallet.inner.lock().unwrap();
+        let txs = coord
+            .finished_signing_sessions()
+            .iter()
+            .filter(|(_, session)| dbg!(session.key_id == key_id))
```
nickfarrow, line 707: Noticed two debugs here, maybe can remove after development, or use `event!` like you have elsewhere for the wildcard one at the end of the match below

LLFourn:
@evanlinjin it looks like Nick and I have the same suggestions. With:

> I don't think we should say "Disconnected" before connecting a device, it sounds like the verb rather than the adjective. I think blank there would be fine?

I agree that disconnected is not good. "connect to sign" or an icon implying that would be good.

nickfarrow:
Made some very minor text changes in 33857115a2603e483f0c947865a8f0c7d73b4983:
* Remove "Disconnected", though this leaves the device names greyed out with nothing besides them, ok? wdyt?
image
* Signature(s) -> Signature, will almost always be more than one. I think it is ok to still show the plural if user has just one left since we show the history.
* Similarly "Connect device(s) to finish signing" -> "Connect a device to sign".

nickfarrow (approved):

LLFourn (approved):

PR #236, opened 2025-03-26
New Restoration workflow

LLFourn:
The two commits should be reviewed separately:

1st one: Implement a new backup restoration workflow. Works only with device shares for now. Also can't yet restore a share after you've finished the restoration workflow (might fix this in this PR unless it's merged fast)
2nd one: refactor backup run modelling:

```
1. Backup state was being persisted without AccessStructureId which
means it will be a mess if we add multiple master access structures.
2. Backup mutations were messy because they were applied differently at
runtime than when loading from persistence.
3. maybeStartBackupRun is called all the time. Now we only call it once when
the backup dialog is dismissed.
4. No more isShowingCreatedWalletDialog
```


device/src/esp32_run.rs
```diff
@@ -82,7 +84,7 @@ where
         };
 
         // The event log gets the reset of the sectors
-        let mut event_log = EventLog::new(partitions.nvs);
+        let mut event_log = EventLog::new(share_partition, partitions.nvs);
```
nickfarrow, line 87: We should rename `EventLog`, maybe `DataPartitions` or `DeviceDatabase` `DeviceData`.

frostsnap_comms/src/lib.rs
```diff
@@ -370,13 +379,29 @@ impl<B: Gist> Gist for DeviceSendMessage<B> {
...
+        firmware_digest: Sha256Digest,
+    },
+    SetName {
+        name: String,
+    },
     DisconnectDownstream,
     NeedName,
     _LegacyAckUpgradeMode, // Used by earliest devices
     Misc(CommsMisc),
+    Announce2 {
```
nickfarrow, line 395: Is this a temporary or longer-term approach to message versioning?
Will we bundle these all up into a higher enum on the model level?

frostsnap_coordinator/src/enter_physical_backup.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    fn disconnected(&mut self, id: DeviceId) {
+        let mut changed = false;
+        if self.chosen_device == Some(id) {
+            self.chosen_device = None;
+            self.sent_req = false;
+            changed = true;
+        }
+        changed |= self.blank_connected.remove(&id);
```
nickfarrow, line 76: TIL `|=`

frostsnap_coordinator/src/enter_physical_backup.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn connected(&mut self, id: DeviceId, is_blank: bool) {
+        if is_blank && self.blank_connected.insert(id) {
+            self.emit_state();
+        }
+    }
+
+    fn process_to_user_message(&mut self, message: CoordinatorToUserMessage) {
+        match message {
+            CoordinatorToUserMessage::PromptRecoverPhysicalBackup(physical_backup_phase) => {
+                self.finished = Some(*physical_backup_phase);
```
nickfarrow, line 114: I think `self.finished` can just be a `bool`

frostsnap_coordinator/src/wait_for_recovery_share.rs
```diff
@@ -0,0 +1,107 @@
...
+        self.sink.send(self.state.clone());
+    }
+}
+
+impl UiProtocol for WaitForRecoveryShare {
+    fn cancel(&mut self) {
+        self.abort = true;
+    }
+
+    fn is_complete(&self) -> Option<crate::Completion> {
```
nickfarrow, line 40: `WaitForRecoveryShare` never completes? Makes sense I think

frostsnap_coordinator/src/wait_for_recovery_share.rs
```diff
@@ -0,0 +1,107 @@
...
+        let mut out = vec![];
+        let need_to_send_to = self
+            .state
+            .connected
+            .difference(&self.sent_request_to)
+            .cloned()
+            .collect::<Vec<_>>();
+
+        for device_id in need_to_send_to {
+            out.push(CoordinatorSendMessage {
```
nickfarrow, line 69: nit: could use your nice `CoordinatorSendMessage::to` here, and in many other places in the codebase

frostsnap_core/src/coordinator.rs
```diff
@@ -1422,6 +1524,48 @@ impl FrostCoordinator {
...
+        device_id: DeviceId,
+        access_structure_ref: AccessStructureRef,
+        index: PartyIndex,
+    ) -> bool {
+        let already_got_under_key = self
+            .keys
+            .get(&access_structure_ref.key_id)
+            .and_then(|coord_key| {
+                let access_structure_id = access_structure_ref.access_structure_id;
+                Some(
```
nickfarrow, line 1567: nit: Don't need `Some` here,
```
      let already_got_under_key = self
          .keys
          .get(&access_structure_ref.key_id)
          .and_then(|coord_key| {
              let access_structure_id = access_structure_ref.access_structure_id;
              coord_key
                  .get_access_structure(access_structure_id)
                  .map(|access_structure| access_structure.contains_device(device_id))
          })
          .unwrap_or(false);
```
but pretty much identical

frostsnap_core/src/coordinator.rs
```diff
@@ -1170,93 +1178,192 @@ impl FrostCoordinator {
...
+            restoration_id,
+            device_id: recover_share.held_by,
+            access_structure_ref: Some(recover_share.held_share.access_structure_ref),
+            share_image: recover_share.held_share.share_image,
+        });
+
+        Ok(())
+    }
+
+    pub fn check_recover_share_compatible_with_restoration(
```
nickfarrow, line 1262: review todo: check if we stop at T, or, what happens at T+1

frostsnap_core/src/coordinator.rs
```diff
@@ -140,25 +186,27 @@ impl FrostCoordinator {
...
         if let Some(reduced_mutation) = self.apply_mutation(&mutation) {
+            event!(Level::DEBUG, gist = mutation.gist(), "mutating");
             self.mutations.push_back(reduced_mutation);
+        } else {
+            event!(Level::DEBUG, gist = mutation.gist(), "ignoring mutation");
         }
     }
 
     pub fn apply_mutation(&mut self, mutation: &Mutation) -> Option<Mutation> {
         fn ensure_key<'a>(
```
nickfarrow, line 198: Note: this never updates a key, so if we try restore the same key under a different purpose, it will not be stored. Probably desired behaviour

frostsnap_core/src/device.rs
```diff
@@ -796,45 +804,63 @@ impl<S: NonceStreamSlot + core::fmt::Debug> FrostSigner<S> {
...
-        let share_index = phase.share_image.share_index;
-        let access_structure_ref = phase.access_structure_ref;
-        let ok = match phase.check_entered_backup(secret_share) {
-            Some(complete_share) => {
-                self.tmp_loaded_backups.push(complete_share);
-                true
+        let mut ret = vec![];
+
+        match phase {
+            LoadBackupPhase::Known(load_known_backup) => {
```
nickfarrow: The loading of known/unknown backups seems quite intertwined with this this design, but in my mind they are quite separate tasks except for the entering share on device stage. Will keep this in mind as I continue reviewing.

frostsnap_core/src/message.rs
```diff
@@ -198,6 +202,7 @@ pub enum DeviceToCoordinatorMessage {
         share_index: PartyIndex,
         success: bool,
     },
+    LoadBackupResult(Box<EnteredPhysicalBackup>),
```
nickfarrow: Always unknown, s/LoadBackupResult/LoadUnknownBackupResult/

nickfarrow: Maybe also the case for `LoadPhysicalBackup`? Will see how this looks after later commits

frostsnap_core/src/coordinator.rs
```diff
@@ -1174,93 +1182,192 @@ impl FrostCoordinator {
...
+        let state = self
+            .restorations
+            .get(&restoration_id)
+            .ok_or(RestorationError::UnknownRestorationId)?;
+
+        let root_shared_key = state
+            .clone()
+            .access_structure
+            .interpolate()
+            .ok_or(RestorationError::NotEnoughShares)?;
```
nickfarrow: In this design it looks like we can recover with >T shares, which in this case, this interpolate and underlying `from_share_images` doesn't detect/handle the case that one of them is incompatible. Ties into the wider design question of how we approach incompatible backups.

frostsnap_core/src/coordinator.rs
```diff
@@ -1726,18 +1966,19 @@ impl Gist for Mutation {
...
-            RecoverShare { .. } => "RecoverShare",
             DeleteKey(_) => "DeleteKey",
             NewNonces { .. } => "NewNonces",
             NewSigningSession { .. } => "NewSigningSession",
             CloseSignSession { .. } => "CloseSignSession",
             ForgetFinishedSignSession { .. } => "ForgetFinishedSignSession",
             GotSignatureSharesFromDevice { .. } => "GotSignatureSharesFromDevice",
             SentSignReq { .. } => "SentSignReq",
+            NewRestoration { .. } => "NewRestoration",
+            RestorationProgress { .. } => "ResotrationProgress",
```
nickfarrow: typo Restoration

device/src/esp32_run.rs
```diff
@@ -68,7 +69,7 @@ where
             }
         };
 
-        let _reserved = partitions.nvs.split_off_front(2);
+        let share_partition = nvs_parition.split_off_front(2);
```
nickfarrow, line 72:
```
let mut n_nonce_sectors = partitions.nvs.n_sectors().div_ceil(2);
```
Bug creates two additional nonce sectors which overwrites eventlog sectors (name disappears)

device/src/esp32_run.rs
```diff
@@ -52,12 +51,14 @@ where
...
             mut hmac_keys,
+            model,
         } = self;
 
         ui.set_busy_task(ui::BusyTask::Loading);
         let flash = RefCell::new(FlashStorage::new());
-        let mut partitions = crate::partitions::Partitions::load(&flash);
-        let header_flash = FlashHeader::new(partitions.nvs.split_off_front(2));
+        let partitions = crate::partitions::Partitions::load(&flash);
+        let mut nvs_parition = partitions.nvs.clone();
```
nickfarrow, line 60: typo: "parition"

frostsnap_coordinator/src/enter_physical_backup.rs
LLFourn: I'm not sure how it was but it can't be now.

frostsnap_comms/src/lib.rs
LLFourn: > Is this a temporary or longer-term approach to message versioning?

long term

>  Will we bundle these all up into a higher enum on the model level?

What's the "model level"?

I should probably delete this enum variant since I didn't end up using `Announce2` for much.

device/src/esp32_run.rs
LLFourn: I was looking at this but in the end the API is just an event log. I'll change  it to MutationLog.

frostsnap_core/src/coordinator.rs
LLFourn: It does handle it. If one is incompatible the threshold won't be what was expected.

frostsnap_core/src/coordinator.rs
LLFourn: I made a macro to generate these now.

frostsnap_core/src/message.rs
LLFourn: Yes I will fix the naming here.

frostsnap_core/src/coordinator.rs
LLFourn: Not sure what the code you're referring to was at the time. Take a look at the latest version.

frostsnap_core/src/coordinator.rs
LLFourn: Continuing beyond `t` is always useful. Hopefully we can automatically tell that one of the shares is invalid when you have `t`. Adding more (correct) shares can tell you which one. The polynomial id thing is of course ideal for distinguishing related and unrelated shares but that doesn't seem to be crucial to the outcome of the restoration.

frostsnap_core/src/device.rs
LLFourn: Telling the user to enter a backup is the same between checking a backup and entering it for restoration. I didn't check if "checking" the backup still works while testing come to think of it...

nickfarrow:
Really incredible craftsmanship for such a complex part of our system. I don't have much to say on the architecture (will try to revisit after some thought). I have some feedback on the behavior to start, later I'll suggest some some code renaming (maybe we best if we go back and forth on call), and some minor UI touchups & code cleanup which I'll make in a commit.

## Behaviour

Restoration from devices currently does not work if multiple are plugged in at once, I think people will do this.

A device is considered "Not blank" after only giving a name -> somewhat annoying at the moment.
- I give a name in the dialog, i go back for a moment, then I go to enter physical backup again. Not blank! Can we mark a device as non-blank based on it having a share or loaded share backup?
* I guess this will be partially resolved once we only act on naming during keygen or restoration.

I can recover the same wallet twice (e.g. different networks) and only the first wallet remains. Reasonable I think, I noticed this in the `NewKey` previously. We could consider showing an error on the restoration page.

Check backup bug: Entered a backup not belonging to the device but belonging to the same key. "Your backup is valid. Done!". This is likely from a previous PR, I can investigate this.

Currently you can load two unrelated backups that have the same share index (user error) -> Panics when pressing `✔ Restore` "Not enough shares to restore". Can we prevent this earlier? I can investigate this also.

Downstream devices now show red circle instead of yellow when coordinator has not been started. I can investigate this also.

Remind me, what did we decide on the different key name problem? Restore from devices first, then restore from physical backups? If you restore physical backup, then plug in a device -> doesn't match.

Let me know if you would like me to look into some more of these!

frostsnap_coordinator/src/enter_physical_backup.rs
```diff
@@ -0,0 +1,132 @@
...
+            _ => false,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct EnterPhysicalBackupState {
+    pub device_id: DeviceId,
+    pub entered: Option<restoration::PhysicalBackupPhase>,
+    /// null if the user is entering the backup not to save but to check it
```
nickfarrow, line 129: What's this referring to? Leftover comment?

nickfarrow:
I'm having a lot of trouble reasoning about the UiStack in its current form, previously we had a much simpler model where only one protocol was active at any one time (bringing the limitations that you mention). I am finding it harder to follow the logic of message flow and protocol completion.

I'm experiencing a bug with finishing restoration: plug two devices in and enter backups on each of them. You can't finish restoration, we never seem to complete `exit_recovery_mode`. With the stack approach, the protocol on the top of the stack receives messages first, can multiple protocols consume the same message? It's also tricky to debug which protocol is handling which message (there may be multiple instances of the same protocol?)

nickfarrow:
Some things I noticed while working on 22f5eb19c4cce08ed8ebedaf3b3dd69b853dd8cc:

We had an outline for `already_held` logic on the rust side in `WaitForRecoveryShare`, but we were handling it separately on the dart side based on the output of 

```
compatibility = coord.restorationCheckShareCompatible(
  restorationId: widget.continuing!,
  recoverShare: detectedShare,
);
```

After removing it on the rust side, i've started to wonder whether we should try house this logic on the rust side (as you had sketched out). The primary logic I added was a map from **all** recoverable shares, not just the first one, and tracking ones we've already received in a set of `DeviceIds`. This allowed for recovery when multiple devices were connected at the same time.

I also noticed that there's large overlap in the errors of `RestorationError` and `RestoreRecoverShareError`. I was thinking the `RestoreRecoverShareError`s could be inside `RestorationError` enum. Scenarios that we currently/would differentiate on:

* We'll disallow `AlreadyGotThisShare` when restoring, but may allow it when adding a device into the key (we currently allow this, but could force the user to remove the old device before adding in a device with the duplicate share).  
* We don't appear to care about name or purpose mismatches when adding device to existing key(?).
* Duplicate index error (`AlreadyGotShareIndex`) is superseded by `ShareImageIsWrong` when adding device into an existing key.

frostsnap_core/src/lib.rs
```diff
@@ -257,15 +258,18 @@ impl CoordShareDecryptionContrib {
...
+    /// knowledge of the main polynomial.
     pub fn for_master_share(
         device_id: DeviceId,
+        share_index: PartyIndex,
         shared_key: &SharedKey<Normal, impl ZeroChoice>,
     ) -> Self {
         Self(
             prefix_hash("SHARE_DECRYPTION")
                 .add(device_id.0)
+                .add(share_index)
```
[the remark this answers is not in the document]
LLFourn: Correct. The only one that is hard to know is (1).

If a device had more than one share it would have the same encryption contrib for both shares. That's wrong so I fixed it. It doesn't mean it's different for different devices -- that's what the `DeviceId` does.

frostsnap_core/src/coordinator/restoration.rs
```diff
@@ -0,0 +1,986 @@
...
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct State {
+    pub(super) restorations: BTreeMap<RestorationId, restoration::RestorationState>,
+    /// This is where we remember consolidations we need to do from restorations we've finished.
+    pub(super) pending_physical_consolidations: BTreeSet<PendingConsolidation>,
+    /// For when we ask a device to enter a backup and we plan to immediately consolidate it because
+    /// we already know the access structure.
+    tmp_waiting_consolidate: BTreeSet<PendingConsolidation>,
+
+    tmp_waiting_save: BTreeMap<(DeviceId, ShareImage), RestorationId>,
```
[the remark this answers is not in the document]
LLFourn: It's modeled like that because we are waiting for a particular device to save a particular share image which is associated with a particular `RestorationId`. It does allow for multiple shares per signer because the share image is part of the key rather than the value.

frostsnap_coordinator/src/signing.rs
```diff
@@ -128,8 +131,11 @@ impl UiProtocol for SigningDispatcher {
         self.emit_state();
     }
 
-    fn connected(&mut self, device_id: DeviceId) {
-        if !self.got_signatures.contains(&device_id) && self.targets.contains(&device_id) {
+    fn connected(&mut self, device_id: DeviceId, state: DeviceMode) {
+        if !self.got_signatures.contains(&device_id)
+            && self.targets.contains(&device_id)
+            && state == DeviceMode::Ready
```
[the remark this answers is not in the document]
LLFourn: Yes because signing can only happen once a device is "ready".

frostsnap_coordinator/src/enter_physical_backup.rs
```diff
@@ -0,0 +1,135 @@
...
+use tracing::{event, Level};
+
+use crate::{DeviceMode, Sink, UiProtocol};
+
+pub struct EnterPhysicalBackup {
+    sink: Box<dyn Sink<EnterPhysicalBackupState>>,
+    enter_physical_id: EnterPhysicalId,
+    chosen_device: DeviceId,
+    sent_req: bool,
+    connected: bool,
```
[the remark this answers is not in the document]
LLFourn: Yes I think you can probably save a bool with the first approach. 

> Also noticed we don't have this small mistruth of connected: false in ::new, though this is handled in start_protocol. Probably ~inconsequential but just something I noticed.

I don't know what's wrong. It's not connected until it's told that the device is connected.

frostsnap_core/src/device/restoration.rs
```diff
@@ -0,0 +1,412 @@
...
+                    }),
+                ))])
+            }
+            &CoordinatorRestoration::SavePhysicalBackup {
+                share_image,
+                threshold,
+                purpose,
+                ref key_name,
+            } => {
+                if let Some(secret_share) = self.restoration.tmp_loaded_backups.remove(&share_image)
```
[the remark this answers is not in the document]
LLFourn: yes.

frostsnapp/lib/device_setup.dart
```diff
@@ -122,10 +117,6 @@ class _DeviceNameField extends State<DeviceNameField> {
                 );
               },
             );
-
-            if (result == null) {
```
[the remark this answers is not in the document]
LLFourn: Good catch. I can't see why I did this but I remember doing it for some reason. Hopefully it can be reverted.

LLFourn:
@nickfarrow can you spin your last three commits which are substantial into their own PRs.

1. [78f9a1d](https://github.com/frostsnap/frostsnap/pull/236/commits/78f9a1db514817363efff0d1c79774b3ac1c1eb5)  indeed the current behavior is a bug but distinguishing between these behaviors is a new feature than belongs in its own PR to master. 
2. [295e16d](https://github.com/frostsnap/frostsnap/pull/236/commits/295e16d7419ba2a9a542a83ed678f5e37d52c01b) I think should be a PR to this branch. There might be a better solution: just look through all the devices to find one with good compatibility rather than storing this "processed" list.
3. [22f5eb1](https://github.com/frostsnap/frostsnap/pull/236/commits/22f5eb19c4cce08ed8ebedaf3b3dd69b853dd8cc) this can be another PR to this branch. Like I said I think we should allow duplicates for now but the modelling should be fixed to not let you proceed until you have `t` unique shares.

> When working on https://github.com/frostsnap/frostsnap/commit/22f5eb19c4cce08ed8ebedaf3b3dd69b853dd8cc I noticed that there's large overlap in the errors of RestorationError and RestoreRecoverShareError. I was thinking the RestoreRecoverShareErrors could be inside RestorationError enum. Scenarios that we currently/would differentiate on:

These errors are for very different things -- one is for adding a new share to recovery and one is for finishing a recovery once it has enough shares.

> I'm experiencing a bug with finishing restoration: plug two devices in and enter backups on each of them. You can't finish restoration, we never seem to complete exit_recovery_mode. With the stack approach, the protocol on the top of the stack receives messages first, can multiple protocols consume the same message? It's also tricky to debug which protocol is handling which message (there may be multiple instances of the same protocol?)

Just put `dbg!` inside the stack to figure out which one got it. No only one can consume it and ui protocols should return `false` to indicate they didn't consume it.

frostsnap_coordinator/src/signing.rs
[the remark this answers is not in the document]
nickfarrow: Ah because when you were connecting devices to sign that had not consolidated restoration, but does UI protocol stack handle that now? Worth also adding on `DisplayBackupProtocol`?

Also related: KeyGen currently does not look at this, and it is possible to overwrite a non-blank device. Though these changes may belong in separate PR

frostsnapp/lib/device_setup.dart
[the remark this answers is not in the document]
nickfarrow: I noticed that in the current state it was possible to pop the dialog and the device would sit in a `Naming` state. Maybe you saw this and were removing manual coordinator cancelling altogether? It seems somewhat necessary for now, unless we make naming a UI protocol. See 7cee7183d8dfd7f7e5d0ba95a80b7c9d1d8f6598

nickfarrow:
What do you think about adding back buttons to the UI screens of "Enter wallet name", "Enter recovery threshold", and "Name this device"? During testing there were a few times where I would have liked them to be there, but maybe that was because I was being quick. E.g. mixing up device/wallet name

Do you mind if we change all the UI occurrences of "Recover" to "Restore"? I think it sounds more optimistic and less scary.

LLFourn:
> What do you think about adding back buttons to the UI screens of "Enter wallet name", "Enter recovery threshold", and "Name this device"? During testing there were a few times where I would have liked them to be there, but maybe that was because I was being quick. E.g. mixing up device/wallet name

Yeah it should be more fleshed out. I plan to leave this to @evanlinjin but I guess I could try and do this. I wonder if there is smarter way than manually managing the navigation myself i.e. just using Navigator.pop() to go back.

> Do you mind if we change all the UI occurrences of "Recover" to "Restore"? I think it sounds more optimistic and less scary.

Sure I'll try and do this now.

evanlinjin:
@evanlinjin I would like to be the person to flesh out the UI.

Since we don't have formal UI guidelines written down yet-and let's be honest, our current design system is mostly just "Evan's gut feeling"-it'll probably be more efficient if I take on the Ul refinement tasks directly.

Happy to sync on design intent anytime, but I can move faster if I'm the one pushing pixels for now.

nickfarrow (approved):
Tested and reviewed. Added 8e6ed66a207c5e0218e4612f3866400fd003dca8 which bubbles some restoration errors onto the snackbar, such as trying to restore an already existing key from devices. Your commit a3b8097ce2086de10bcd35f77e254eaaacb20ad3 solves https://github.com/frostsnap/frostsnap/pull/246.

I couldn't see an easy way to allow named-but-blank devices to be used in restoration. Naming currently exist in isolation and needing to wipe a device which only has a name still feels a bit clunky. We can either fix this with a future PR that links initial naming of a device to keygen or restoration (ideally atomically persisted), or we could consider a more involved determination if a device is blank (belongs to no access structures, no held shares, or signal through a device message)

Hmm latest clippy wants us to put Box on `Core(Box<frostsnap_core::message::CoordinatorToDeviceMessage>)` and also on

```
ToDevice {
    message: Box<CoordinatorToDeviceMessage>, 
     ..
}
```

Makes a bit of a mess.. Ignore warning?

LLFourn:
Ok 111fac4640fed1151a96a5862a860bcd35b2be64 was probably the last commit from me. Notes:

1. Yes @nickfarrow  the UI is not great because of named and blank being a possibility. @evanlinjin will have to take a look at this and figure out how to give the same treatment here as he gave in keygen.
2. I stop trying to figure out if the share "already exists" in another key or restoration. We need to do this but it I want to make it work for every situation which isn't quite possible atm because you can't just look up existing shares by the share image. I will take another look at this once we see if we have integrated a "verification polynomial" which will make things easier. A verification polynomial allows us to have a look up of share images without having to worry about information that the share images can leak (i.e. root key).

nickfarrow (approved):
Documenting the fact that: the change from `u32` to `u8` in prefix_hash calculation is a breaking change for key_ids and access structures. In case we ever come across this again.

Tested! Works really well. Congratulations!

PR #239, opened 2025-04-09
Receive Flow v2

evanlinjin:


frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -147,23 +178,43 @@ impl CoordSuperWallet {
...
         AddressInfo {
             index,
             address: bitcoin::Address::from_script(spk, self.network).expect("has address form"),
             external: true,
-            used: self
-                .tx_graph
-                .index
-                .is_used((master_appkey, keychain), index),
+            used,
+            shared,
```
nickfarrow, line 214: One of fresh and shared is now redundant, since `fresh = !shared`

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.spks.is_empty()
+    }
+}
+
+/// Reserves spks.
+#[derive(Debug, Clone)]
+pub struct ReservedSpks {
+    spks: HashSet<ScriptBuf>,
+    /// So that our changesets can be fully monotone.
+    next_seq: u64,
```
nickfarrow, line 30: We use `next_seq` instead of `seq` so we don't have to start out from `-1`? Using `seq` would (very) minorly simplify in other places.

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
```diff
@@ -0,0 +1,130 @@
...
+use bdk_chain::{
+    bitcoin::{Script, ScriptBuf},
+    rusqlite_impl::migrate_schema,
+    Impl, Merge,
+};
+use rusqlite::named_params;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Default)]
+pub struct ChangeSet {
+    spks: BTreeSet<(u64, ScriptBuf, bool)>,
```
nickfarrow, line 12: could comment that this bool is `is_add` to add/remove changes to changeset

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
```diff
@@ -0,0 +1,130 @@
...
+            self.next_seq += 1;
+            changeset.spks.insert((seq, spk.as_ref().to_owned(), false));
+        }
+        changeset
+    }
+}
+
+impl ChangeSet {
+    /// Schema name for the changeset.
+    pub const SCHEMA_NAME: &'static str = "fs_reserved_spks";
```
nickfarrow, line 75: This will be versioned like `fs_reserved_spks_v2` if we were to add labels or something?

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    pub fn init_sqlite_tables(db_tx: &rusqlite::Transaction) -> rusqlite::Result<()> {
+        migrate_schema(db_tx, Self::SCHEMA_NAME, &[&[&Self::schema_v0()]])
+    }
+
+    pub fn from_sqlite(db_tx: &rusqlite::Transaction) -> rusqlite::Result<Self> {
+        let mut changeset = Self::default();
+        let mut select_statement = db_tx.prepare(&format!(
+            "SELECT spk, seq FROM {}",
```
nickfarrow, line 97: Should we be reading reserved from here also? Instead of assumed true. This wouldn't cause bugs currently since we are always only reserving.

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
evanlinjin: How would having `seq` instead of `next_seq` simplify things?

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
evanlinjin: No. We would create a `schema_v1()` method which migrates from `schema_v0()`, then add it to `migrate_schema`.

frostsnapp/lib/main.dart
```diff
@@ -128,7 +129,7 @@ class _MyAppState extends State<MyApp> {
   void _setColorTheme() =>
       colorScheme = ColorScheme.fromSeed(
         brightness: Brightness.dark,
-        seedColor: Color(0xFF1595B2),
+        seedColor: seedColor,
```
nickfarrow, line 132: good call moving this into theme!

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
evanlinjin: Good catch. We should only be selecting spks that have `reserved` as `true` here.

You are right that this would not cause problems right now since we are only reserving.

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
evanlinjin: "reserve/unreserve spks".

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -185,6 +236,91 @@ impl CoordSuperWallet {
...
+                    Some(spk) => spk,
+                    None => return Ok((false, (tx_graph_changeset, reserved_spks_changeset))),
+                };
+                reserved_spks_changeset.merge(reserved_spks.reserve(spk));
+                let reserved = reserved_spks_changeset.is_empty();
+                Ok((reserved, (tx_graph_changeset, reserved_spks_changeset)))
+            })
+    }
+
+    pub fn unmark_address_shared(
```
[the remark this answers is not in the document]
evanlinjin: You are right we don't use this now. I'm also not sure if this is useful haha.

frostsnapp/native/src/api.rs
```diff
@@ -94,39 +97,130 @@ impl TxOutInfo {
...
+        SyncReturn((|| -> Option<i64> {
+            let owned_inputs_sum: i64 = self
+                ._sum_inputs(Some(&self.is_mine))?
+                .try_into()
+                .expect("net spent value must convert to i64");
+            let owned_outputs_sum: i64 = self
+                ._sum_outputs(Some(&self.is_mine))
+                .try_into()
+                .expect("net created value must convert to i64");
+            Some(owned_outputs_sum.saturating_sub(owned_inputs_sum))
```
nickfarrow, line 210: Nice I think this is a good decision instead of throwing an error

frostsnapp/lib/wallet_receive.dart
evanlinjin, line 1: Always reserving the address when we exit the "show QR" screen will be unintuitive when the user hits the back/ESC button.

I think we should make the dialog `barrierDismissible: false` so that it forces the user to hit "Done".

frostsnapp/lib/wallet_tx_details.dart
```diff
@@ -331,21 +336,22 @@ class _TxDetailsPageState extends State<TxDetailsPage> {
...
-          centerTitle: true,
-          forceMaterialTransparency: true,
-          leading: IconButton(
-            onPressed: () => Navigator.pop(context),
-            icon: Icon(Icons.close),
-          ),
-          actionsPadding: EdgeInsets.symmetric(horizontal: 10.0),
-          automaticallyImplyLeading: false,
-        ),
+        //SliverAppBar(
```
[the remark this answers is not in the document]
evanlinjin: Can rm.

frostsnapp/lib/main.dart
evanlinjin: Light theme option soon!

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -185,6 +236,91 @@ impl CoordSuperWallet {
...
+            .multi(&mut self.reserved_spks)
+            .mutate(&mut db, |tx_graph, reserved_spks| {
+                let mut tx_graph_changeset = WalletIndexedTxGraphChangeSet::default();
+                let mut reserved_spks_changeset = ReservedSpksChangeSet::default();
+                match tx_graph
+                    .index
+                    .reveal_to_target((master_appkey, keychain), derivation_index)
+                {
+                    Some((_, changeset)) => tx_graph_changeset.merge(changeset.into()),
+                    None => return Ok((false, (tx_graph_changeset, reserved_spks_changeset))),
```
nickfarrow, line 286: should reveal_to_target `None` map to error?

frostsnapp/lib/wallet_receive.dart
evanlinjin: As discussed, we can disable dismissing in all other manners except for hitting "Done". This can be done by wrapping the `Dialog` widget with `PopScope`.

frostsnap_coordinator/src/bitcoin/wallet_reserved_spks.rs
nickfarrow: Extremely minor disregard

frostsnap_coordinator/src/bitcoin/wallet.rs
evanlinjin: I guess it can. Also at line `293`. However, I tend to avoid errors in general because we are not `try-catch`-ing them on the dart side. So if it ever errors, the UI state will not progress and the user will be confused.

nickfarrow: This particular scenario would involve something going seriously wrong, I think we'd rather freeze than silently continue

evanlinjin:
I don't think automatically shifting focus to the "Activity" section is a good idea.

https://chatgpt.com/share/6806f72d-05c0-800d-a2ef-66c4f831d104

I suggest we use a more passive approach, like a notification-style visual cue instead.

If we really want to shift focus, it should only happen when the user is already on the "Verify" section and it's the first incoming transaction. Even then, I would prefer not shifting focus automatically.

In any case, I think it would be best to handle this in a separate PR.

nickfarrow:
Wrt to ai points:
 - I don't see it as disorienting since it's in the same workflow and previous panels are still there.
 - Agree it could feel like a loss of control.
 - Don't see this being an interruption (receiving implies already copied, too late to verify)

Agree a notification would also be good here, but less magical. Some other wallets completely switch the screen when money is received (e.g. WOS). May be worth giving @LLFourn a try to see what he thinks.

nickfarrow:
Dropped activity focus commit

nickfarrow (dismissed):

frostsnap_coordinator/src/bitcoin/address_metadata.rs
```diff
@@ -0,0 +1,152 @@
...
+        .to_string()
+    }
+}
+
+impl AddressMetadata {
+    fn apply_mutation(&mut self, mutation: &Mutation) -> bool {
+        match mutation {
+            Mutation::Reveal {
+                key_and_path,
+                label,
```
LLFourn, line 39: I don't think that label should be part of the `Reveal` mutation. Assigning a label is a different mutation since you can do it after an address has been revealed. I will push a commit to fix this.

LLFourn (approved):

PR #242, opened 2025-04-15
[comms] Devices send Reset signal when they reset

LLFourn:
Weird bugs in the Linux kernel driver for ACM CDC devices was being triggered by our aggressive disconnect and reconnect for devices that had finished firmware upgrade. It happened so fast that data got corrupted somehow by the CDC ACM control message `2122030000000000`. Now we only disconnect the port once the device has cleanly finished the upgrade and sent our reset signal.

### DIFF
--- a/device/src/esp32_run.rs
+++ b/device/src/esp32_run.rs
@@ -221,6 +221,9 @@ where
                                         .expect("failed to write conch upstream");
                                 }
                                 ReceiveSerial::Reset => {
+                                    upstream_connection.send_to_coordinator([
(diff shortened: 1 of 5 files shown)


LLFourn:
ACK 98ad4ef. I just got rid of an unused method @nickfarrow introduced in the last commit. Please approve and merge @nickfarrow.

nickfarrow:
Which unused method? Hard to see because of rebase + changes in same forcepush

nickfarrow (approved):

PR #245, opened 2025-04-22
Keygen workflow delayed key commit

evanlinjin:


LLFourn:
This is about the keygen workflow right?

frostsnap_core/src/message/keygen.rs
```diff
@@ -0,0 +1,93 @@
+use super::*;
+
+/// Hello
```
LLFourn, line 3: Hi @evanlinjin!

frostsnap_core/src/coordinator.rs
```diff
@@ -495,53 +496,73 @@ impl FrostCoordinator {
...
-                let mut outgoing = vec![];
-                match self.pending_keygens.get_mut(&keygen_id) {
-                    Some(KeyGenState::WaitingForAcks {
-                        agg_input,
-                        device_to_share_index,
-                        acks,
-                        ..
-                    }) => {
-                        let session_hash = SessionHash::from_agg_input(agg_input);
+                let wrong_state_err = Error::coordinator_invalid_message(
```
LLFourn, line 499: :pick: Better return the same error from different branches so it's easier to debug

LLFourn (approved):

LLFourn:
I rebased and fixed a few things up. Looks great. Minor gripe is that I think this was the time to move things into modules for `keygen` on the coordinator and device side like I did for restoration. That's ok we can do it later.

nickfarrow (approved):
Reviewed and tested ☃️

PR #246, opened 2025-05-06
Recover shares from multiple connected devices

nickfarrow:
Allows for restoration from multiple devices plugged in at once


nickfarrow:
Added in a3b8097ce2086de10bcd35f77e254eaaacb20ad3

PR #247, opened 2025-05-06
Handle restoration cases

nickfarrow:
This needs a bit more work as it doesn't yet handle the case where the same backup is entered on multiple devices during restoration, it only handles duplicate shares on a *plugged in device*


nickfarrow:
Currently I think that we should disallow multiple devices to hold the same share, at least until we address representations of redundant devices, and their behavior in signing and restoration.

It is important to give strong protections against user mistakes at this stage. If we make small preventions against duplicate shares during restoration, it keeps things simple for now and we can later allow it via a targeted PR.

PR #248, opened 2025-05-06
Verify backups are entered on the correct device

nickfarrow:
Draft as it builds upon https://github.com/frostsnap/frostsnap/pull/236


LLFourn:
I wonder if we want to enable a user to check a backup without having to have the full access structure loaded in. E.g. someone who has been entrusted with a frostsnap device wants to check their backup is correct.

If we want to do this then I think the better solution is to have the device include whether the loaded share image matches its already held share image. It could just be a `bool` on `EnteredPhysicalBackup`. 

Another approach would be to ask for `HeldShares` and see if the entered backup matches any of the held shares. This is more elegant but requires an extra message.

We still need the functionality offered by this feature which is to check a specific backup of course. I think it's orthogonal. For doing checking a specific share on a known access structure I think this is the right approach.

LLFourn:
Fixed by #460

PR #249, opened 2025-05-13
[appimage] update ubuntu archive

LLFourn:
App images were not working.

1. Use linuxdeploy to create app images for the sake of simplicity
2. hash pin the version of linuxdeploy using newly invented "just fetch"
3. hash pin the riscv gcc compiler as well
4. stop caching the riscv gcc since it downloads fast


musdom:
tested ACK

nickfarrow (approved):
App image created through CI works for me, however build-appimage doesn't really work locally on arch, need to add:
```
    # Disable stripping on systems with .relr.dyn ELF sections (e.g arch)
    # to avoid errors when linuxdeploy's strip tool encounters newer ELF features.
    # See https://github.com/linuxdeploy/linuxdeploy/issues/311
    if readelf -S /usr/lib/libc.so.6 2>/dev/null | grep -q "\.relr\.dyn"; then
        echo "Libraries with .relr.dyn detected - disabling stripping" >&2
        export NO_STRIP=1
    fi
```

And even then, after seemingly successful build the produced AppImage has gtk errors (I think it wants https://github.com/linuxdeploy/linuxdeploy-plugin-gtk).

I think this is fine, I don't think i'll be needing to build this locally.

PR #250, opened 2025-05-14
Keygen UI overhaul

evanlinjin:


LLFourn:
Rebased and added some improvements.
Note I unfortunately brought he new dart formatting chanages into the new keygen workflow accidentally. This is the last PR in the group so it won't affect any other so I vote to leave this mistake in there.

frostsnapp/lib/wallet_create.dart
```diff
@@ -0,0 +1,1196 @@
...
+      ),
+    );
+  }
+
+  @override
+  void dispose() {
+    _nameController.dispose();
+    _deviceListSub.cancel();
+    for (final device in _deviceList.devices) {
+      coord.sendCancel(id: device.id);
```
LLFourn, line 321: We shouldn't be using `sendCancel` anymore. We can fix this up in another PR.

device/src/esp32_run.rs
```diff
@@ -151,6 +151,7 @@ where
                 upstream_connection.set_state(UpstreamConnectionState::PowerOn, &mut ui);
                 next_write_magic_bytes_downstream = Instant::from_ticks(0);
                 upgrade = None;
+                ui.set_device_name(name.clone());
```
nickfarrow, line 154: Curious as to why this was needed?

frostsnapp/lib/device_setup.dart
```diff
@@ -40,15 +41,23 @@ class DeviceSetup extends StatelessWidget {
   }
 }
 
+enum DeviceNameMode {
+  /// The name field renames the device and prompts the user for confirmation.
+  rename,
+
+  /// The name field stages the device name and persists it after keygen finalizes.
```
nickfarrow, line 48: Is this true (it may be)?
Noticed that we send `DeviceSendBody::SetName` in FinalizeKeyGen

nickfarrow: Possibly related:

After creating a wallet, if I immediately go to the devices list page, one of the devices is missing a name. But if i focus out of the app and back in, it becomes named.

device/src/esp32_run.rs
LLFourn: The ui might not have the fixed name of the device when a soft reset is triggered.

frostsnapp/lib/device_setup.dart
LLFourn: Good catch we probably are forgetting to update the device list when we finish keygen.

nickfarrow:
Threshold slider is great, and as is the UI of device list and security check. 

Made some minor text changes in 769b0f012597fbb9c06d68e43534eadfb7e01d35, possible that some may be controversial. My linter went a bit crazy.. diff is readable but let me know if you want me to isolate just the text changes.

Quite incredible how few changes were needed to the rust code for such a drastic improvement!

frostsnapp/lib/device_setup.dart
LLFourn: > After creating a wallet, if I immediately go to the devices list page, one of the devices is missing a name. But if i focus out of the app and back in, it becomes named.

I don't think you need to focus out of the app for this to fix itself. Just wait a few seconds. This is because nonces are being gathered at this moment so it takes a little while for the devices to get their turn to set their name.

LLFourn:
@nickfarrow I made the keygen fullscreen dialog not have a list of devices that you need to confirm. This makes the dialog have fixed size content like the others and therefore does not need to be special cased. @evanlinjin agreed that we don't need detailed progress on the full screen dialog since it's more of a distraction since you're meant to be focused on the devices.

nickfarrow (approved):

PR #255, opened 2025-05-31
Add Logo and Minor Rewording (Korea Demos)

nickfarrow:


frostsnapp/lib/keygen.dart
```diff
@@ -295,7 +295,7 @@ class _ThresholdPageState extends State<ThresholdPage> {
           crossAxisAlignment: CrossAxisAlignment.center,
           children: [
             Text(
-              "How many devices will be needed to sign under this key?",
+              "How many devices will be needed to sign to spend or make changes to this wallet?",
```
LLFourn:
```suggestion
              "How many devices will be needed spend from this wallet?",
```

There is no way to make changes to a wallet yet so I suggest this (also remove `to sign`).

LLFourn (approved):

PR #257, opened 2025-06-12
[app+coord] Use nusb to handle USB serial on android

LLFourn:
because it's far cleaner to just let android open the USB device and pass it down as a file descriptor. We could potentially do this on all platforms.


nickfarrow (approved):
Tested and works great!

Rebase re-added futures-timer dependency, possible that there's a way to achieve the same using existing dependencies.

nickfarrow:
Replaced future timeout with tokio

PR #262, opened 2025-06-23
Upgrade BDK, Use new sans-io chain source

evanlinjin:


frostsnap_coordinator/src/bitcoin/chain_sync.rs
```diff
@@ -373,14 +475,82 @@ pub struct ChainStatus {
...
+
 #[derive(Clone, Copy)]
 pub enum ChainStatusState {
     Connected,
     Disconnected,
     Connecting,
 }
 
-fn connect(url: &str) -> Result<Conn> {
+async fn check_conn<R, W>(rh: R, mut wh: W, genesis_hash: BlockHash) -> Result<()>
```
evanlinjin, line 493: I forgot to make a comment on this connection check method!

nickfarrow:
I don't know if this is related to this PR, but I managed to get into a state where unconfirmed transactions are not appearing in the transaction list. I made a transaction from sparrow, double spent RBFd it back to myself, and made a bunch of subsequent transactions. I tried disconnected in reconnecting the electrum server connection with the button in the top right but they're still not loading. It could just be a signet thing? Need to do a more rigorous test.

nickfarrow (approved):

PR #265, opened 2025-06-25
Backup Electrum Server and Ping-Pong logic

evanlinjin:
# To Do

- [x] It seems that the `mempool.space` servers do not support the `server.features` method. We need to use an alternative endpoint to get the network.
- [x] Test in the real world. Maybe we can disconnect our server in the office and see if the app reconnects to the backup server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,9 +201,9 @@ dependencies = [
 
 [[package]]
 name = "bdk_electrum_streaming"
-version = "0.5.0"
+version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eb76a331c33cf14b4a47378964c114d193343cbb7e3840f7683f787336e5f87d"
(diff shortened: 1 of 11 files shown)


evanlinjin:
The strange thing is that hitting the Broadcast button should not hang, even if the `broadcast` method in Rust does not return.

The dart future already has a timeout (as we can see here):

https://github.com/frostsnap/frostsnap/blob/eb45a3556a41de42c285d88d0e63e3e738aa3d21/frostsnapp/lib/wallet_tx_details.dart#L611-L614

The only explanation I can think of is that Rust methods block the UI thread. Then rust methods exposed as "async" in dart are not really async...

nickfarrow:
You guys mentioned something about the OS determining whether a connection has dropped, I was expecting an instant reconnection after a failed broadcast (or any other failed request) but realise that may require more encompassing connection management.

We could make the ping-pongs more frequent but I don't think that's quite the right approach for failed calls. What do you think?

PR #271, opened 2025-07-11
Nonce Generation UI Protocol

nickfarrow:
There are two (similar) nonce-replenish UIs: one streamlined in the wallet creation workflow, and one generic that is called after wallet recovery. 

The determination on whether we need actually need to replenish nonces is somewhat unique since it is independent of having a key. For example you can reach the nonce-replenish screen and then back-out and revisit it. I have decided to always show it for consistency.

Should we also conditionally have this this generic widget display on the wallet home-screen if nonces get out of sync? Or alternatively, run it entirely in the background?

I stopped working on the device UI to display that it is generating nonces. The code for `BusyTask` was not being called due to new `WaitingFor` states, and after a minor rabbit hole trying to reimplement I decided to leave this out-of-scope for this PR.

### DIFF
--- a/frostsnapp/lib/restoration.dart
+++ b/frostsnapp/lib/restoration.dart
@@ -8,14 +8,17 @@ import 'package:frostsnap/global.dart';
 import 'package:frostsnap/id_ext.dart';
 import 'package:frostsnap/maybe_fullscreen_dialog.dart';
 import 'package:frostsnap/secure_key_provider.dart';
+import 'package:frostsnap/nonce_replenish.dart';
 import 'package:frostsnap/settings.dart';
 import 'package:frostsnap/snackbar.dart';
 import 'package:frostsnap/src/rust/api.dart';
 import 'package:frostsnap/src/rust/api/bitcoin.dart';
 import 'package:frostsnap/src/rust/api/device_list.dart';
 import 'package:frostsnap/src/rust/api/recovery.dart';
+import 'package:frostsnap/stream_ext.dart';
 import 'package:frostsnap/theme.dart';
 import 'package:frostsnap/wallet_add.dart';
+import 'package:frostsnap/wallet_create.dart';
 
 class WalletRecoveryPage extends StatelessWidget {
   final RestoringKey restoringKey;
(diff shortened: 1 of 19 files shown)


frostsnapp/lib/restoration.dart
```diff
@@ -453,12 +432,21 @@ class _WalletRecoveryFlowState extends State<WalletRecoveryFlow> {
...
+
   @override
   void initState() {
     super.initState();
 
     final initialStep = widget.initialStep;
-    if (initialStep != null) currentStep = initialStep;
+    if (initialStep != null) {
+      // Map string to enum for backwards compatibility
+      currentStep = _mapStringToStep(initialStep);
```
nickfarrow: backwards compatibility with what?

frostsnapp/lib/restoration.dart
```diff
@@ -736,13 +810,40 @@ class _WalletRecoveryFlowState extends State<WalletRecoveryFlow> {
...
+
   void goBackOrClose(BuildContext) {
     if (!tryPopPrevState(context)) Navigator.pop(context);
   }
 }
 
 enum MethodChoiceKind { startRecovery, continueRecovery, addToWallet }
 
+// Recovery flow step states
+enum RecoveryFlowStep {
```
nickfarrow, line 835: Good idea

frostsnapp/lib/restoration.dart
nickfarrow: `typedef NonceReplenishWidget = NonceReplenishDialog;`?

frostsnapp/lib/stream_ext.dart
```diff
@@ -12,12 +12,23 @@ extension StreamToBehaviorSubjectExtension<T> on Stream<T> {
...
-      onDone: () => subject.close(),
+      onDone: () {
+        subject.close();
+        // No need to cancel subscription here as stream is done
+      },
     );
 
+    // CRITICAL FIX: Cancel subscription when subject is closed manually
+    // This prevents memory leaks when the subject is disposed before stream completes
+    subject.onCancel = () {
```
nickfarrow, line 25: Was this a real problem? Maybe plugging in and unplugging devices over and over would take up lots of memory? My understanding is that this stream is short-lived and would be freed by simply closing the dialog.

I'm probably over considering due to "CRITICAL".

nickfarrow:
Tested on keygen and recovery, works nicely!

Minor bug: I can go back from the Naming Devices page and get stuck.

Minor nits: massive commit messages filled with unnecessary detail, this could be a helpful reference:
https://cbea.ms/git-commit/. Many commits should be amends/squashed & forcepushed if forgot to run linter.

nickfarrow:
ACK -- minor commit cleaning

PR #272, opened 2025-07-16
Device Factory Initialization Process

nickfarrow:
WIP

- I need to test on another device as im currently getting a panic due to existing efuses
- Need to incorporate the screen test
- Need to streamline as a CLI with metadata (case colour)
- Need to build genuine cert partition atop of the work Adam is doing with secureboot


device/partitions.csv
musdom, line 1: Disregard previous deleted comment. I will rebase Secure Boot on this PR with partition table and efuse changes.

nickfarrow:
### Bricked devices with disabled flashing
In the past, when testing DS signer, I was ~pausing the device with an empty `loop {}`.

https://github.com/frostsnap/frostsnap/commit/a66ca683328ab4c6de928c0f0738fe6b5fadf541#diff-3d81bc63b39a52a82b977dac0f539e45c5264167f292e2c2732b5c30f16bbee8R164

After fixing the DS signer I then restructured the code and the program progressing further, calling:
```
let hmac_keys = EfuseHmacKeys::init_with_keys(
  efuse,
  hal_hmac,
  read_protect,
  share_encryption_key,
  factory_entropy,
)
```

Shortly after this progression I noticed my device was bricked, unable to update the firmware. In an attempt to resolve this I implemented a **single call to set key purposes**. So the bricking-bug was here **before** this change (fbe37e907b7f2b77fc5539a5666790fc8598f1a9). And similarly, before inclusion of the DS efuse key.

It's possible (/likely) there's an unresolved bug in `write_key_purposes`, or something else related to key slots.

nickfarrow:
Turns out I was wrong about BLOCK0 - you can program each bit 
> Limitations
In BLOCK0, each bit can be programmed separately. However, we recommend to minimize programming
cycles and program all the bits of a parameter in one programming action. In addition, after all parameters
controlled by a certain bit of EFUSE_WR_DIS are programmed, that bit should be immediately programmed.
The programming of parameters controlled by a certain bit of EFUSE_WR_DIS, and the programming of the bit
itself can even be completed at the same time. Repeated programming of already programmed bits is strictly
forbidden, otherwise, programming errors will occur.
BLOCK1 cannot be programmed by users as it has been programmed at manufacturing.
BLOCK2 ~ 10 can only be programmed once. Repeated programming is not allowed.

https://www.espressif.com/sites/default/files/documentation/esp32-c3_technical_reference_manual_en.pdf#efuse

Looking at `write_key_purposes` and the order more closely.

device/src/efuse.rs
```diff
@@ -22,69 +21,88 @@ impl EfuseController {
...
-            3 => buff[12] = kp << 4,
-            4 => buff[13] = kp,
-            5 => buff[13] = kp << 4,
-            _ => return Err(EfuseError::EfuseError),
-        }
-
-        // We bundle every config in Block 0 to minimize write operations
-        // Todo: Write key purpose and rw flags for multiple keys
-        // Disable write to key block
-        let mut write_disable: u32 = 0x01 << (WR_DIS_KEY_OFFSET + key_number);
```
nickfarrow, line 77: Edit: I was completely wrong. This is correct, see Table 4.3-3, bits from `WR_DIS_KEY_OFFSET = 23` are key data protection bits.

nickfarrow:
Turns out the devices weren't actually bricked and it had nothing to do with EFUSE, we accidentally always established the upstream UART which prevented flashing over JTAG. It is possible to recover these devices by repeatedly reconnecting devices while attempting to flash in a while loop. Fixed in e11ea2185ac6e683c74bcdc34b6654264fe1baa4

device/.cargo/config.toml
```diff
@@ -3,7 +3,7 @@ target = "riscv32imc-unknown-none-elf"
 rustflags = ["-C", "link-arg=-Tlinkall.x"]
 
 [target.riscv32imc-unknown-none-elf]
-runner = "espflash flash --baud 921600 --partition-table partitions.csv"
+runner = "espflash flash --baud 921600 --partition-table partitions.csv --bootloader bootloader-dev.bin"
```
nickfarrow, line 6: should we also add this to `just flash`?
Remind me, this `bootloader-dev` doesnt have secure boot keys right?

device/src/secure_boot.rs
```diff
@@ -278,43 +279,38 @@ pub fn is_secure_boot_enabled() -> bool {
...
-        // fail to pass signature verification upon upgrading; and in that incredibly improbably
-        // situation: we as the vendor would be the first notice first and not release it.
-        match app_partition.read_sector(i) {
+pub fn find_signature_sector(partition: &EspFlashPartition) -> Option<(u32, [u8; SECTOR_SIZE])> {
+    let mut signature_block = [0x00; SECTOR_SIZE];
+    // Note: with 1280K firmware, 320 sectors, each has a 1 in 2^(8*4) chance of a false
+    // positive with these magic bytes. So 320/2^(8*4) ~ 1 in 13.4m firmware updates will
+    // fail to pass signature verification upon upgrading; and in that incredibly improbably
+    // situation: we as the vendor would be the first notice first and not release it.
+    for i in 0..partition.n_sectors() {
```
LLFourn: instead of this comment just read it backwards.

PR #273, opened 2025-07-17
Secure Boot

musdom:
Enables Secure Boot v2 on Frostsnaps without disabling boot mode. Production Frontiers will be locked down. Backwards compatible with dev Alpha devices.

In addition to checks listed in https://docs.espressif.com/projects/esp-idf/en/v5.1.6/esp32c3/security/secure-boot-v2.html#verify-signature-block, the app firmware updater will verify new firmware before switching the active partition.

To test SB in this PR
1. Install esptool https://docs.espressif.com/projects/esptool/en/latest/esp32/installation.html so you can use espsecure.py
2. Get secure_boot_signing_key.pem from me. Otherwise you have to build and sign your own bootloader.
3. `just flash-secure-new` if enabling SB on new devices, which will flash bootloader.bin and partitions.bin
4. `just flash-secure` to sign and flash new firmware
5. `just monitor` and Ctrl+R to see expected secure boot status in the bootloader logs:

```
I (281) esp_image: Verifying image signature...
I (282) secure_boot_v2: Verifying with RSA-PSS...
I (285) secure_boot_v2: Signature verified successfully!
I (287) boot: Loaded app from partition at offset 0x160000
I (293) secure_boot_v2: enabling secure boot v2...
I (298) secure_boot_v2: secure boot v2 is already enabled, continuing..
```

Dev devices must `just flash` to be on the new partition table and bootloader-dev.bin.
 
Todo: 
1. Rewrite espsecure.py in Rust for frostsnap-factory
2. Make a test firmware signature block with correct RSA public key hash but wrong signature.
3. Make a production bootloader for Frontier devices


device/partitions.csv
```diff
@@ -1,11 +1,7 @@
...
-# Notes:
-# - The partition table is at 0x8000..0x9000 which I think is why things start after that.
-# - Before that is the bootloader starting at 0x0
+# ESP-IDF Partition Table
+# Name, Type, SubType, Offset, Size, Flags
+esp_secure_cert,data,nvs,0xe000,16K,
+otadata,data,ota,0x12000,8K,
+ota_0,app,ota_0,0x20000,1280K,
+ota_1,app,ota_1,0x160000,1280K,
+nvs,data,nvs,0x2a0000,1400K,
```
nickfarrow, line 7: ACK

Checked over this and it looks great.
The only gap between partitions is 48K after `otadata` and the start of `ota_0`, is this reserved for something? Insubstantial regardless, ~1.1% of our available space.
All the other partitions fit neatly alongside eachother, and all are sector aligned. Everything fits within the 4MB space.

Edit: Ah i see in the TRM:
> Up to 8 MB instruction bus address space can be mapped into the external flash. The mapped address
space is organized as individual 64-KB blocks

So our ota_0, ota_1, nvs are aligned to the 64kb blocks.

device/src/bin/v2.rs
```diff
@@ -736,5 +726,7 @@ fn panic(info: &core::panic::PanicInfo) -> ! {
     let mut display = init_display!(peripherals: peripherals, delay: &mut delay);
     graphics::error_print(&mut display, panic_buf.as_str());
     bl.set_high();
+
+    // switch OTA partition to ota_0/last good one
```
nickfarrow, line 730: TODO: interesting idea to consider, we don't want to allow a malicious coordinator to downgrade and exploit old firmware but maybe there's a safe way to do this with some protection?

musdom: https://docs.espressif.com/projects/esp-idf/en/v5.1.6/esp32c3/api-reference/system/ota.html#anti-rollback

device/src/efuse.rs
```diff
@@ -204,8 +204,14 @@ impl<'a> EfuseHmacKeys<'a> {
         read_protect: bool,
         rng: &mut impl RngCore,
     ) -> Result<Self, EfuseError> {
-        let share_encryption_key_id = hmac::KeyId::Key0;
-        let fixed_entropy_key_id = hmac::KeyId::Key1;
+        // backwards compatibility for Alpha devices
+        let mut share_encryption_key_id = hmac::KeyId::Key0;
+        let mut fixed_entropy_key_id = hmac::KeyId::Key1;
+
+        if Efuse::read_field_le::<u8>(hal_efuse::KEY_PURPOSE_0) == KeyPurpose::SecureBootDigest0 as u8 {
```
nickfarrow, line 211: Really nice, so this means we can keep using our existing dev-boards as they are now with already burnt efuses.

device/src/partitions.rs
```diff
@@ -22,13 +22,12 @@ impl<'a> Partitions<'a> {
...
                 ota_0: EspFlashPartition::new(flash, 0, 0, "ota_0"),
                 ota_1: EspFlashPartition::new(flash, 0, 0, "ota_1"),
-                factory: EspFlashPartition::new(flash, 0, 0, "factory"),
             },
             nvs: EspFlashPartition::new(flash, 0, 0, "nvs"),
         }
     }
     pub fn load(flash: &'a RefCell<FlashStorage>) -> Self {
-        let table = esp_partition_table::PartitionTable::new(0x8000, 10 * 32);
+        let table = esp_partition_table::PartitionTable::new(0xd000, 10 * 32);
```
nickfarrow, line 30: 👍 bootloader size increased

device/src/secure_boot.rs
```diff
@@ -0,0 +1,398 @@
...
+    let mut signature_block = vec![0x00; SECTOR_SIZE];
+    let mut signature_found = false;
+    let mut signature_sector_index = 0u32;
+
+    for i in 0..app_partition.n_sectors() {
+        // Read the sector using FlashPartition's read_sector method
+        match app_partition.read_sector(i) {
+            Ok(sector_data) => {
+                // Check for signature block magic bytes: 0xE7, 0x02, 0x00, 0x00
+                if sector_data[0..4] == [0xE7, 0x02, 0x00, 0x00] {
```
nickfarrow, line 302: Noting that with 1280K firmware, that's 320 sectors, each with a `1 in 2^(8*4)` chance of a false positive with these magic bytes. So `320/2^(8*4)` ~ 1 in 13.4m firmware updates will fail to pass signature verification upon upgrading (and in that improbably situation: we'd notice first and not release). We good!

If this number were higher then I would have suggested that we search for magic bytes backwards through the firmware, such that we pass through mostly `0xff`s.

device/src/secure_boot.rs
```diff
@@ -0,0 +1,398 @@
...
+
+    let mut modulus = [0u32; 96];
+    let mut r_value = [0u32; 96];
+    let mut signature = [0u32; 96];
+    let mut exponent = [0u32; 96];
+
+    modulus.copy_from_slice(&modulus_u32);
+    r_value.copy_from_slice(&r_u32);
+    signature.copy_from_slice(&signature_u32);
+    exponent[0] = exponent_u32[0]; // Exponent is typically small, just copy first word
```
nickfarrow, line 106: Typically or always for this variant of RSA PSS? Is there a reference for this? Any downside to of copying it all? Probably doesn't matter given it works

justfile
```diff
@@ -6,20 +6,38 @@ ordinary_crates := "-p frostsnap_core -p frostsnap_coordinator -p frostsnap_comm
...
+    just flash-secure
+    just monitor
+
+monitor +ARGS="":
+    espflash monitor --no-stub
 
 erase-device +ARGS="nvs":
     cd device && espflash erase-parts --partition-table partitions.csv {{ARGS}}
 
 build-device BOARD=default_board +ARGS="":
```
nickfarrow, line 28: maybe we should have

`just build-device-secure`
which tacks on the espsecure command and is used in
`just flash-secure`

And keep
`just build-device`
for dev boards
`just flash`

what do you think?

musdom:
> Phenomenal @musdom, added some notes and minor restructuring for clarity during review in [5ad4d8a](https://github.com/frostsnap/frostsnap/commit/5ad4d8a943681533bbbe34d5449b79c3455d6d4f)
> 
> Now to test it!
> 
> Update: for testing you need a device that has efuses free!!
> 
> If using an old device you also need to make sure to erase prior to flashing secure boot otherwise you'll have to rescue with a firmware hack since you can no longer `just erase-device`
> 
> oops
> I think there may be a ~bug in the updating process when trying to flash unsigned firmware, before the update even begins the device just blackscreens and soft-bricks. Investigating
> 
> note: we also need to integrate bundling the signed firmware into the app.

I have been "erasing" dev Secure Boot devices using my cli tool that uses espflash write-bin to write an empty binary to relevant partitions.

musdom:
> Phenomenal @musdom, added some notes and minor restructuring for clarity during review in [5ad4d8a](https://github.com/frostsnap/frostsnap/commit/5ad4d8a943681533bbbe34d5449b79c3455d6d4f)
> 
> Now to test it!
> 
> Update: for testing you need a device that has efuses free!!
> 
> If using an old device you also need to make sure to erase prior to flashing secure boot otherwise you'll have to rescue with a firmware hack since you can no longer `just erase-device`
> 
> oops
> I think there may be a ~bug in the updating process when trying to flash unsigned firmware, before the update even begins the device just blackscreens and soft-bricks. Investigating
> 
> note: we also need to integrate bundling the signed firmware into the app.

Perhaps we just put the key I gave you into github secrets and have github actions use it to sign firmware images using espsecure.py, for now. Just for dev devices with SB enabled.

nickfarrow:
Could you help me understand why we can't `just erase-device` once secure boot is enabled? Is it because we don't use the stub?

Since we can't easily erase otadata (without your CLI), we can get in a situation where we update a secure-boot device via the app (switching to `ota_1`), then when we next `just flash-secure` it only flashes `ota_0` and the device doesn't run the new firmware. 

We could make `flash-secure` write to both OTA partitions to avoid this issue. Wdyt?

musdom:
> Could you help me understand why we can't `just erase-device` once secure boot is enabled? Is it because we don't use the stub?

Yes

> Since we can't easily erase otadata (without your CLI), we can get in a situation where we update a secure-boot device via the app (switching to `ota_1`), then when we next `just flash-secure` it only flashes `ota_0` and the device doesn't run the new firmware.

Yeah I should make `just flash-secure` erase the otadata partition 

> We could make `flash-secure` write to both OTA partitions to avoid this issue. Wdyt?

I think we should make devices fall back to a previous good firmware (unless stopped by anti-rollback) if the new firmware breaks, by switching partitions in the panic handler.

device/src/secure_boot.rs
```diff
@@ -0,0 +1,382 @@
...
+) -> Result<(), SecureBootError<'a>>
+where
+    S: ReadNorFlash + embedded_storage::nor_flash::NorFlash,
+{
+    let mut signature_block = vec![0x00; SECTOR_SIZE];
+    let mut signature_found = false;
+    let mut signature_sector_index = 0u32;
+
+    // Find signature block in app partition using magic bytes (magic byte, secure boot v2, padding)
+    for i in 0..app_partition.n_sectors() {
```
LLFourn, line 294:
```suggestion
    for i in 0..app_partition.n_sectors().rev() {
```

iterate backwards so we don't have to have this crazy comment below?

nickfarrow:
Included in https://github.com/frostsnap/frostsnap/pull/272

PR #274, opened 2025-07-24
UI Refresh: Device List + Fullscreen Action Dialog

evanlinjin:
* Changed styling of FAD
* Use FAD in more places
* Redo device list


evanlinjin:
Alex gave me some valuable feedback on this PR. I plan to change the following:

- [ ] The heavy blur behind the Fullscreen Action Dialog is confusing when the display is compact. Let's make the background fully opague when the screen is compact.
- [ ] Device page: Rename `Name` to `Device Name`.
- [ ] Remove `Firmware Digest` on the Fullscreen Action Dialog for upgrading a device.
- [ ] Add a dialog `Device successfully upgraded` after a device has been upgraded.
- [ ] Backup Checklist: Wrap the contents in a `SafeArea` so OS elements do not cover our content.
- [ ] Backup Checklist: Disable buttons for disconnected devices.
- [ ] Fullscreen Action Dialog: Action buttons as a row looks strange on compact display. Either: Change to single button + cancel on top left corner. Or move buttons + hint to actionable row attached to bottom edge for compact displays.
- [ ] Use yellow for (caution) things - I.e. backup checklist (instead of red).
- [ ] Frostsnap full logo is too much in the app. Change to just logo.

nickfarrow:
### Device List

"Connected Devices" and "Unnamed" is great.

For empty devices I was thinking that we don't necessarily need a Create Wallet button, doesn't really hurt - though maybe it's simplest to have minimal entry points (is that what they're called?) into workflows? Probably overthinking it.

Any thoughts on Empty Device, Blank Device, Fresh Device? I'm thinking Fresh Device may sound least scary and most inviting? But maybe that's confusing as to what Fresh means. No strong preference.

Not sure if warning/alert symbol is appropriate for a device requiring an upgrade, maybe an arrow or some other symbol? There is a `system_update_alt` icon, or `mobile_arrow_down`.

### Upgrade Screen
I like the phrasing of "Upgrade" over "Update"!

Upgrade screen looks great, we could put "New" at the front of "Firmware Digest". In the future it will be great to put a green tick and "Verified Signature from Frostsnap". No strong preference whether to remove this or not.

Bug: Confirm on device starts at 1/2!

### Display Backup on Device
Small nits on this one:
* Alignment of text is a little strange in the box, actually the text is in a nice position (central) but the box extents almost to the edges horizontally. (on desktop).
* Maybe rename share -> key? I don't think we use "share" anywhere else.
* Not 100% sure about the tick next to Backup Recorded - it looks like it is telling me that the backup has already been recorded when actually it is a button for me to prompt that it has been recorded.

### Erase Device
Looks fantastic. Maybe could do with slightly more information?

"This will wipe the key from the device and will render it blank."
"This action can not be reverted, and the only way to restore this key is through loading of a backup."

Device disconnected sad face screen remains after wiping a device (this has been finicky for a while and I think I may have hacked in a dialog close previously).

### Restoration
Prior to entering the backup on a new device, we still use the old "Confirm name" device action dialog but I think you just haven't got around to this one yet.

nickfarrow:
Wallet home width:
image

evanlinjin:
@nickfarrow Thanks for taking the time to review this.

Some UI aspects (such as sizing, spacing, and alignment) are hard to resolve over PR comments since they often lack a clear objective basis. It would really help if we had a set of Frostsnap design guidelines/principles. In the meantime, I’ve tried to follow the Material design language as much as possible. Some parts were tricky to align with the framework, though the newer Material Expressive update has helped a bit.

Given how subjective these discussions can get, how about we set aside some time next week to talk through the “why” behind our design choices and work toward a more objective foundation for future decisions?

nickfarrow:
Sounds like a good idea, but next week may prove tricky timewise. This discussion would be quite open-ended since I and others don't have much idea of exactly which decision points we need principles for (though you probably do!). Maybe we should start collecting specific ones on an issue?

The styling you've done is fantastic, and if any of my suggestions don't align with Material Design then we can drop them! For this PR, most of my feedback is on information clarity and user-flow rather than style.

* Navigation drawer - this one is subjective and Lloyd pointed out it will look better with more items, so with that second opinion let's run with how it is.
* Wallet home width - No specific Material guidelines against narrow centered columns, but full-width layout looks really nice on desktop IMO and is the only remaining thing I feel strongly about.
* Back arrow from device settings - From a quick look it seems material guidelines don't specifically cover dialog navigation buttons, so this is more of a UX decision. I think `Icons.arrow_back` makes more sense for navigating within a flow since it does not dismiss the entire dialog stack.

frostsnapp/lib/wallet.dart
```diff
@@ -76,11 +76,12 @@ class WalletHome extends StatelessWidget {
...
-              color: theme.colorScheme.outline,
-              height: 48,
-              alignment: Alignment.center,
-              image: AssetImage('assets/icons/frostsnap-icon-trimmed.png'),
+            child: Center(
+              child: Image(
+                color: theme.colorScheme.primary,
+                height: 75,
+                image: AssetImage('assets/icons/frostsnap-icon-trimmed.png'),
+              ),
```
evanlinjin: I agree the logo looks nicer larger.

I think the `primary` color is too bright for the log and takes the focus away from the buttons. How about we use `theme.colorScheme.secondary` so the logo blends in with the text?

image

frostsnapp/lib/wallet.dart
```diff
@@ -584,24 +585,15 @@ class WalletDrawer extends StatelessWidget {
...
-                      image: AssetImage('assets/frostsnap-logo-trimmed.png'),
-                    ),
-                  ),
-                ],
+              padding: const EdgeInsets.all(25),
+              child: Image(
+                color: theme.colorScheme.primary,
+                image: AssetImage('assets/frostsnap-logo-trimmed.png'),
+                fit: BoxFit
+                    .fitWidth, // This will scale to fit the available width
```
evanlinjin: I disagree with these measurements. However, my previous measurements is also wrong.

To be consistent with the rest of the elements; height should be 56, vertical padding should be 16, horizontal padding should be 28. The logo should be aligned to the left.

I think this will make it conform more to a typical Material Navigation rail/drawer.

* [Navigation rail - Specs](https://m3.material.io/components/navigation-rail/specs)
* [Navigation drawer - Specs](https://m3.material.io/components/navigation-drawer/specs)

This is how it would look on mobile:

signal-2025-08-25-165003_002

frostsnapp/lib/wallet.dart
```diff
@@ -650,7 +642,20 @@ class WalletDrawer extends StatelessWidget {
...
+                                backgroundColor: theme.colorScheme.primary,
+                                child: Text(
+                                  '$n',
+                                  style: TextStyle(
+                                    color: theme.colorScheme.onPrimary,
+                                    fontSize: 12,
+                                    fontWeight: FontWeight.w500,
+                                  ),
+                                ),
+                              );
```
evanlinjin: Navigation drawer badges are not highlighted: https://m3.material.io/components/navigation-drawer/specs

frostsnapp/lib/wallet.dart
nickfarrow: I think the others thought it looked too washed out with this colour, maybe we gauge over message
image

frostsnapp/lib/wallet.dart
nickfarrow: Easy feel free to fixup, thanks!

frostsnapp/lib/wallet.dart
nickfarrow: Is this for the logo? IMO it looks far too small in this screenshot, but happy to merge it and reconsider later

evanlinjin: @nickfarrow It's a navigation drawer though. I don't think it's a place for big logos.

Some Google apps use an appbar-sized logo on the top (64 tall). That's the largest I'm willing to go.

image

frostsnapp/lib/wallet.dart
evanlinjin: Just to kind of explain where I am coming from; Here are some examples of "start pages" where the logo colour is the same as the text on the page. I don't think the logo should take away from the actionable options.





Let me know if this changes your mind, otherwise I'm happy to recede.

evanlinjin: I've went along with the general consensus for now. However, I disagree with this choice.

nickfarrow (approved):

PR #277, opened 2025-07-30
Deterministic Firmware Builds

nickfarrow:
```
just build-deterministic
```

```
[2025-08-17T14:10:37Z INFO ] Image successfully saved!
Build complete - 711776 bytes
a4f60d306e9189961e23795aa4bd51a92ccea578070c3d25e39b8b5d5202f954  /var/tmp/frostsnap/target/riscv32imc-unknown-none-elf/release/firmware.bin
```

### DIFF
--- a/.github/actions/setup-nix/action.yml
+++ b/.github/actions/setup-nix/action.yml
@@ -1,0 +1,9 @@
+name: "Setup Nix with Flakes"
+description: "Installs and configures Nix"
+runs:
+  using: "composite"
+  steps:
+    - name: Install Nix
+      uses: DeterminateSystems/nix-installer-action@main
+    - name: Setup Magic Nix Cache
+      uses: DeterminateSystems/magic-nix-cache-action@main
--- a/.github/workflows/deterministic-build.yml
+++ b/.github/workflows/deterministic-build.yml
@@ -1,0 +1,41 @@
+name: Deterministic Build
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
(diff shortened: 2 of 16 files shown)


nickfarrow:
The remaining challenge has been the digest difference between sha256sum of `firmware.bin` and the `0xff` padded digests we're using inside app & device. We're currently calculating a padded digest so that the device can simply hash the entire partition, including empty 0xff NVS bytes, without needing to know the actual firmware size.

For robustness the firmware size should be determined from the firmware bytes alone, the approach I have taken in this PR. (Other less desirable options being: bundle padded firmware, accept padding as a prerequisite to digests, store firmware size in OtaMetadata).

Following code from [espflash](https://github.com/esp-rs/espflash/blob/main/espflash/src/image_format/idf.rs) we parse the ESP-IDF binary format, consisting of: header, segment headers + data, padding, sha256 digest, and optional signatures.

We validate the firmware header, then iterate through all data segments applying padding rules.
Then we look for signatures and return the total firmware size.

Failure in this code would be a mess so we should scrutinize it strongly. It needs to be tested on signed firmware.

LLFourn:
just noting I was able to reproduce the same hash: 

```
a4f60d306e9189961e23795aa4bd51a92ccea578070c3d25e39b8b5d5202f954  /var/tmp/frostsnap/target/riscv32imc-unknown-none-elf/release/firmware.bin
```

LLFourn (approved):
ACK

- cleaned up the firmware size a little (no traits made it clear it's sort of temp hacky code)
- moved nix off into a corner in CI because it takes so long
- removed signature from size calc.

PR #279, opened 2025-08-14
Wait after changing BAUD before OTA

nickfarrow:
Upgrading devices in a daisy chain from android mysteriously broke in 408870de7267d57097d874ada495d6f25169fd69.

The top-most device fails to read the ready signal from the downstream device after it has changed its BAUD rates, waiting forever it never reads in the firmware from the coordinator as it believes the downstream device is not ready.

Adding a small wait after changing BAUD rates ensures that every device is on the same page before communicating again.

edit: Lloyd acked me tacking on the app lint fixes here

### DIFF
--- a/device/src/ota.rs
+++ b/device/src/ota.rs
@@ -6,6 +6,8 @@ use crate::{
 use alloc::boxed::Box;
 use bincode::config::{Fixint, LittleEndian};
 use esp_hal::sha::Sha;
+use esp_hal::time::Duration;
+use esp_hal::timer;
 use frostsnap_comms::{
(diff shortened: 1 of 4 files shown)


device/src/ota.rs
```diff
@@ -318,6 +321,11 @@ impl FirmwareUpgradeMode<'_> {
             downstream_io.change_baud(OTA_UPDATE_BAUD);
         }
 
+        let now = timer.now();
```
LLFourn:
```suggestion
        let start = timer.now();
```

LLFourn (approved):
Thanks!

PR #285, opened 2025-08-15
Upgrade to bincode v2 and latest secp256kfun

LLFourn:
- Upgrade bincode from rc3 to 2.0.1
- Update to latest secp256kfun/schnorr_fun master
- Migrate from PartyIndex to ShareIndex
- Use upstream ShareImage instead of local definition
- Fix bincode::Decode bounds to use () context where functions control decoding

### DIFF
--- a/frostsnap_comms/src/lib.rs
+++ b/frostsnap_comms/src/lib.rs
@@ -44,7 +44,11 @@ pub const BINCODE_CONFIG: bincode::config::Configuration<
 > = bincode::config::standard().with_limit::<MAX_MESSAGE_ALLOC_SIZE>();
 
 #[derive(Encode, Decode, Debug, Clone)]
-#[bincode(bounds = "D: Direction")]
+#[bincode(
+    encode_bounds = "D: Direction",
+    decode_bounds = "D: Direction, <D as Direction>::RecvType: bincode::Decode<__Context>",
+    borrow_decode_bounds = "D: Direction, <D as Direction>::RecvType:  bincode::BorrowDecode<'__de, __Context>"
+)]
 pub enum ReceiveSerial<D: Direction> {
     MagicBytes(MagicBytes<D>),
     Message(D::RecvType),
@@ -287,7 +291,7 @@ pub trait HasMagicBytes {
 }
 
 pub trait Direction: HasMagicBytes {
-    type RecvType: bincode::Decode + bincode::Encode + for<'a> bincode::BorrowDecode<'a> + Gist;
+    type RecvType: bincode::Encode + Gist;
     type Opposite: Direction;
 }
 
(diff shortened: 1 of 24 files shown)


frostsnap_coordinator/src/persist.rs
```diff
@@ -236,7 +236,10 @@ impl<T: bincode::Encode> ToSql for BincodeWrapper<T> {
     }
 }
 
-impl<T: bincode::Decode> FromSql for BincodeWrapper<T> {
+impl<T> FromSql for BincodeWrapper<T>
+where
+    for<'a> T: bincode::Decode<()>,
```
LLFourn: useless `for<'a>`.

frostsnap_core/src/symmetric_encryption.rs
```diff
@@ -12,7 +12,10 @@ pub struct Ciphertext<const N: usize, T> {
     tag: [u8; 16],
 }
 
-impl<const N: usize, T: bincode::Encode + bincode::Decode> Ciphertext<N, T> {
+impl<const N: usize, T: bincode::Encode> Ciphertext<N, T>
+where
+    for<'a> T: bincode::Decode<()>,
```
LLFourn: useless `for<'a>`.

frostsnap_core/src/tweak.rs
```diff
@@ -429,6 +429,10 @@ mod test {
             3,
             5,
             5,
+            schnorr_fun::frost::Fingerprint {
+                tag: "frostsnap-v0",
+                bit_length: 10,
```
nickfarrow, line 434: 🔥

frostsnap_comms/src/lib.rs
```diff
@@ -287,7 +291,7 @@ pub trait HasMagicBytes {
 }
 
 pub trait Direction: HasMagicBytes {
-    type RecvType: bincode::Decode + bincode::Encode + for<'a> bincode::BorrowDecode<'a> + Gist;
+    type RecvType: bincode::Encode + Gist;
```
nickfarrow, line 294: I think RecvType should always be decodable? I had a go at adding it back and it requires this new context, but we could go with the default context since we don't need any extra data to deserialize for our purposes. 
```
type RecvType: bincode::Encode + bincode::Decode<()> + for<'a> bincode::BorrowDecode<'a, ()> + Gist;
```
but this is probably inconsequential, it already compiles

nickfarrow (approved):

frostsnap_comms/src/lib.rs
LLFourn: You're right that's probably what I should have done. What I did is a little bit more general but probably completely useless in the end. I don't think we should be afraid of doing `bincode::Decode<()>` as a bound.

PR #290, opened 2025-08-20
Use HMAC for Device Secrets

nickfarrow:
This PR enables the devices to use their HMAC EFUSE keys for various secrets.

By using these EFUSE keys, an attacker wanting to extract the secrets would not only have to dump the flash from the device but they would additionally have to extract the relevant key from EFUSE; significantly more difficult (requiring chip decapping/probing or sophisticated sidechannel attacks). 

### Device Keypair
The device keypair is currently used for DeviceId and encryption of keygen shares sent to  other parties.

Since this keypair is created on startup, this call could prove weak to side-channel attacks. Because device keypair compromise only affects device identity and keygen share encryption (rather than signing keys themselves), we choose to use the less critical `fixed-entropy` HMAC for this purpose.

### Nonce secrets

An attacker learning the nonce secret is essentially equivalent to them learning the secret share itself: with a signature and nonce secret they can solve for the secret share. We use the `share-encryption` HMAC to derive nonce-secrets to secure these crucial secrets.

### DIFF
--- a/device/src/efuse.rs
+++ b/device/src/efuse.rs
@@ -1,4 +1,3 @@
-// use core::num::NonZeroU8;
 use esp_hal::efuse::{self as hal_efuse, Efuse};
 use esp_hal::hmac::{self, Hmac};
 use esp_hal::peripherals::EFUSE;
@@ -251,7 +250,7 @@ impl<'a> EfuseHmacKey<'a> {
 
     pub fn hash(
         &mut self,
-        domain_separator: &'static str,
+        domain_separator: &str,
         input: &[u8],
     ) -> Result<[u8; 32], esp_hal::hmac::Error> {
         let mut hmac = self.hmac.borrow_mut();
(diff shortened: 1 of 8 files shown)


nickfarrow:
This does not yet implement protection against a coordinator repeatedly requesting nonces in attempt to carry out timing attacks,

nickfarrow:
TODO: review use of `self.nonce_stream_id` and initial stream seed (somewhat unrelated to this PR)

device/src/efuse.rs
```diff
@@ -308,4 +307,18 @@ impl frostsnap_core::device::DeviceSymmetricKeyGen for EfuseHmacKey<'_> {
 
         frostsnap_core::SymmetricKey(output)
     }
+
+    fn derive_nonce_seed(
+        &mut self,
+        nonce_stream_id: frostsnap_core::nonce_stream::NonceStreamId,
+        index: u32,
```
LLFourn, line 314: Adding these two params here was the main change I made.

LLFourn (approved):
I added 636ea2b992bea4eef8a31cefc509d9934ee0baa3 to add the nonce_stream_id and the index to the hmac derivation -- this doesn't do anything formally for security but it structures the derivation a bit more by giving the trait method a bit more of an identity.

PR #292, opened 2025-08-22
Modify `Persist` to handle migrations

evanlinjin:
* Add `Persist::migrate`.
* Rename `Persist::init` to `load` and `Persist::InitParams` to `LoadParams`.


evanlinjin (commented):
ACK 8db4227526809ba783fe4359fed0ae935d622082

PR #293, opened 2025-08-26
Restoration Workflow: styling and wording changes

evanlinjin:
Based on #274


aphanley:
Tested manually with devices. Looks good!

A few minor UI feedback points:
key index tooltip - "the key number", other tooltips are capitalised "Valid Key" "Remove Share"
change "Remove Share" tooltip to "Remove Key"
Clicking restore with existing device goes to heading "Restore with existing key", could consider making this
	=> "Restore from an existing device", or
	=> "Restore from a device with an existing key"
Location of the (X) to exit tabs is positioned in the top left, but i think it's more normal for it to be top right
Not so sure about the tertiary colour being purple, but I think it's fine for now we can revisit colours and theming later
Colour of "cancel" button shouldn't be red  in (it's the safe option) in the Recover from physical backup > Device not blank flow
When prompted to erase a device because it is not blank, we should probably have it be consistent with how it is for create a wallet. In the create a wallet flow, if a device isn't blank, then we have to exit out of the flow and erase the device from the connected devices tab. Evan already added more friction, so at least we have the warning about erasing the device now

One more significant thing which i will create an issue for:
It's possible to start restoring a wallet and add some keys and then start a new wallet restoration flows and add other keys, which then conflicts with the original one. It would be better to only allow a wallet to be restored once at a time.

evanlinjin:
> When prompted to erase a device because it is not blank, we should probably have it be consistent with how it is for create a wallet. In the create a wallet flow, if a device isn't blank, then we have to exit out of the flow and erase the device from the connected devices tab. Evan already added more friction, so at least we have the warning about erasing the device now

> One more significant thing which i will create an issue for:
It's possible to start restoring a wallet and add some keys and then start a new wallet restoration flows and add other keys, which then conflicts with the original one. It would be better to only allow a wallet to be restored once at a time.

@aphanley let's address this separately. Everything else I've fixed.

aphanley (approved):

PR #294, opened 2025-08-28
[frost_backup] 🎉

LLFourn:
Adds a backup for FROST shares based with a simple share index and mnemonic


frost_backup/README.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+Shares written to/read from `./shares/` directory.
+
+[secp256k1]: https://en.bitcoin.it/wiki/Secp256k1
+[Shamir secret shares]: https://en.wikipedia.org/wiki/Shamir%27s_secret_sharing
+[BIP39]: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki
+[FROST]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-frost/
+[SHA256]: https://en.wikipedia.org/wiki/SHA-2
+[fingerprint-grinding]: https://github.com/BlockstreamResearch/bip-frost-dkg#fingerprinting-key-generation
```
nickfarrow, line 123: Wrong link?

frost_backup/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Attack scenario:**
+1. Multiple devices hold shares of the same secret key with associated public key
+2. After restoring from a backup, a device knows its share but not the public key
+3. Malicious coordinator provides incorrect public key for the share
+4. Device derives and displays incorrect addresses
+5. User sends funds to attacker-controlled addresses
+
+The attack requires specific timing (user receiving funds immediately after restoration) and is detectable through any spending operation. If successful, the attacker who controls the false public key can steal any funds sent to the incorrect addresses.
+
```
nickfarrow, line 84: Our fingerprint grinding does not protect against a malicious coordinator public key attack because:

When the coordinator receives our share-image, it could produce a malicious set of other share images that when interpolated with our share-image produces the correct fingerprint while still interpolating to an incorrect public key?

To avoid this you'd need to send over the *other* share images & public key first, committing to it, but this leaks the public keys to any fresh device. Furthermore when recovering multiple devices the coordinator wouldn't have enough share images at this point in time to commit to it.

frost_backup/README.md
```diff
@@ -0,0 +1,123 @@
...
+|----------|--------|----------|------------------------|
+| 0-255    | 256    | 1-24     | Secret share scalar    |
+| 256-263  | 8      | 24       | [Polynomial checksum](#polynomial-checksum)    |
+| 264-274  | 11     | 25       | [Words checksum](#words-checksum)         |
+
+## Checksums
+
+The checksums are [SHA256] hashes of the input data.
+
+- `share_index` (32 bytes): although share indices are usually small numbers they are encoded in the hash as 32-byte secp256k1 scalar (big endian encoded 256-bit integer).
```
nickfarrow, line 35: This doesnt match the code:

```
    fn compute_words_checksum(
        index: u32,
        scalar: &Scalar<impl Secrecy, impl ZeroChoice>,
        poly_checksum: u16,
    ) -> u16 {
        let hash = Sha256::new()
            .add(index.to_be_bytes())
            .add(scalar)
            .add(poly_checksum.to_be_bytes())
            .finalize();
        // Take first 2 bytes as u16 and shift right to keep only the top 9 bits
...
```

Is it preferable to encode as Scalars for some minor futurproofing towards supporting longer share-indicies (avoiding complexity)?

frost_backup/README.md
LLFourn: lol yup

frost_backup/README.md
LLFourn: Nice find yeah that's meant to take a Scalar. Looks like I fixed it for the polynomial one but not that one.

frost_backup/src/share_backup.rs
```diff
@@ -0,0 +1,330 @@
...
+        index: ShareIndex,
+        scalar: &Scalar<impl Secrecy, impl ZeroChoice>,
+        shared_key: &SharedKey<Normal, Z>,
+    ) -> u16 {
+        let hash = Sha256::new()
+            .add(index.to_bytes())
+            .add(scalar)
+            .add(shared_key.point_polynomial())
+            .finalize();
+        // Take first 2 bytes as u16 and shift right to keep only the top 10 bits
```
nickfarrow, line 226: outdated comment: 8 bits

frost_backup/src/share_backup.rs
```diff
@@ -0,0 +1,330 @@
...
+        index: u32,
+        scalar: &Scalar<impl Secrecy, impl ZeroChoice>,
+        poly_checksum: u16,
+    ) -> u16 {
+        let hash = Sha256::new()
+            .add(index.to_be_bytes())
+            .add(scalar)
+            .add(poly_checksum.to_be_bytes())
+            .finalize();
+        // Take first 2 bytes as u16 and shift right to keep only the top 9 bits
```
nickfarrow, line 240: Same here

frost_backup/src/recovery.rs
```diff
@@ -0,0 +1,326 @@
...
+                // The poly must have at least 2 coefficients for us to discover
+                // shares that are compatible with each other.
+                if shared_key.point_polynomial().len() < 2 {
+                    continue;
+                }
+
+                // Check if it matches the fingerprint
+                if shared_key.check_fingerprint::<sha2::Sha256>(&fingerprint) {
+                    found = Some(shared_key);
+                    break;
```
nickfarrow, line 201: Could break 'outer here for performance?

We already capture all compatible shares below (which makes sense since we could have interpolated a correct subset, while have extras that lay on the same polynomial.

PR #296, opened 2025-08-30
CST  IRQ

LLFourn:
Handle touch events in the interrupt so we don't miss them.


musdom:
Tested on 3 devices, no touch issues

PR #305, opened 2025-09-04
[device] Arrange peripheral initialisation nicely

LLFourn:


.github/actions/set-up-app-build/action.yml
```diff
@@ -17,5 +17,5 @@ runs:
     - name: Retrieve device firmware
       uses: actions/download-artifact@v4
       with:
-        name: "frostsnap-${{ github.sha }}.bin"
+        name: "frostsnap-frontier-${{ github.sha }}.bin"
```
nickfarrow: Do we want to put `dev` or `frontier` inside github artifacts?

I suspect these artifacts will generally be used by us for testing on-the-move, and not downloaded by users (requires github logins) -- but im not 100% sure. Could do both but maybe that's too much mess.


I think it definitely makes sense to have the deterministic build be of `frontier` though

nickfarrow: Addressed in 507953267c52ff7b9c40b83d10eafad126b903e1

PR #310, opened 2025-09-06
CST IRQ LFG Edition 🫰🫰🫰

LLFourn:
[cst] Use hardware IRQ so we don't miss touches 🫰

+ vendor the old crate here.
+ Make the Peripherals into a static ref
+ unify touch calibration

### DIFF
--- a/cst816s/src/lib.rs
+++ b/cst816s/src/lib.rs
@@ -1,0 +1,314 @@
+#![no_std]
+use core::fmt::Debug;
+
+use embedded_hal as hal;
+use hal::delay::DelayNs;
+
+/// Errors in this crate
+#[derive(Debug)]
+pub enum Error<CommE, PinE> {
+    Comm(CommE),
+    Pin(PinE),
+    GenericError,
+}
+
+pub struct CST816S<I2C, PINT, RST> {
+    i2c: I2C,
+    pin_int: PINT,
+    pin_rst: RST,
+    blob_buf: [u8; BLOB_BUF_LEN],
+}
+
+impl<I2C, PINT, RST> CST816S<I2C, PINT, RST> {
+    /// Default queue capacity for touch events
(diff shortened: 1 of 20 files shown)


device/src/bin/widget_dev.rs
```diff
@@ -81,61 +78,14 @@ fn main() -> ! {
                         now.duration_since_epoch().to_millis(),
                     );
 
-                    // Check for touch events
-                    if let Some(touch_event) = capsense.read_one_touch_event(true) {
-                        let touch_point = adjust_touch_point(
```
nickfarrow, line 86: touch calibration on simulator 😅

cst816s/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct TouchEvent {
+    pub x: i32,
+    pub y: i32,
+    /// the gesture that this touch is part of
+    pub gesture: TouchGesture,
+    /// 0 down, 1 lift, 2 contact
+    pub action: u8,
```
nickfarrow, line 58: Many of these could be enums, but let's address in a follow-up PR

justfile
```diff
@@ -94,7 +95,7 @@ fix: fix-dart fix-rust
 
 fix-rust:
     cargo clippy --fix --allow-dirty --allow-staged {{ordinary_crates}} --all-features --tests --bins
-    ( cd device && cargo clippy --fix --allow-dirty --allow-staged --all-features --bins; )
+    cargo clippy --fix --allow-dirty --allow-staged {{device_crates}} --all-features
```
nickfarrow, line 98: Confirmed still lints our bins

nickfarrow:
Reviewed code, good to merge if @aphanley has tested

PR #312, opened 2025-09-06
UART IRQ LFG edition 🎨

LLFourn:
[device] Use UART IRQs to fill buffers

So we don't freeze everything with conches. Upgrading from conch devices
should work still.

They fire when the FIFO buffer gets full enough (96 bytes/128).


aphanley:
keygen_check screen on devices going back to standby before the user has actioned 'yes' or 'no' in the 'final check' in the app

PR #314, opened 2025-09-06
[coord] Remove address_metadata

LLFourn:
stale file


evanlinjin (approved):
Such an important change

PR #315, opened 2025-09-07
Reserve some nvs space

nickfarrow:


device/src/esp32_run.rs
```diff
@@ -75,6 +75,9 @@ pub fn run<'a>(resources: &'a mut Resources<'a>) -> ! {
         NonceAbSlot::load_slots(nvs.split_off_front(n_nonce_sectors))
     };
 
+    // Keep some space reserved for other potential uses in the future, 8 AB slots
+    let _reserved = nvs.split_off_end(8 * 2);
```
LLFourn: split it off the front before we spit off the nonces IMO.

Also see there's a `.max(16)` -- I don't think we want a MAX? (Oh wait that means at least 16).

nickfarrow: After the share partition and before the nonces?

LLFourn: yes I think that would be good.

nickfarrow: force pushed pls see change

PR #318, opened 2025-09-07
Modularize enums

LLFourn:
Because it's a breaking change to do this later.

It also includes an API improvement with `BeginKeygen` (the first commit).

### DIFF
--- a/frostsnap_coordinator/src/keygen.rs
+++ b/frostsnap_coordinator/src/keygen.rs
@@ -3,9 +3,7 @@ use std::collections::BTreeSet;
 use crate::{Completion, Sink, UiProtocol};
 use frostsnap_comms::CoordinatorSendMessage;
 use frostsnap_core::{
-    coordinator::{CoordinatorToUserKeyGenMessage, CoordinatorToUserMessage, FrostCoordinator},
-    message::keygen,
-    schnorr_fun::fun::KeyPair,
+    coordinator::{BeginKeygen, CoordinatorToUserKeyGenMessage, CoordinatorToUserMessage, FrostCoordinator},
     AccessStructureRef, DeviceId, KeygenId, SessionHash,
 };
 use tracing::{event, Level};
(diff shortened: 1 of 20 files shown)


frostsnap_core/src/coordinator.rs
```diff
@@ -1543,58 +1545,28 @@ impl std::error::Error for StartSignError {}
...
-    },
-    CloseSignSession {
-        session_id: SignSessionId,
-        finished: Option<Vec<EncodedSignature>>,
-    },
-    ForgetFinishedSignSession {
-        session_id: SignSessionId,
-    },
+    #[delegate_kind]
+    Keygen(keys::KeyMutation),
```
LLFourn, line 1549: This variant should be renamed `Key` or something like that. It's a non-breaking change to rename it so it can stay like this for this PR.

frostsnap_core/tests/malicious.rs
```diff
@@ -24,21 +25,23 @@ fn keygen_maliciously_replace_public_poly() {
     let mut run = Run::generate(1, &mut test_rng);
     let device_set = run.device_set();
     let mut shadow_device = run.devices.values().next().unwrap().clone();
-    let coordinator_keygen_keypair = FrostCoordinator::short_lived_keygen_keypair(&mut test_rng);
+
+    use rand_chacha::rand_core::{RngCore, SeedableRng};
+    let mut seed = [0u8; 32];
+    test_rng.fill_bytes(&mut seed);
+    let mut coordinator_rng = ChaCha20Rng::from_seed(seed);
```
LLFourn, line 32: This is necessary because otherwise proptest can end in a situation where a device and the coordinator have the same public key which causes weird stuff to happen (nothing insecure).

frostsnap_coordinator/src/keygen.rs
```diff
@@ -22,14 +20,13 @@ impl KeyGen {
...
-        coordinator_keypair: KeyPair,
-        begin_keygen: keygen::Begin,
+        begin_keygen: BeginKeygen,
         rng: &mut impl rand_core::RngCore,
     ) -> Self {
         let mut self_ = Self {
             sink: Box::new(keygen_sink),
             state: KeyGenState {
-                devices: begin_keygen.devices.clone(),
+                devices: begin_keygen.devices().into_iter().collect(),
```
nickfarrow: 😱

nickfarrow (approved):
Had to hack around device ordering for now

PR #320, opened 2025-09-07
[device] fix erase

LLFourn:


device/src/esp32_run.rs
```diff
@@ -582,6 +583,7 @@ pub fn run<'a>(resources: &'a mut Resources<'a>) -> ! {
                                 .expect("state changed while confirming keygen"),
                         );
                         ui.clear_busy_task();
+                        ui.set_workflow(default_workflow!(name, signer));
```
nickfarrow: We got rid of this in https://github.com/frostsnap/frostsnap/pull/319

LLFourn: well spotted.

nickfarrow:
Tested

PR #325, opened 2025-09-08
Fix touch screen freeze when pressed during startup

aphanley:
Problem: Touch screens become unresponsive when touched during device boot. Testing revealed that holding touch during startup causes a race condition where the interrupt fires before the handler is registered.

Changes:
- Move touch controller setup into register() function for atomic initialization
- Register interrupt handler BEFORE enabling interrupt pin (critical fix)
- Remove interrupt enable from new_esp32() constructor
- Optimize reset recovery delay from 400ms to 50ms
- Add delay parameter to register() function

The key fix is ensuring the interrupt handler is registered before the interrupt pin is enabled. This prevents interrupts from firing without a handler, which was causing the touch system to enter an undefined state.

Testing confirmed that just changing the handler registration order fixes the issue - the interrupt must have a handler ready before it can fire.

### DIFF
--- a/cst816s/src/lib.rs
+++ b/cst816s/src/lib.rs
@@ -83,7 +83,8 @@ where
         self.pin_rst.set_low().map_err(Error::Pin)?;
         delay_source.delay_us(20_000);
         self.pin_rst.set_high().map_err(Error::Pin)?;
-        delay_source.delay_us(400_000);
+        // ❗ This used to 400_000 but this seems to work in practice.
+        delay_source.delay_us(50_000);
 
         Ok(())
     }
@@ -272,7 +273,7 @@ pub mod interrupt {
         io: &mut impl InterruptConfigurable,
     ) -> TouchReceiver {
         use esp_hal::gpio::Event;
-        
+
         unsafe {
             // Split the queue into producer and consumer
             let event_queue = &raw mut EVENT_QUEUE;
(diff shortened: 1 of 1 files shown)


cst816s/src/lib.rs
```diff
@@ -266,13 +265,24 @@ pub mod interrupt {
...
+        delay_source: &mut impl hal::delay::DelayNs,
     ) -> TouchReceiver {
+        use esp_hal::gpio::Event;
+        
+        // Perform setup (moved from setup_esp32)
+        // Reset the chip
+        instance.pin_rst.set_low();
+        delay_source.delay_us(20_000);
+        instance.pin_rst.set_high();
+        delay_source.delay_us(50_000);
```
LLFourn: I think we should leave this in the old set up function? I'd actually like to know more about this code. Since we run this at the start these delays may not be necessary (the chip has already just been reset).

device/src/peripherals.rs
```diff
@@ -256,11 +256,10 @@ impl<'a> DevicePeripherals<'a> {
         .with_sda(&mut peripherals.GPIO4)
         .with_scl(&mut peripherals.GPIO5);
 
-        let mut capsense = CST816S::new_esp32(i2c, &mut peripherals.GPIO2, &mut peripherals.GPIO3);
-        capsense.setup(&mut delay).unwrap();
+        let capsense = CST816S::new_esp32(i2c, &mut peripherals.GPIO2, &mut peripherals.GPIO3);
 
-        // Register the capsense instance with the interrupt handler
-        let touch_receiver = frostsnap_cst816s::interrupt::register(capsense, &mut io);
+        // Register the capsense instance with the interrupt handler (includes setup)
```
LLFourn: Why are we changing where the setup is done? What if we change the code not to use the interrupt handler in some circumstances? Then it won't run. The register function should only be for interrupt handling.

LLFourn (approved):

PR #330, opened 2025-09-16
[ci,cst] Fix device linting

LLFourn:
The cst crate was not being linted properly.


evanlinjin (approved):
ACK d25548893b6e56fff3bae259149360218efe46e8

PR #332, opened 2025-09-16
[device] Fix weird type inference thing in ota.rs

LLFourn:
This doesn't lead to problems on master but does elsewhere!


evanlinjin (approved):
ACK https://github.com/frostsnap/frostsnap/pull/332/files

PR #335, opened 2025-09-16
Fix blank screen issue on restart by adding delay to usb

aphanley:
When restarting after a firmware update or after plugging in device to android coordinator, often the screen on the first device was blank.

When it is the first time the device has been plugged in the android jtag notice/accept caused a delay which resulted in screens operating fine

Restoring the delay for the touchscreen initalisation from 50ms to 400ms and also adding a 1000ms delay before opening the usb port from the android coordinator eliminated the blank screen issue.


LLFourn:
I fixed a comment and snuck in the fix for the panic handler also into this one.

LLFourn (approved):

PR #336, opened 2025-09-16
WIP: One unsigned tx at one time

evanlinjin:


frostsnapp/rust/src/api/bitcoin.rs
```diff
@@ -166,6 +167,45 @@ pub struct Transaction {
...
+            txid: txid.to_string(),
+            confirmation_time: None,
+            last_seen: None,
+            prevouts,
+            is_mine,
+        }
+    }
+
+    pub(crate) fn fill_signatures(&mut self, signatures: &[EncodedSignature]) {
+        for (txin, signature) in self.inner.input.iter_mut().zip(signatures) {
```
nickfarrow, line 198: I think we should be iterating locally owned inputs here, otherwise we could be overwriting external signatures (a bug we have hit before iirc).

Could be worth considering whether it is appropriate to ~assert on the lengths (return None)  like you do in `attach_signatures_to_psbt`. 

Noticing (again) that transaction inputs have an implicit ordering that's preserved throughout the signing process, maintained by the signing coordinator when collecting signatures and handled in order through all signature collection points. I wonder if it is worth making less fragile at some point in the future.

frostsnapp/rust/src/api/signing.rs
```diff
@@ -19,6 +19,27 @@ use frostsnap_core::{
 };
 use std::collections::{HashMap, HashSet};
 
+/// An outgoing Bitcoin transaction that is not canonical.
+///
+/// May be signed or unsigned, but is guaranteed to have a signing session associated with it.
+#[derive(Debug, Clone)]
+#[frb]
+pub struct UncanonicalTx {
```
nickfarrow: This makes total sense, but I'm starting to wonder whether we're creating a zoo of transaction types:

  - TransactionTemplate - Template for building transactions (frostsnap_core)
  - UnsignedTx - Template wrapper for signing API (frostsnapp)  
  - Transaction - Runtime wrapper with metadata (frostsnapp)
  - UncanonicalTx - Transaction + signing session state (frostsnapp)
  - SignedTx - Signed transaction pair (frostsnapp)

Each serves a distinct purpose, but wondering if there are opportunities to streamline patterns/interfaces across them as the codebase evolves. I couldn't come up with any but you would have a better grasp on this than me.

frostsnapp/lib/contexts.dart
```diff
@@ -89,6 +89,7 @@ class SuperWalletContext extends InheritedWidget {
     // Get or create tx stream
     var stream = _txStreams[keyId];
     if (stream == null) {
+      print('Made tx stream for ${masterAppkey.field0.inner}');
```
nickfarrow: Leftover debug print

frostsnapp/rust/src/api/bitcoin.rs
evanlinjin: I agree with making this less fragile.

Note that the current codebase assumes that the signer owns all the inputs (check [`UnsignedTx::complete()`](https://github.com/frostsnap/frostsnap/blob/affa6fc338b52dd39ca1cdca4bd2a6d5180069c1/frostsnapp/rust/src/api/signing.rs#L198-L214)). I was only following this assumption.

I rather tackle this design flaw in a separate PR. Can we make a ticket and go from there?

frostsnapp/rust/src/api/signing.rs
evanlinjin: There definitely is.

Note that there is `SignedTxDetails` to add to the confusion (I don't think it's used though).

* `Transaction` is created from `TransactionTemplate` + signatures (optional).
* `UnsignedTx` is created from `TransactionTemplate`.
* `SignedTx` is create from `TransactionTemplate` + signatures.
* `UncanonicalTx` is `TransactionTemplate` + signing session.

I want the distinction between uncanonical vs canonical since we are handling them differently.

We can probably get rid of `SignedTx` and rename `UnsignedTx` to `TransactionTemplate`.

Can we tackle this in a separate PR?

LLFourn (changes requested):
The thing about this PR that I don't like is that it makes us see non-canonical transactions in the transaction list. Even if the transaction was double spent it will be there forever. e.g. if I do a fee bump with sparrow that tx will still live in the list forever and be at the top of the list.

This PR should be simpler and just disallow adding more than one unbroadcasted tx. To get the correct behavior it needs to be thought through carefully and BDK is the place to figure that out. Not here.

LLFourn:
@evanlinjin thanks for pointing this out. The solution is to have a single stream for unbroadcasted and broadcasted txs. I've started doing that in another branch a little while ago but it's a lot of work. I've made it re-trigger on either a signingSessionSignal or txStream like you had it before.

PR #337, opened 2025-09-17
[mac] Mac support 🍎

LLFourn:


LLFourn:
I tested the dmg artifact and it works on an m1 mac. You have to jump through some hoops because it's not signed with a developer key.

nickfarrow (approved):
I haven't been able to test but changes look great to me!

oleonardolima:
tACK aa87f26304559bdc223c6f761524eb8d720a44f2
I did a quick test with my ARM mac, and was able to update the firmware and create a single device wallet.

PR #339, opened 2025-09-18
[comms] Fixed length DeviceName and key name

LLFourn:
This enforces string length limits for device names at the type level. It doesn't do key names because that's part of frostsnap_core but soft limits are in place.

Longer device names will stilld decode they'll just be truncated.

### DIFF
--- a/frostsnap_comms/src/fixed_string.rs
+++ b/frostsnap_comms/src/fixed_string.rs
@@ -1,0 +1,229 @@
+use alloc::string::{String, ToString};
+use core::fmt;
+use core::ops::Deref;
+use core::str::FromStr;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct FixedString<const N: usize> {
+    inner: String,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StringTooLong {
+    pub max_len: usize,
+    pub actual_len: usize,
+}
+
+impl fmt::Display for StringTooLong {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "String too long: max length is {} but got {}",
+            self.max_len, self.actual_len
+        )
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for StringTooLong {}
+
+impl<const N: usize> FixedString<N> {
+    pub fn new(s: String) -> Result<Self, StringTooLong> {
+        let char_count = s.chars().count();
+        if char_count > N {
+            Err(StringTooLong {
+                max_len: N,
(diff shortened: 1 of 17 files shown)


frostsnap_comms/src/fixed_string.rs
```diff
@@ -0,0 +1,198 @@
...
+                actual_len: s.len(),
+            })
+        } else {
+            Ok(Self { inner: s })
+        }
+    }
+
+    pub fn truncate(mut s: String) -> Self {
+        if s.len() > N {
+            s.truncate(N);
```
nickfarrow: Found an edge case that can cause panics:

* `.len()` and `truncate(N)` are byte based
* Some UTF-8 characters are larger than 1 byte
* Truncate panics if not on a char boundary

So it is possible in this function to panic on UTF-8 characters since the truncation doesn't end up at a character boundary.

E.g.

```
    #[test]
    fn test_utf8_boundary_panic() {
        let s = "ServerRoomCafé".to_string();
        let _result = FixedString::<14>::truncate(s);
    }
```

We could address this with:
```
pub fn truncate(mut s: String) -> Self {
    if s.len() > N {
        let mut end = N;                           // Start at our max length
        while end > 0 && !s.is_char_boundary(end) { // Keep backing up until we find a valid boundary
            end -= 1;
        }
        s.truncate(end);  // Now safe to truncate
    }
    Self { inner: s }
}
```

Perhaps important for european and spanish languages, and future language character sets.

Note: when you try `ServerRoomCafé` within the app, the coordinator silently panics with "String too long: max length is 14 but got 15" but you can continue and eventually the device panics on not having a pending name after keygen.

LLFourn: Good point. I changed:

1. The FixedString to be character based
2. The flutter app so names can only use a restricted character set

frostsnapp/lib/global.dart
```diff
@@ -9,6 +10,10 @@ late Coordinator coord;
 late Api api;
 late HostPortHandler? globalHostPortHandler;
 
+final nameInputFormatter = FilteringTextInputFormatter.allow(
+  RegExp(r"[a-zA-Z0-9_\-']"),
```
nickfarrow: I think we should also allow spaces, i often use them. What do you think?
Alternative could be to automatically substitute inputted spaces with `_` or `-`

nickfarrow:
Ok there is a little complexity when allowing spaces 008f5fe299fac25fd93ef61ce3361272a4d4bb8c: we don't want to allow trailing spaces, but you can't prevent them during the input formatter since typing "My Wallet" will get stuck at "My", so it necessitates having some `trim()`s in places.


Decided against allowing accented characters for now because it's not simple to determine which ones will render for particular device fonts, and we will be changing to our own fonts soon.

nickfarrow (approved):

PR #340, opened 2025-09-18
[core] Save name once backup is saved, don't confirm

LLFourn:


justfile
```diff
@@ -38,7 +38,7 @@ erase-device +ARGS="nvs":
     cd device && espflash erase-parts --partition-table partitions.csv {{ARGS}}
 
 build-device BOARD=default_board +ARGS="":
-    cd device && cargo build --release --bin {{BOARD}} {{ARGS}}
+    cd device && cargo build --release --bin {{BOARD}} ${DEVICE_BUILD_ARGS:-} {{ARGS}}
```
nickfarrow, line 41: How do these `${DEVICE_BUILD_ARGS:-}` differ to just `{{ARGS}}`? Or is it to allow for setting via environment variables?

LLFourn: Yep. Can set it via environment variable.

frostsnapp/lib/restoration.dart
```diff
@@ -1918,26 +2023,109 @@ class _PhysicalBackupFailView extends StatelessWidget with _TitledWidget {
...
+
+class _FirmwareUpgradeViewState extends State<_FirmwareUpgradeView> {
+  late final DeviceActionUpgradeController _controller;
+  bool _isUpgrading = false;
+
+  @override
+  void initState() {
+    super.initState();
+    _controller = DeviceActionUpgradeController();
+    // Don't handle disconnection - firmware upgrade causes device to reset which is normal
```
nickfarrow, line 2081: @musdom found an issue where the phone-connected device shows a black screen on android after a firmware update. Not sure if related.

PR #344, opened 2025-09-24
[core] Fix nonce accounting

LLFourn:
There was a method `after_signing` which was called on both the device and he coordinator and did the wrong thing in both places. Fixing proptests to try and account for nonce streams revealed both these bugs.


frostsnap_core/src/device_nonces.rs
```diff
@@ -188,9 +190,11 @@ pub trait NonceStreamSlot {
             .expect("guaranteed")
             .signature_shares;
 
-        let implied_coord_nonce_state = coord_nonce_state.after_signing(signature_shares.len());
-        let replenishment =
-            self.reconcile_coord_nonce_stream_state(implied_coord_nonce_state, device_hmac);
+        let replenishment = self.reconcile_coord_nonce_stream_state(
+            coord_nonce_state,
```
nickfarrow, line 194: 👍 pass the original state in to reconcile

frostsnap_core/src/device_nonces.rs
```diff
@@ -86,11 +86,12 @@ pub trait NonceStreamSlot {
         &mut self,
         state: CoordNonceStreamState,
         device_hmac: &mut impl DeviceSecretDerivation,
+        nonce_batch_size: u32,
```
nickfarrow, line 89: Aside, might be worth documenting

```
    /// Cases:
    /// 1) Our index is ahead of the coordinator's perspective
    ///     -> Replenish the full batch size starting from our index
    /// 2) The coordinator has less nonces than the batch size
    ///     -> Replenish the full batch size starting from our index
    /// 3) Our index is in sync with the coordinator's perspective
    ///     -> Nothing to replenish
    /// 4) Our index is behind the coordinator's perspective (unreceived requests / cancels)
    ///     -> Resend nonces from our index up to the coordinator's index, they're unused.
```

nickfarrow (approved):

PR #347, opened 2025-09-29
Support firmware digests with or without signature block

nickfarrow:
In our haste to release we hashed the entire firmware+signatures (meaning the digest is not equivalent to the unsigned deterministic build digest).

In hindsight I should've noticed this because we were not doing any signature block decoding on the coordinator side.

This commit prepares devices to function with future coordinators that calculate the firmware digest without the signature block.

I propose that we introduce this commit now, and later follow up with a future update to the coordinator which has it calculate the OTA digest without the signature block. This follow-up commit would come at a time where we have reasonable confidence that the small number of devices in circulation have received the change in this PR.

Neglected devices which do not receive this change, but then do OTA with the future coordinator will panic -- but this is recoverable by using an older version of the app to do OTA first.


device/src/ota.rs
```diff
@@ -423,16 +423,24 @@ impl FirmwareUpgradeMode<'_> {
         } = &self
         {
             let partition = &ota.ota_partitions()[*ota_slot];
-            let (_firmware_size, firmware_and_signature_block_size) =
+            let (firmware_size, firmware_and_signature_block_size) =
                 partition.firmware_size().unwrap();
-            assert_eq!(*size, firmware_and_signature_block_size);
+            if (*size != firmware_size) && (*size != firmware_and_signature_block_size) {
+                panic!("coordinator sent firmware with size that did not match expected")
```
LLFourn: :pick: This panic message is a little weird. This sounds like the coordinator sent firmware that is a different size than what it said but that's not what this panic means.

We don't need this check I think but I guess it's useful for debugging. We check the "size" implicitly by copying over the correct number of bytes.

device/src/ota.rs
```diff
@@ -423,16 +423,24 @@ impl FirmwareUpgradeMode<'_> {
...
-                    expected_digest
+            // New digest approach: without signature block
+            let digest_without_signature = partition.sha256_digest(sha, Some(firmware_size));
+            if digest_without_signature != *expected_digest {
+                // Old approach: digest with signature block (for backwards compatibility)
+                let digest_with_signature =
+                    partition.sha256_digest(sha, Some(firmware_and_signature_block_size));
+                if digest_with_signature != *expected_digest {
+                    panic!(
+                    "upgrade downloaded did not match intended digest. \nGot:\n{}\nExpected:\n{}\n\n{}",
```
LLFourn: if we're going to print out three digests can you annotate what they are.

device/src/ota.rs
nickfarrow: Yeah makes sense. It was getting hit a lot during development, but in its absence the digest mismatch would have hit anyway.

nickfarrow:
Closing in favor of https://github.com/frostsnap/frostsnap/pull/360

PR #348, opened 2025-09-29
BIP21 Receive and Send Flows

nickfarrow:
Addresses the user-reported issue of not being able to decode BIP21 URIs.

Builds upon https://github.com/frostsnap/frostsnap/pull/329 because we removed some redundant address detection in there.

In the _Send Flow_, pasting or scanning a BIP21 URI will decode into the address, with network validation like before.
If an amount parameter is present then we use that to prefill the amount field on the next step.

In the _Receive Flow_, we now prepend `bitcoin:` to address QR codes, but not to copied addresses.


LLFourn (approved):
approved. Added a commit: ab955608b0570b8f2bb00170207db923ab8df414

one note is that this won't yet support all of [BIP321](https://bips.xyz/321).

nickfarrow:
Thanks!!

PR #354, opened 2025-10-02
Recovery threshold optional

LLFourn:


device/src/flash/log.rs
```diff
@@ -12,6 +12,7 @@ pub enum Mutation {
 #[derive(Debug, Clone, bincode::Encode, bincode::Decode)]
 pub enum ShareSlot {
     SecretShare(Box<SaveShareMutation>),
+    LegacySavedBackup(device::restoration::LegacySavedBackup),
```
nickfarrow: So the strategy for making something legacy is to rename the enum variant (keeping its same index) and add the new mutation at the end?

nickfarrow (commented):
Ok everything looks good to me, I understand the testing approach and looked through the changes to recovery not requiring a specified threshold. Going to battle-test now

device/src/flash/log.rs
LLFourn: That's what I did here. It might not be wise. I could consider just naming it SaveBackup0 and SaveBackup1 etc.

nickfarrow: I guess we can always rename both of them if we end up having multiple legacy versions. Approach makes sense to me.

nickfarrow:
1. Noticed that if you add two compatible keys, and then an incompatible one, you can still press "Restore" which causes the incompatible device to panic: "we can't consolidate a share we don't know about". I guess the UI should prevent restoration until you remove the incompatible ones.

2. When we allow differing wallet names, it correctly sets the known name (from device) to the physical backup device, but doesn't override the name in the flutter (this may be more complex, given mutations? Relatively minor and can be fixed by a recovery workflow that guides users to recover from keys before physical backups).
Reproduce by loading physical backup first (with differing wallet name) then restore from another device.

3. Also, @musdom noticed that we probably don't want a dropdown for the recovery threshold, at least not limited to 5.

Perhaps
```
Threshold (optional): <Leave blank if unsure>
```
where any number can be entered.

aphanley:
Trying to recover with multiple devices plugged in seems to cause issues. You have to try again multiple times to get it to detect the devices that are still yet to be recovered

I was able to trigger the null check operator used on a null value issue when recovering with multiple devices plugged in.
It was inconsistent but only ever seemed to occur after nonce generation

aphanley:
On restoring a wallet with a physical device, device remains in recovery mode
Then since the wallet is now restored I am able to try to verify a wallet address, triggering a device panic (since it is still in recovery mode)

nickfarrow:
1. Minor: if you restore two compatible keys (from devices) and then enter an incompatible seedwords for a third device, you can remove a compatible device and get into quite a weird state:
image

2. We lost the ability to simply plug in T existing devices and restore, now you have to connect them one-by-one which is minorly less satisfying. It could be argued that we should discourage bringing all the devices back to a single location, but I think it does suit the situation of an inheritance estate settlement meeting.
 
Was there a motivation in the code design for this change? Previously it was a little weird how it would flicker "not compatible" and then detect a compatible device, but it was somewhat nice (also for testing) to be able to progressively connect devices.
 
Above, @aphanley reported having issues with multiple connected devices (before this prevention): "You have to try again multiple times to get it to detect the devices", was this when connecting multiple at once? or one-by-one? I don't recall experiencing these issues.

5. This state inconsistency is still possible: When we allow differing wallet names, it correctly sets the known name (from device) to the physical backup device, but doesn't override the name in the flutter (this may be more complex, given mutations?  Reproduce by loading physical backup first (with differing wallet name) then add key from existing device.

We could help resolve 3. by guiding users to recover from existing devices before physical backups. I think a guaranteeing fix would be quite involved (overriding wallet name during consolidation?). Perhaps a TODO.

Tested "This key share belongs to existing wallet xxxx". Cool to see this also works for keys that hasn't actually been added to that existing wallet yet (virtual share).

LLFourn:
> 1. Minor: if you restore two compatible keys (from devices) and then enter an incompatible seedwords for a third device, you can remove a compatible device and get into quite a weird state:
> 
> 
> image
> 

The problem was saving the access structure ref in the metadata even when the share is not compatible. In the the future we should save nothing when the share is not compatible.

>     2. We lost the ability to simply plug in T existing devices and restore, now you have to connect them one-by-one which is minorly less satisfying. It could be argued that we should discourage bringing all the devices back to a single location, but I think it does suit the situation of an inheritance estate settlement meeting.
> 
> Was there a motivation in the code design for this change? Previously it was a little weird how it would flicker "not compatible" and then detect a compatible device, but it was somewhat nice (also for testing) to be able to progressively connect devices.
> 
> Above, @aphanley reported having issues with multiple connected devices (before this prevention): "You have to try again multiple times to get it to detect the devices", was this when connecting multiple at once? or one-by-one? I don't recall experiencing these issues.
> 

Despite the interaction being less satisfying, one by one is simpler to implement and has educational value.  I think that all "plug in device" things should force one by one except for keygen obviously. @aphanley problem was was connecting many devices at once. That used to work but I replaced that intricate logic with a simpler version. 

>     5. This state inconsistency is still possible: When we allow differing wallet names, it correctly sets the known name (from device) to the physical backup device, but doesn't override the name in the flutter (this may be more complex, given mutations?  Reproduce by loading physical backup first (with differing wallet name) then add key from existing device.
> 
> 
> We could help resolve 3. by guiding users to recover from existing devices before physical backups. I think a guaranteeing fix would be quite involved (overriding wallet name during consolidation?). Perhaps a TODO.
> 

Yes this PR did not address that. It requires us to detect and prompt. Or we could just allow starting physical backups with no name and then if no device has been plugged in, name it at the end. 

> Tested "This key share belongs to existing wallet xxxx". Cool to see this also works for keys that hasn't actually been added to that existing wallet yet (virtual share).

nickfarrow (approved):

PR #356, opened 2025-10-02
[app,coord] filter out invalid PSBTs

LLFourn:
Where we don't have anything that we can sign. Previously if you added a PSBT that has nothing to sign the app will never start up again.

Also tried to add a generic error dialog that we can iterate on.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/wallet.rs
+++ b/frostsnap_coordinator/src/bitcoin/wallet.rs
@@ -648,13 +648,17 @@ impl CoordSuperWallet {
         &mut self,
         psbt: &bitcoin::Psbt,
         master_appkey: MasterAppkey,
-    ) -> Result<TransactionTemplate> {
+    ) -> Result<TransactionTemplate, PsbtValidationError> {
         let bitcoin_app_xpub = master_appkey.derive_appkey(AppTweakKind::Bitcoin);
         let our_fingerprint = bitcoin_app_xpub.fingerprint();
         let mut template = frostsnap_core::bitcoin_transaction::TransactionTemplate::new();
         let rust_bitcoin_tx = &psbt.unsigned_tx;
         template.set_version(rust_bitcoin_tx.version);
         template.set_lock_time(rust_bitcoin_tx.lock_time);
+
+        let mut already_signed_count = 0;
+        let mut foreign_count = 0;
+        let mut owned_count = 0;
 
         for (i, input) in psbt.inputs.iter().enumerate() {
             let txin = rust_bitcoin_tx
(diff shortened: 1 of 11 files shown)


frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -760,10 +778,82 @@ impl CoordSuperWallet {
...
+                already_signed_count,
+            } => {
+                let input_word = if *total_inputs == 1 {
+                    "input"
+                } else {
+                    "inputs"
+                };
+                write!(
+                    f,
+                    "This PSBT has {} {} but this wallet can not sign any of them",
```
nickfarrow: Nit: Kind of pointless doing this plurality branch since we have "any of them" on the end. 
Could have `("input", "it")` and `("inputs", "of them")` to do it correctly, but having
`This PSBT has 1 input but this wallet can not sign any of them` is ~ok.

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -760,10 +778,82 @@ impl CoordSuperWallet {
...
+                )?;
+
+                if *foreign_count > 0 && *already_signed_count > 0 {
+                    write!(
+                        f,
+                        " ({} not owned by this wallet, {} already signed)",
+                        foreign_count, already_signed_count
+                    )?;
+                } else if *foreign_count > 0 {
+                    write!(f, " ({} not owned by this wallet)", foreign_count)?;
```
nickfarrow: Could be worth stating that frostsnap does not yet support signing of collaborative transactions, but not sure if here is the right place

frostsnap_widgets/src/lib.rs
```diff
@@ -119,8 +119,8 @@ pub const FONT_MED: fonts::u8g2_font_profont22_mf = fonts::u8g2_font_profont22_m
 pub const FONT_SMALL: fonts::u8g2_font_profont17_mf = fonts::u8g2_font_profont17_mf;
 
 pub const HOLD_TO_CONFIRM_TIME_SHORT_MS: u32 = 1000;
-pub const HOLD_TO_CONFIRM_TIME_MS: u32 = 2000;
-pub const HOLD_TO_CONFIRM_TIME_LONG_MS: u32 = 6000;
+pub const HOLD_TO_CONFIRM_TIME_MS: u32 = 1000;
+pub const HOLD_TO_CONFIRM_TIME_LONG_MS: u32 = 1000;
```
nickfarrow: Intentional? Or for testing?

LLFourn: Oops unintentional.

frostsnap_coordinator/src/bitcoin/wallet.rs
nickfarrow: fixed up

frostsnap_coordinator/src/bitcoin/wallet.rs
nickfarrow: Not including for now

PR #357, opened 2025-10-03
[app/android] Sleep a bit more (and less)

LLFourn:
- sleep more in the app
- sleep less on the device

Seems to stop the issue where after an erase the device will fail to start up properly.


musdom (approved):
tested on Android

PR #358, opened 2025-10-06
Screen test widget

musdom:
Turned screen test into another widget. It shows the touch test grid, then draws the hold_to_confirm widget in a smaller canvas.

### DIFF
--- a/frostsnap_widgets/src/screen_test.rs
+++ b/frostsnap_widgets/src/screen_test.rs
@@ -2,7 +2,7 @@ use crate::{
     palette::PALETTE, DynWidget, HoldToConfirm, Instant, SuperDrawTarget, Text, Widget, FONT_MED,
     HOLD_TO_CONFIRM_TIME_SHORT_MS,
 };
-use alloc::format;
+use alloc::{format, vec::Vec};
 use embedded_graphics::{
     mono_font::{iso_8859_1::FONT_10X20, MonoTextStyle},
     pixelcolor::Rgb565,
@@ -12,7 +12,6 @@ use embedded_graphics::{
 };
 use u8g2_fonts::U8g2TextStyle;
 
-// Constant for the "lift up" action.
 const ACTION_LIFT_UP: u8 = 1;
 
 // Target cells with colors for each corner
@@ -22,8 +21,6 @@ const TARGET_CELLS: [(i32, i32, Rgb565); 4] = [
     (0, 6, Rgb565::YELLOW), // Bottom left
     (5, 6, Rgb565::BLUE),   // Bottom right
 ];
-
-use alloc::vec::Vec;
 
 /// Current phase of the screen test
 enum Phase {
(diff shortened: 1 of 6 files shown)


justfile
```diff
@@ -29,7 +29,7 @@ flash-secure-new BOARD=default_board +ARGS="":
 flash-frontier-new BOARD="frontier" +ARGS="":
```
nickfarrow, line 29: So this is for production devices. Should we rename `flash-secure-new` to `flash-dev-new`? Or even obsolete that command altogether?

Do we want/need to flash dev devices with dev secure boot? If no, then maybe we can delete `flash-secure-new` altogether?

justfile
```diff
@@ -29,7 +29,7 @@ flash-secure-new BOARD=default_board +ARGS="":
...
     espflash write-bin --chip esp32c3 --port /dev/ttyACM0 --baud 921600 --no-stub 0x0 device/bootloader-frontier.bin {{ARGS}}
     espflash write-bin --chip esp32c3 --port /dev/ttyACM0 --baud 921600 --no-stub 0xD000 device/partitions.bin {{ARGS}}
-    just flash-secure {{BOARD}}
```
nickfarrow, line 32: Thanks for remembering this!

justfile
nickfarrow: We want to keep it since Adam has some dev-secure boot devices, may be useful for future testing.

justfile
nickfarrow, line 32: Noting that removing this removes the `blank-otadata.bin`.

I wondered if [this scenario](https://github.com/frostsnap/frostsnap/pull/273#issuecomment-3218383227) could become relevant again:

1) Testing on non-secure boot device, do a firmware upgrade --> increments to `ota_1`.
2) Decide to `just flash-frontier-new`, enables firmware signature verification.
3) On boot, bootloader checks `ota_1`, signature not present, tries `ota_0` which would work(?).

So I think it's not a potential issue with this bootloader? But then this begs the question, why was I ever having issues with this when using the development bootloader in the first place?

frostsnap_widgets/src/screen_test.rs
```diff
@@ -0,0 +1,408 @@
...
+        needs_redraw: bool,
+        touch_points: Vec<Point>, // Store all touch points to draw red dots
+    },
+    Menu {
+        failures: i32,
+        hold_to_confirm: HoldToConfirm<Text<U8g2TextStyle<Rgb565>>>,
+        start_again_rect: Rectangle,
+        prev_action: Option<u8>,
+        needs_redraw: bool,
+    },
```
LLFourn, line 40: As a matter of style you'd usually have different widgets for these two pages and then one widget that uses a Switcher<..> to switch between them or has an AnyOf<...> inside it and manually handles the display clearing and switching.

not a big deal for the screen test though :)

PR #360, opened 2025-10-08
[device] Support firmware digests with or without signature block

LLFourn:
Previously, v0.0.1 firmware digest included the signature block and padding, preventing users from verifying reproducible builds against the digest displayed on the device screen.

This commit extracts firmware parsing logic into a shared FirmwareReader trait and implements firmware_only_digest() to hash only the deterministic firmware content (excluding padding and signature blocks).

Key changes:
- Created frostsnap_comms::firmware_reader with FirmwareReader triat to unify the logic to figure out the firmware offsets so both device and coordinator are using the same code (jsut reading sectors in a different way)
- Added PrepareUpgrade2 message variant which sends the raw firmware hash (without signature block)
- Added comprehensive tests with v0.0.1 signed/unsigned firmware binaries (binaries are committed).

The device now displays the deterministic firmware digest, allowing users to verify it matches their locally-built reproducible firmware.


nickfarrow:
[Cleaned up](https://github.com/frostsnap/frostsnap/compare/66159b4ca2cb6359d61e1d0059839086dd02f961..e3d68cccb37f05c0225c0830718d3dfb61f3207e) by removing unused `firmware_size.rs` file, kept some `fn firmware_size` documentation, and rebased.

frostsnap_coordinator/src/firmware_bin.rs
```diff
@@ -0,0 +1,113 @@
...
+            Err(_) => {
+                let firmware_only = match self.find_signature_block() {
+                    Some(sig_start) => &self.bin[..sig_start],
+                    None => self.bin,
+                };
+                let mut state = sha2::Sha256::default();
+                state.update(firmware_only);
+                Sha256Digest(state.finalize().into())
+            }
+        }
```
LLFourn: Hmm I'm not sure about the above two functions.

LLFourn: Fixed my concerns in https://github.com/frostsnap/frostsnap/pull/360/commits/d3e7296a22f2c8f8761c384ee557434acca81258

frostsnap_coordinator/src/firmware.rs
```diff
@@ -0,0 +1,351 @@
...
+        Self {
+            digest,
+            version: VersionNumber::from_digest(&digest),
+        }
+    }
+
+    pub fn features(&self) -> FirmwareFeatures {
+        match self.version {
+            Some(version) => version.features(),
+            None => FirmwareFeatures::all(),
```
nickfarrow, line 35: If the version is unknown then we should assume no features are supported.


Upgrading unversioned dev device ->  `any_device_needs_legacy=false`, when it should be true.

frostsnap_coordinator/src/firmware.rs
```diff
@@ -0,0 +1,351 @@
...
+                    "Device firmware version newer app. Cannot downgrade firmware. Upgrade the app."
+                        .to_string(),
+            },
+            (Some(_), None) => FirmwareUpgradeEligibility::CannotUpgrade {
+                reason: "This is a development app. Cannot upgrade proper device.".to_string(),
+            },
+            (Some(device_ver), Some(app_ver)) => {
+                if app_ver > device_ver {
+                    FirmwareUpgradeEligibility::CanUpgrade
+                } else {
```
nickfarrow, line 281: I would think that a coordinator would never know about a version of device firmware that is newer than what it has bundled(?)

frostsnap_coordinator/src/firmware.rs
nickfarrow: But `None` would likely be annoying on dev builds since we want to be able to test everything and closely similar to production.

frostsnap_coordinator/src/firmware.rs
LLFourn: Yes. device_ver will be 'None' in that case.

LLFourn: Ah so this branch should be no upgrade needed.

LLFourn: ah the no upgrade case needed is handled earlier in the function:

```rust
        if *device_digest == self.firmware_version.digest {
            return FirmwareUpgradeEligibility::UpToDate;
        }
```

LLFourn: I made the branch like this for clarity:

```rust
                if app_ver > device_ver {
                    FirmwareUpgradeEligibility::CanUpgrade
                } else if app_ver == device_ver {
                    // This should be unreachable -- if they have the same version then they should have the same digest
                    FirmwareUpgradeEligibility::UpToDate
                } else {
                    FirmwareUpgradeEligibility::CannotUpgrade {
                        reason: "Device firmware version newer app. Cannot downgrade firmware. Upgrade the app.".to_string(),
                    }
                }
```

frostsnap_coordinator/src/firmware.rs
```diff
@@ -0,0 +1,351 @@
...
+
+        let mut firmware_only_state = sha2::Sha256::default();
+        firmware_only_state.update(&firmware.as_bytes()[..firmware_size as usize]);
+        let firmware_only_digest = Sha256Digest(firmware_only_state.finalize().into());
+
+        let is_signed = firmware_size < total_size;
+
+        if is_signed {
+            // Signed firmware MUST be in KNOWN_FIRMWARE_VERSIONS
+            VersionNumber::from_digest(&digest_with_signature).ok_or(
```
nickfarrow: This seems to disagree with the documentation on `KNOWN_FIRMWARE_VERSIONS`. Which should we store? 

I was expecting to lookup based on the deterministic build digest.

PR #367, opened 2025-10-13
Fine grained nonce streams

LLFourn:
UX improvement over #366

### DIFF
--- a/frostsnap_coordinator/src/nonce_replenish.rs
+++ b/frostsnap_coordinator/src/nonce_replenish.rs
@@ -1,15 +1,18 @@
 use crate::{Completion, Sink, UiProtocol};
 
-use frostsnap_comms::CoordinatorSendMessage;
+use frostsnap_comms::{CoordinatorSendBody, CoordinatorSendMessage, Destination};
 use frostsnap_core::{
     coordinator::{CoordinatorToUserMessage, NonceReplenishRequest},
+    message::signing::OpenNonceStreams,
     DeviceId,
 };
-use std::collections::{BTreeSet, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet, VecDeque};
 
 pub struct NonceReplenishProtocol {
     state: NonceReplenishState,
-    messages: Vec<CoordinatorSendMessage>,
+    pending_messages: HashMap<DeviceId, VecDeque<OpenNonceStreams>>,
+    awaiting_response: HashSet<DeviceId>,
+    completed_streams: u32,
     sink: Box<dyn Sink<NonceReplenishState>>,
 }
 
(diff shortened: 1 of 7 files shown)


frostsnap_core/src/device/device_to_user.rs
```diff
@@ -21,4 +22,5 @@ pub enum DeviceToUserMessage {
         bip32_path: BitcoinBip32Path,
     },
     Restoration(restoration::ToUserRestoration),
+    NonceJobs(NonceJobBatch),
```
nickfarrow, line 25: Curious what the purpose of this ToUser message is, is this laying the groundwork for a hold-to-confirm to access EFUSE keys?

frostsnap_core/src/device_nonces.rs
```diff
@@ -16,6 +15,123 @@ use schnorr_fun::{
...
+        chacha.apply_keystream(&mut secret_nonce_bytes);
+        let secret_nonce = binonce::SecretNonce::from_bytes(secret_nonce_bytes)
+            .expect("computationally unreachable");
+
+        // Update state for next iteration
+        self.seed_material = next_seed_material;
+        self.index += 1;
+
+        // Only collect nonces once we reach the skip_to index
+        if current_index >= self.skip_to {
```
nickfarrow, line 64: Am curious if there's any purpose of still doing the cryptographic computation before skipping?

frostsnap_core/src/device_nonces.rs
```diff
@@ -376,18 +483,15 @@ impl SecretNonceSlot {
...
+            seed_material: self.ratchet_prg_seed_material,
+            // Start from the slot's current index, not the requested start
+            // The task will skip forward as needed
+            index: self.index,
+            target_index: last,
+            nonces: Vec::with_capacity(length),
+            skip_to: start,
         }
     }
     pub fn iter_pub_nonces<'a>(
```
nickfarrow: This is now unused, we do this inline within `do_work`: 
```
self.nonces
    .push(NonceKeyPair::from_secret(secret_nonce).public());
}
```

frostsnap_core/src/device/device_to_user.rs
LLFourn: The nonce jop is executed at the app level (esp32_run.rs) so that's the only way top get something to go there. But yes it naturally allows us to remove the the hmac from the recv_coord_message method and easily opens up `hold_to_confirm` for any nonce stuff.

frostsnap_core/src/device_nonces.rs
LLFourn: oh that's an interesting catch. The most computationally intensive thing is the `from_secret` so yeah we could just frontload the computation for the skipped ones.

LLFourn: In the end I left the code as is and left a comment explaining that some `do_works` are faster than others.
The intensive cryptographic computation is `from_secret` which is what we are skipping.

frostsnap_core/src/device_nonces.rs
LLFourn: I got rid of it.

PR #371, opened 2025-10-15
[app] Make cameras work on all platforms

LLFourn:
Add a native camera for the platforms that mobile_scanner doesn't support (windows, linux).

Not tested on windows.


LLFourn:
This doesn't work on windows but even on windows it's still an improvement. It says no cameras found rather than just dooming the app.

LLFourn:
I think we did this already.

PR #376, opened 2025-10-16
[app/coord] Fix ConnectedDevice upgrade eligibility

LLFourn:
We should not have been trying to put the entire firmware binary even as an opaque type up to rust. We just needed the version.

In general it's best to avoid opaque types in non-opaque types unless you want to be very careful about doing it. You have to wrap stuff in `RustOpaque` or `RustAutoOpaque` which creates a lot of friction.

### DIFF
--- a/frostsnapp/rust/src/device_list.rs
+++ b/frostsnapp/rust/src/device_list.rs
@@ -74,19 +74,18 @@ impl DeviceList {
                 firmware_digest,
                 latest_firmware_digest,
             } => {
-                use frostsnap_coordinator::frostsnap_comms::Sha256Digest;
                 use frostsnap_coordinator::FirmwareVersion;
 
-                let connected = self.connected.entry(id).or_insert(api::ConnectedDevice {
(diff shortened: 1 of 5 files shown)


frostsnapp/rust/src/device_list.rs
```diff
@@ -74,19 +74,18 @@ impl DeviceList {
                 firmware_digest,
                 latest_firmware_digest,
             } => {
-                use frostsnap_coordinator::frostsnap_comms::Sha256Digest;
                 use frostsnap_coordinator::FirmwareVersion;
 
-                let connected = self.connected.entry(id).or_insert(api::ConnectedDevice {
-                    firmware: FirmwareVersion::new(Sha256Digest([0u8; 32])),
-                    latest_firmware: crate::FIRMWARE.and_then(|fw| fw.validate().ok()),
```
LLFourn, line 82: Doing this firmware validation for every new device was not great.

nickfarrow: Oh good find

nickfarrow (approved):
Tested backup workflow and upgrades working

PR #381, opened 2025-10-21
[app,api] Rework send flow modelling

evanlinjin:
* Move the majority of the modelling to rust
* Simplify feerate picker
* Simplify send max

Also some styling adjustments:

* Make input borders rounder in send flow
* Fee summary is now not red
* Removed text-wrapping in wallet toolbar

### DIFF
--- a/frostsnapp/rust/src/api/super_wallet.rs
+++ b/frostsnapp/rust/src/api/super_wallet.rs
@@ -217,16 +217,18 @@ impl SuperWallet {
         master_appkey: MasterAppkey,
         target_addresses: Vec<RustAutoOpaque<Address>>,
         feerate: f32,
-    ) -> i64 {
+    ) -> u64 {
         let mut wallet = self.inner.lock().unwrap();
-        wallet.calculate_avaliable_value(
-            master_appkey,
-            target_addresses
-                .into_iter()
-                .map(|a| a.blocking_read().clone()),
-            feerate,
-            true,
-        )
+        wallet
+            .calculate_avaliable_value(
+                master_appkey,
+                target_addresses
+                    .into_iter()
+                    .map(|a| a.blocking_read().clone()),
+                feerate,
+                true,
+            )
+            .max(0) as u64
     }
 
     /// Start building transaction.
--- a/frostsnapp/rust/src/api/transaction.rs
+++ b/frostsnapp/rust/src/api/transaction.rs
@@ -147,7 +147,7 @@ impl BuildTxInner {
         super_wallet: &SuperWallet,
         master_appkey: MasterAppkey,
         recipient: u32,
-    ) -> Option<i64> {
+    ) -> Option<u64> {
         Some(
             super_wallet.calculate_available(
                 master_appkey,
@@ -260,7 +260,7 @@ impl BuildTxState {
     ///
     /// Returns `None` if no feerate is specified.
     #[frb(sync)]
-    pub fn available_amount(&self, recipient: u32) -> Option<i64> {
+    pub fn available_amount(&self, recipient: u32) -> Option<u64> {
         let inner = self.inner.read().unwrap();
         inner.available_amount(
             &self.super_wallet,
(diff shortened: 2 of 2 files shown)


frostsnapp/flutter_rust_bridge.yaml
```diff
@@ -2,3 +2,4 @@ rust_input: crate::api
 rust_root: rust/
 dart_output: lib/src/rust
 stop_on_error: true
+type_64bit_int: true
```
evanlinjin, line 5: This is so we can remove all `#[frb(type_64bit_int)]`.

evanlinjin:
Turns out this PR is not ready yet - the behavior when we are offline is a little quirky.

We cannot set the amount without an "available balance", and we can't get an "available balance" without the feerate, and the default feerate is "medium - ETA 2 blocks" which we can only fetch with an internet connection. Therefore, we can only set the amount if we have an internet connection or if the user sets the feerate manually.

I'm going to change the behavior when we are offline to ask for the feerate explicitly after getting the recipient address.

frostsnapp/rust/src/api/broadcast.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+#[derive(Debug, Copy, Clone)]
+pub enum StartError {
+    /// Occurs when `BroadcastSubscription` is already started.
+    AlreadyRunning,
+}
+
+impl<T> BroadcastSubscription<T> {
+    fn _id(&self) -> u32 {
```
LLFourn: please no `_` prefix in rust unless it's unused.

frostsnapp/rust/src/api/broadcast.rs
```diff
@@ -0,0 +1,133 @@
...
+                tracing::event!(Level::ERROR, id, "Failed to add to sink");
+            }
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct BroadcastSubscription<T> {
+    id: u32,
+    is_running: Arc<AtomicBool>,
```
LLFourn: here too.

aphanley:
A couple of things I noticed during testing:

1. If you are offline and attempt to broadcast a transaction, then it will fail (obviously - though this is an example of an ugly technical error message we should improve). If you then come back online it will automatically broadcast. I think intuitively this should probably require manual broadcast again rather than happen automatically. The worst part of this is that if you "forget" a transaction which you have attempted to broadcast while offline, it will still be broadcast automatically when you come back online.

As an aside, we should also change "forget" to "cancel":
It should be cancel when the transaction is not yet broadcast or exported as psbt. It should be forget if it has been broadcast or exported

2. On android the fullscreen action dialog's look a bit strange, maybe on mobile we hide the tab at the top and the separator line at the bottom?

3. On android the "B" bitcoin symbol is cut off on my pixel 6a in a lot of places like transaction history, balance, send amount etc.

4. On android the scan button says "Scan (coming soon)" even though it is blue and fully functional

LLFourn:
I addressed alex's very helpful review comments:

> A couple of things I noticed during testing:
> 
>     1. If you are offline and attempt to broadcast a transaction, then it will fail (obviously - though this is an example of an ugly technical error message we should improve). If you then come back online it will automatically broadcast. I think intuitively this should probably require manual broadcast again rather than happen automatically. The worst part of this is that if you "forget" a transaction which you have attempted to broadcast while offline, it will still be broadcast automatically when you come back online.

This can't be fixed here. Has to be fixed in the underlying library.

> 
> 
> As an aside, we should also change "forget" to "cancel": It should be cancel when the transaction is not yet broadcast or exported as psbt. It should be forget if it has been broadcast or exported

I did this except I didn't bother to check whether it has been broadcast as a PSBT or not. It does mention that you have may have done this when you press cancel.

> 
>     2. On android the fullscreen action dialog's look a bit strange, maybe on mobile we hide the tab at the top and the separator line at the bottom?

this is fixed #403 
> 
>     3. On android the "B" bitcoin symbol is cut off on my pixel 6a in a lot of places like transaction history, balance, send amount etc.

This is not introduced by this PR.
> 
>     4. On android the scan button says "Scan (coming soon)" even though it is blue and fully functional

Fixed.

Merging. There's nothing more to do on this PR.

PR #386, opened 2025-10-29
Backup record workflow confirm after recorded

LLFourn:
Built on top of #354 

1. You have to confirm on the device to mark the backup as recorded. 
2. Remove the "backup manager" API which returned streams that didn't have rich data to display (it had to be zipped together with other API calls which was messy). It's just part of the coordinator now.
3. Add a dialog to force the user to write down the outside backup card data explicitly.
4. Make the user put one device in at a time when backing up to avoid confusion and have a more elegant UI.

### DIFF
--- a/frostsnapp/rust/src/coordinator.rs
+++ b/frostsnapp/rust/src/coordinator.rs
@@ -736,9 +736,7 @@ impl FfiCoordinator {
             })?
         };
 
-        if let Some(stream) = &*self.key_event_stream.lock().unwrap() {
-            stream.send(self.key_state());
-        }
+        self.emit_key_state();
 
         Ok(restoration_id)
     }
@@ -766,9 +764,7 @@ impl FfiCoordinator {
             })?
         };
 
-        if let Some(stream) = &*self.key_event_stream.lock().unwrap() {
-            stream.send(self.key_state());
-        }
+        self.emit_key_state();
 
         Ok(restoration_id)
     }
@@ -793,9 +789,7 @@ impl FfiCoordinator {
             })?;
         }
 
-        if let Some(stream) = &*self.key_event_stream.lock().unwrap() {
-            stream.send(self.key_state());
-        }
+        self.emit_key_state();
         Ok(())
     }
 
@@ -805,29 +799,21 @@ impl FfiCoordinator {
         recover_share: &RecoverShare,
         encryption_key: SymmetricKey,
     ) -> Result<()> {
+        let held_by = recover_share.held_by;
         {
-            let held_by = recover_share.held_by;
-            {
-                let mut db = self.db.lock().unwrap();
-                let mut coordinator = self.coordinator.lock().unwrap();
-                coordinator.staged_mutate(&mut *db, |coordinator| {
-                    coordinator.recover_share(
-                        access_structure_ref,
-                        recover_share,
-                        encryption_key,
-                    )?;
-                    Ok(())
-                })?;
-            }
-
-            self.exit_recovery_mode(held_by, encryption_key);
-
-            if let Some(stream) = &*self.key_event_stream.lock().unwrap() {
-                stream.send(self.key_state());
-            }
-
-            Ok(())
+            let mut db = self.db.lock().unwrap();
+            let mut coordinator = self.coordinator.lock().unwrap();
+            coordinator.staged_mutate(&mut *db, |coordinator| {
+                coordinator.recover_share(access_structure_ref, recover_share, encryption_key)?;
+                Ok(())
+            })?;
         }
+
+        self.exit_recovery_mode(held_by, encryption_key);
+
+        self.emit_key_state();
+
+        Ok(())
     }
 
     pub fn get_restoration_state(&self, restoration_id: RestorationId) -> Option<RestorationState> {
@@ -865,24 +851,22 @@ impl FfiCoordinator {
             self.exit_recovery_mode(device_id, encryption_key);
         }
 
-        if let Some(stream) = &*self.key_event_stream.lock().unwrap() {
-            stream.send(self.key_state());
-        }
+        self.emit_key_state();
 
         Ok(assid)
     }
 
     pub fn delete_key(&self, key_id: KeyId) -> Result<()> {
-        let mut db = self.db.lock().unwrap();
-        let mut coordinator = self.coordinator.lock().unwrap();
-        coordinator.staged_mutate(&mut *db, |coordinator| {
-            coordinator.delete_key(key_id);
-            Ok(())
-        })?;
-        drop(coordinator);
-
-        // Clean up backup runs for deleted key
         {
+            let mut db = self.db.lock().unwrap();
+            let mut coordinator = self.coordinator.lock().unwrap();
+            coordinator.staged_mutate(&mut *db, |coordinator| {
+                coordinator.delete_key(key_id);
+                Ok(())
+            })?;
+            drop(coordinator);
+
+            // Clean up backup runs for deleted key
             let mut backup_state = self.backup_state.lock().unwrap();
             backup_state.mutate2(&mut *db, |state, mutations| {
                 state.clear_backup_run(key_id, mutations);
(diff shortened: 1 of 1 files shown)


frostsnap_coordinator/src/backup_run.rs
```diff
@@ -1,43 +1,40 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DisplayBackupState {
+    pub confirmed: bool,
+    pub legacy_display_confirmed: bool,
+}
+
 #[derive(Default, Debug)]
 pub struct BackupState {
-    // Maps each access_structure_ref to a map of device_id -> timestamp.
-    runs: BTreeMap<AccessStructureRef, BTreeMap<DeviceId, Option<u32>>>,
```
nickfarrow, line 12: Even though we're removing the "too quick" feature, it could be useful to leave the confirmation as a timestamps that can be used to prompt a backup/health check in the future.

Though on second thought this would probably be an entirely separate feature, "connect device X or validate backup". If we did introduce health checks, we could set the time of the first Backup Run completion as the first timestamp, then prompt at t+N from there.

frostsnap_coordinator/src/backup_run.rs
```diff
@@ -1,43 +1,40 @@
...
-                        true
-                    } else {
-                        false
-                    }
-                } else {
-                    false
-                }
+                let run = self.runs.entry(*access_structure_ref).or_default();
+                run.insert(*share_index, true);
+                true
```
nickfarrow: We always have a mutation here since it isn't possible to `MarkShareComplete` twice? Otherwise it would be possible to store redundant mutations.

Was wondering if we should return something like `true != run.insert(*share_index, true)` here.

frostsnap_coordinator/src/backup_run.rs
```diff
@@ -1,43 +1,40 @@
...
+use frostsnap_core::{AccessStructureId, AccessStructureRef, Gist, KeyId};
 use rusqlite::params;
 use std::collections::BTreeMap;
-use std::time::{SystemTime, UNIX_EPOCH};
 use tracing::{event, Level};
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DisplayBackupState {
+    pub confirmed: bool,
+    pub legacy_display_confirmed: bool,
```
nickfarrow: Why do we need this `legacy_display_confirmed` field?

Can't we just map legacy `DisplayBackupConfirmed` messages to `confirmed: true`? The share is backed up either way.

The current code skips `markBackupComplete` for legacy devices:
```dart
if (isComplete && !result.legacyDisplayConfirmed) {
  await coord.markBackupComplete(...);
}
```

But legacy devices are still backing up a specific share at a specific index. Then we wouldn't need a new `BackupRecorded` enum variant either. Is there a reason we need to distinguish between legacy and new confirmations?

frostsnap_comms/src/lib.rs
```diff
@@ -465,7 +465,11 @@ pub enum CommsMisc {
     /// Tells the coordinator that a device has confirmed to show it's backup.
     /// core doesn't provide a way to tell the coordinator that showing the backup was confirmed so
     /// we have this here.
+    /// DEPRECATED: Use BackupRecorded instead.
     DisplayBackupConfrimed,
```
nickfarrow, line 469: Just noticed we had this as "Confrimed", oops

frostsnap_core/src/coordinator.rs
```diff
@@ -1710,6 +1710,16 @@ impl CoordAccessStructure {
         }
         map
     }
+
+    pub fn iter_shares(&self) -> impl Iterator<Item = (DeviceId, ShareIndex)> + '_ {
+        self.share_index_to_devices()
```
nickfarrow: We already have `self.device_to_share_index`, could just 

```
pub fn iter_shares(&self) -> impl Iterator<Item = (DeviceId, ShareIndex)> + '_ {
    self.device_to_share_index
        .iter()
        .map(|(&device_id, &share_index)| (device_id, share_index))
}
```

frostsnapp/lib/backup_workflow.dart
```diff
@@ -254,121 +394,346 @@ class _BackupChecklistState extends State<BackupChecklist> {
...
+                    Widget deviceStatusContent;
+                    IconData statusIcon;
+                    Color? statusIconColor;
+
+                    if (deviceCount > 1) {
+                      // Multiple devices warning
+                      statusIcon = Icons.warning_sharp;
+                      statusIconColor = theme.colorScheme.primary;
+                      deviceStatusContent = Text(
+                        'Multiple devices connect. Connect only one device at a time.',
```
nickfarrow: Typo: connect -> connected.

Maybe `Multiple devices connected. Backup only one device at a time.`

frostsnapp/lib/device_action_backup.dart
```diff
@@ -28,53 +25,85 @@ class DeviceActionBackupController with ChangeNotifier {
...
+                    crossAxisAlignment: CrossAxisAlignment.start,
+                    children: [
+                      Icon(
+                        Icons.warning_amber_rounded,
+                        color: theme.colorScheme.error,
+                      ),
+                      const SizedBox(width: 12),
+                      Expanded(
+                        child: Text(
+                          'This key backup is secret information. Anyone with access to ${accessStructure.threshold()} of the ${accessStructure.devices().length} keys can steal all your Bitcoin.',
```
nickfarrow: nit: `bitcoin` over `Bitcoin`

nickfarrow (commented):
Looks good, nice migration tests. Most important question for me is whether we can get rid of the need for `legacy_backup_confirmed`

Haven't battle tested yet

frostsnap_coordinator/src/backup_run.rs
LLFourn: This is a good point. I think perhaps health checks can just be signing something. Optionally you could check the backup as well. but I don't think that determines when you need to do the health check. Also signing a tx would reset the timer.

I think if we want a backup check timer then that also has to start when you check the backup rather than writing it down.

frostsnap_coordinator/src/backup_run.rs
LLFourn: I'm not sure what you mean. The logic marks it as done regardless of whether it was in the false state beforehand. This doesn't have much affect on the UI but is a better model and a bit more robust.

frostsnap_coordinator/src/backup_run.rs
LLFourn: We can't map them because the new `DisplayBackupConfirmed` will close the fullscreen dialog.

I've rewritten the logic here it was a mess.

frostsnap_core/src/coordinator.rs
LLFourn: no it's not a bijection.

frostsnap_coordinator/src/backup_run.rs
nickfarrow: My point here was about not needing to `mutations.push(mutation);` in the case where `run` already has `true` for that share index, pushing a redundant mutation onto the log. Repeated `mark_backup_complete`s.

But this is probably prevented by the protocol flow.

frostsnap_core/src/coordinator.rs
nickfarrow: Currently `iter_shares` calls `share_index_to_devices()` to group devices by share, then immediately flattens it back to `(DeviceId, ShareIndex)` pairs - the same pairs already in `device_to_share_index` (?)

LLFourn: You're right sorry I totally misunderstood the context.

aphanley:
I was able to get it in a situation where it said 1/2 backups recorded, but both are ticked

To recreate:
create a 2/2 in a previous version
backup one of the devices
close the app and restart on this branch (resetting the backup completed since it was only partially completed)
backup one of the keys

I think it's probably fine to just ignore this since it is a migration issue (we already said we would set it to mark as none backed up if backups were only partially complete)

image

aphanley:
Looks good

PR #389, opened 2025-11-05
[app] Improve startup error page

LLFourn:


nickfarrow (approved):
Rebased onto master. Two things changed:
- Dropped the `main()` restructure -- the unconditional `runApp(MyApp(...))` left `buildMainWidget` uncalled, so the normal startup path lost contexts (plus wakelock/orientation/recovery-mode setup). May have been the start of a greater redesign? 
- Copy button is now the shared `CopyIconButton` to match master.

Tested, looks good.

PR #406, opened 2025-12-04
[app] Add install udev rules thing in settings

LLFourn:


frostsnapp/lib/udev_setup.dart
```diff
@@ -0,0 +1,317 @@
...
+    '-c',
+    'cat > $_rulesPath << \'EOF\'\n$_rulesContent\nEOF\n'
+        'udevadm control --reload-rules && udevadm trigger',
+  ]);
+  return result.exitCode == 0;
+}
+
+Future<bool> _removeUdevRules() async {
+  final result = await Process.run('pkexec', [
+    'bash',
```
LLFourn: there's no reason to use bash here.

frostsnapp/lib/udev_setup.dart
```diff
@@ -0,0 +1,317 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final isInstalled = _rulesInstalled == true;
+    final isLoading = _rulesInstalled == null;
+
+    return Center(
+      child: Container(
+        constraints: BoxConstraints(maxWidth: 500),
```
LLFourn: I don't really like this random width constraint we threw in here.

frostsnapp/lib/udev_setup.dart
```diff
@@ -0,0 +1,317 @@
...
+Future<bool> _areRulesInstalled() async {
+  return File(_rulesPath).exists();
+}
+
+Future<bool> _installUdevRules() async {
+  final result = await Process.run('pkexec', [
+    'bash',
+    '-c',
+    'cat > $_rulesPath << \'EOF\'\n$_rulesContent\nEOF\n'
+        'udevadm control --reload-rules && udevadm trigger',
```
LLFourn: how come in this we do udevadm trigger but in the manual instructions we don't.

LLFourn: (we just ask them to plug the device back in, in the manual instruction)

frostsnapp/lib/udev_setup.dart
```diff
@@ -0,0 +1,317 @@
...
+const _rulesContent = '''# Frostsnap USB device
+SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ATTRS{idVendor}=="303a", ATTRS{idProduct}=="1001", MODE="0666"
+''';
+
+const _manualCommands = '''sudo cat > $_rulesPath << 'END'
+# Frostsnap USB device
+SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ATTRS{idVendor}=="303a", ATTRS{idProduct}=="1001", MODE="0666"
+END
+sudo udevadm control --reload-rules''';
+
```
nickfarrow, line 20: This still gives `/etc/udev/rules.d/99-frostsnap.rules: Permission denied` due to order of things opening.

An alternative:
```
cat << 'END' | sudo tee /etc/udev/rules.d/99-frostsnap.rules
SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ATTRS{idVendor}=="303a", ATTRS{idProduct}=="1001", MODE="0666"
END
```

LLFourn: Thanks fixed.

frostsnapp/lib/udev_setup.dart
LLFourn: I got rid of the triggers in favor of unplugging and plugging back in.

PR #423, opened 2025-12-18
App cleanup and enable camera for all platforms

nickfarrow:


frostsnapp/lib/settings.dart
```diff
@@ -1236,20 +1252,26 @@ Future<void> _showEraseAllDialog(BuildContext context) async {
...
 
-  final currentUpdate = await GlobalStreams.deviceListSubject.first;
-  final devicesToErase = currentUpdate.state.devices
-      // intentionally erasing all and not filtering out "blank"
-      .map((device) => device.id)
-      .toList();
+  // Listen for devices being wiped/disconnected and remove them from action list
+  final subscription = GlobalStreams.deviceListChangeStream.listen((change) {
+    if (change.kind == DeviceListChangeKind.removed) {
+      controller.removeActionNeeded(change.device.id);
```
LLFourn, line 1261: I wonder why this would fix the problem. My guess is there's a race condition where if a device reconnects fast enough or the app is slow enough it can put the dialog builder into a bad state (maybe a panic or something?) where the device is back but the dialog has already closed and have disposed itself.

LLFourn (approved):

PR #425, opened 2026-01-03
[app] Don't edit settings in a bottom sheet

LLFourn:
Because it can be covered by the keyboard. Quick edits of things that need keyboard should go in normal dialogs.


LLFourn:
Thanks @nickfarrow I separated the unrelated fix into a different commit.

PR #426, opened 2026-01-03
[coordinator] implement happy eyeballs for electrum

LLFourn:
- Implement RFC 8305 Happy Eyeballs for Electrum server connections
  - Fixes connection failures when IPv6 is unavailable but the server resolves to both address families

I didn't study the RFC or anything but it seems like a legit approach.


nickfarrow (approved):
tACK

PR #429, opened 2026-01-27
Fix just run with app flavours

nickfarrow:
Introduces little warning when doing `just run` on non-android targets:
```
--flavor is only supported for Android, macOS, and iOS devices. Flavor-related
features may not function properly and could behave differently in a future release.
```
but I think this worth experiencing, rather than messily reacting to platform within our justfile.

### DIFF
--- a/frostsnapp/justfile
+++ b/frostsnapp/justfile
@@ -24,14 +24,14 @@ run +ARGS="": maybe-gen
     just ../gen-firmware
     BUILD_COMMIT=$(just get-build-commit)
     BUILD_VERSION=$(just get-build-version)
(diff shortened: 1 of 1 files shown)


LLFourn:
What error/issue were you seeing that prompted this fix? Was `just run` failing on Android without the explicit flavor?

nickfarrow:
Yep, for me at least:
```
INFO: Building rust_lib_frostsnapp for aarch64-linux-android
Running Gradle task 'assembleDebug'...                            112.3s
Error: Gradle build failed to produce an .apk file. It's likely that this file was
generated under /home/nick/repos/frostsnap/frostsnapp/build, but the tool couldn't
find it.
error: Recipe `run` failed with exit code 1
error: Recipe `run` failed on line 107 with exit code 1
```

PR #432, opened 2026-02-20
[widget] Recovery keyboard improvements (Keyboard B - Remove letters)

aphanley:
This branch overhauls the backup recovery word entry UI on the device. The main changes:                                                                  
                                                                                                                                                            
  Keyboard & Input                                                                                                                                          
  - Compact alphabetic keyboard that hides invalid letters based on BIP39 word list context
  - Numeric keyboard for share index entry, with styled disabled states (greyed text/icon only, normal button background)                                   
  - Checkmark icon turns green when enabled                                                                                                                 
  - Word selector shown for 1-8 possible matches with fixed positions
  - Backspace navigation fixed to properly return to previous words

  Input Preview & Scrolling
  - Migrated to Gray4 anti-aliased font rendering in a framebuffer
  - Smooth scroll animation when advancing to the next word
  - Scroll animation is independent of keyboard scrolling — dragging the keyboard no longer interrupts it
  - fast_forward_scrolling only triggers on actual key selection, not touch-down
  - Word numbers displayed with dots and proper spacing

  Touch Handling
  - Touch outlines clipped to keyboard area (backspace draws unclipped in preview area)
  - Touch feedback rectangles vertically adjusted to center on text
  - Lift-up events guarded against pending model updates to prevent double-actions

  Status Bar (bottom feedback)
  - Flat edges (no rounded corners)
  - Border color matches fill
  - Valid state shows "Success" / "Valid Backup", with 1-second delay before auto-submitting and touches disabled
  - Text capitalization fixes ("Invalid Backup", "Tap word to edit")

  Visual Polish
  - Progress bar color changed to Frostsnap logo blue
  - Scrollbar flicker reduced by only repainting changed thumb regions
  - Buttons expand to fill flex allocation properly

### DIFF
--- a/frostsnap_widgets/src/backup/alphabetic_keyboard.rs
+++ b/frostsnap_widgets/src/backup/alphabetic_keyboard.rs
@@ -63,10 +63,9 @@ fn draw_char_to_framebuffer(
             && py >= 0
             && (px as u32) < FRAMEBUFFER_WIDTH
             && (py as u32) < FRAMEBUFFER_HEIGHT
+            && gray.luma() > 0
         {
-            if gray.luma() > 0 {
-                let _ = Pixel(Point::new(px, py), gray).draw(fb);
-            }
+            let _ = Pixel(Point::new(px, py), gray).draw(fb);
         }
     }
 }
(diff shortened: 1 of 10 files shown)


aphanley:
I have pushed two branches so we can do some internal testing and even some user testing to see which version we prefer

recovery-keyboard-A
recovery-keyboard-B

A: is the version where the keyboard is always the same and only greys out invalid letters
B: is the version where invalid letters are removed, reducing the size of the keyboard

Having played with both and getting feedback from a couple of others, I lean towards option A.

Option B is faster once you get used to it, but because the keyboard jumps around I think it causes a little bit of uncertainty and anxiety for new users. Option A is more intuitive from the start. My view is that recovery (even just testing recovery) is a stressful process and that we want it to feel as straightforward and calming as possible and that this perception is more important than speed.

PR #440, opened 2026-03-11
Improve backup reminder UX and wallet creation polish

aphanley:
- Backup reminder dialog now intercepts the Receive flow when backups are incomplete — user can dismiss it to proceed or go to the backup screen
  - "Unfinished backups" warning banner is amber by default, upgrades to red when the wallet has funds
  - Added "Mark backups as complete" option on backup keys screen to manually dismiss the notification
  - Removed auto-popup of backup dialog after wallet creation — it now only appears via the warning banner or when opening the receive flow without having completed backups
  - Device name input uses default font and fixes key icon alignment with text fields in wallet creation

### DIFF
--- a/frostsnapp/lib/keygen.dart
+++ b/frostsnapp/lib/keygen.dart
@@ -5,11 +5,13 @@ import 'package:frostsnap/maybe_fullscreen_dialog.dart';
 import 'package:frostsnap/src/rust/api/coordinator.dart';
 import 'package:frostsnap/theme.dart';
 
-void showWalletCreatedDialog(
+/// Returns `true` if the user dismissed the dialog (pressed "Later"),
+/// `false` if they chose to secure the wallet.
+Future<bool> showWalletCreatedDialog(
   BuildContext context,
   AccessStructure accessStructure,
 ) async {
-  await showDialog(
+  final dismissed = await showDialog<bool>(
     context: context,
     barrierDismissible: false,
     builder: (BuildContext context) {
@@ -80,7 +82,7 @@ void showWalletCreatedDialog(
           actions: [
             TextButton(
               onPressed: () {
-                Navigator.of(context).pop();
+                Navigator.of(context).pop(true);
               },
               child: const Text('Later'),
             ),
@@ -104,4 +106,5 @@ void showWalletCreatedDialog(
       );
     },
   );
+  return dismissed == true;
 }
--- a/frostsnapp/lib/wallet.dart
+++ b/frostsnapp/lib/wallet.dart
@@ -1,10 +1,10 @@
 import 'dart:async';
 import 'package:flutter/material.dart';
-import 'package:frostsnap/backup_workflow.dart';
 import 'package:frostsnap/contexts.dart';
 import 'package:frostsnap/device_list.dart';
 import 'package:frostsnap/global.dart';
 import 'package:frostsnap/id_ext.dart';
+import 'package:frostsnap/keygen.dart';
 import 'package:frostsnap/maybe_fullscreen_dialog.dart';
 import 'package:frostsnap/nonce_replenish.dart';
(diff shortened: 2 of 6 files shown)


frostsnapp/lib/keygen.dart
```diff
@@ -5,11 +5,13 @@ import 'package:frostsnap/maybe_fullscreen_dialog.dart';
 import 'package:frostsnap/src/rust/api/coordinator.dart';
 import 'package:frostsnap/theme.dart';
 
-void showWalletCreatedDialog(
+/// Returns `true` if the user dismissed the dialog (pressed "Later"),
+/// `false` if they chose to secure the wallet.
+Future<bool> showWalletCreatedDialog(
```
LLFourn, line 10: This function is no longer only shown post-creation — it's called from the Receive button and the backup banner too. Should be renamed (e.g. `showSecureWalletDialog`) and the title/body copy ("Wallet created! Now let's secure it.") updated to reflect its new role as a backup reminder rather than a creation celebration.

frostsnapp/lib/wallet.dart
```diff
@@ -713,17 +713,37 @@ class WalletBottomBar extends StatelessWidget {
...
-        title: Text('Receive'),
-        builder: (context, scrollController) => walletCtx.wrap(
-          ReceivePage(
-            wallet: walletCtx.wallet,
-            txStream: walletCtx.txStream,
-            scrollController: scrollController,
+      onPressed: () async {
+        // Check if backups are incomplete — if so, show the secure wallet dialog first
+        final backupStream = walletCtx.backupStream;
+        if (backupStream is BehaviorSubject<BackupRun>) {
```
LLFourn, line 719: This `is BehaviorSubject<BackupRun>` cast is fragile — if the stream gets wrapped or the implementation changes, the check silently fails and the backup gate doesn't fire. `coord.getBackupRun(keyId: walletCtx.keyId)` already exists as a synchronous API, just use that directly.

frostsnapp/lib/wallet.dart
```diff
@@ -713,17 +713,37 @@ class WalletBottomBar extends StatelessWidget {
...
+          final backupRun = backupStream.valueOrNull;
+          if (backupRun != null &&
+              !backupRun.devices.every((device) => device.complete == true)) {
```
LLFourn, line 722: `device.complete == true` treats `null` as incomplete, but `null` means the device predates backup tracking and should be treated as complete. Should be `device.complete != false` to match the banner's logic.

LLFourn (commented):
The core idea is good — moving the backup prompt out of the creation flow and into contextual moments where the user has real stake (like receiving funds) is a better UX. A few things need addressing:

### Must fix

**1. `popUntil` crash in the "Secure Wallet" button** (`keygen.dart` ~line 91)

The "Secure Wallet" handler does `Navigator.popUntil(context, (r) => r.isFirst)` and then keeps using the dialog builder's `context` — which is now dead. This will throw in debug mode.

The `popUntil` isn't needed here. `MaybeFullscreenDialog` can open from anywhere, and the backup flow is a big enough interruption that the user won't return to whatever was on the stack. Just pop the dialog normally and show the checklist.

### Should fix

**2. "Mark backups as complete" button** (`backup_workflow.dart` ~line 748)

- Should mark devices individually, not all-or-nothing — let the user say "I backed up this one but not that one yet."
- Needs a confirmation dialog warning about the importance of backups. One accidental tap currently silently marks everything as done.

### Design feedback

**3. Banner `onTap` flow**

When the user taps the "unfinished backups!" banner, they've already expressed intent to deal with backups. Showing them the "do you want to secure?" dialog first feels like a redundant question. That said, there's value in having some kind of backup explainer for users who haven't been through the flow before — maybe something integrated into the checklist itself (an expandable intro section, an info button, or a prompt to watch a video) rather than a separate dialog that pops up, gets dismissed, then opens another dialog.

PR #442, opened 2026-03-12
Verify address dialog improvements

aphanley:
- Redesigned the verify address fullscreen dialog: replaced the old 3-step icon/text instruction layout with a single     
clear instruction and a blurred tap-to-reveal address at the bottom                                                       
- Added _BlurredAddress widget with animated blur/reveal and a caution warning that fades in after reveal                 
- Made the fullscreen dialog body rebuildable via a new rebuild() method on FullscreenActionDialogController, so the    reveal toggle works within the dialog                                                                                  
- Added cautionColor constant to theme

### DIFF
--- a/frostsnapp/lib/wallet_receive.dart
+++ b/frostsnapp/lib/wallet_receive.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:collection';
+import 'dart:ui';
 
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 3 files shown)


frostsnapp/lib/wallet_receive.dart
```diff
@@ -705,71 +707,212 @@ class _ReceiverPageState extends State<ReceivePage> {
...
-            Expanded(
-              flex: 3,
-              child: Text.rich(
-                TextSpan(
+        if (_address != null) ...[
+          SizedBox(height: 8),
+          Text(
+            'Receive Address #${_address!.index}',
+            style: Theme.of(context).textTheme.titleSmall,
+          ),
```
LLFourn, line 724: The state management here has gotten tangled up. Let me walk through what's happening:

1. `_dialogAddressRevealed` lives here on `_ReceiverPageState`
2. But the `_BlurredAddress` widget that reads it lives inside the fullscreen dialog — which is in a completely separate widget subtree (the Navigator overlay). `setState` on `_ReceiverPageState` can't reach it.
3. So when the user taps to reveal, the `onReveal` callback reaches *out* of the dialog back to `_ReceiverPageState` to toggle the boolean, then calls `fullscreenDialogController.rebuild()` to poke the controller into firing `notifyListeners()`, which makes the dialog's `ListenableBuilder` rebuild, which calls `_dialogBodyBuilder` again, which finally passes the new value back *into* the dialog as a prop on `_BlurredAddress`.
4. Then on top of that, `_dialogBodyBuilder` uses a `GlobalKey` to reach *back into* the `_BlurredAddress` it just created to call `buildCaution()` and extract a widget from the child's state — which doesn't even work on the first frame because the child hasn't mounted yet, so there's an invisible duplicate as a fallback.

So the data flow is: dialog → parent state → controller notify → dialog rebuild → dialog child. And then back again via GlobalKey for the caution widget. That's a round-trip across the overlay boundary and back, plus a cross-tree widget extraction, all to manage a single boolean: "is the address blurred or not."

**This is entirely local UI state.** `_BlurredAddress` is already a `StatefulWidget` with animation controllers. It can just hold a `_revealed` boolean and toggle it with its own `setState`. The tap handler, the blur animation, and the caution widget all live inside it already — the state should too.

Suggested fix:
- Move `_revealed` into `_BlurredAddressState`, toggle it with `setState` in the tap handler
- Build the caution widget inside `_BlurredAddress.build()` (eliminates the GlobalKey and the invisible placeholder)
- Delete `_dialogAddressRevealed`, `_blurredAddressKey`, and the `rebuild()` method on the controller
- The reset on dialog open (line 216) is unnecessary — a fresh widget instance starts unrevealed

This removes ~40 lines and three cross-boundary coupling points.

LLFourn (approved):

PR #447, opened 2026-03-18
Add animated progress circle for nonce replenishment

aphanley:
- Replace the basic CircularProgressIndicator in the nonce replenishment widget with a custom-painted progress circle
  featuring an animated multi-colour gradient, glow effects, and a completion celebration animation
  - Colour scheme and gradient style matches the threshold selector widget (primary/secondary/tertiary sweep)
  - Progress estimation smoothly interpolates between coarse-grained stream completion events using time-based extrapolation with an asymptotic curve
  - Move completion/abort callback handling from NonceGenerationPage into MinimalNonceReplenishWidget so the animation finishes before transitioning

### DIFF
--- a/frostsnapp/lib/nonce_replenish.dart
+++ b/frostsnapp/lib/nonce_replenish.dart
@@ -1,119 +1,540 @@
+import 'dart:async';
+import 'dart:math' as math;
+import 'dart:ui' as ui;
+
 import 'package:flutter/material.dart';
 import 'package:frostsnap/src/rust/api/nonce_replenish.dart';
+import 'package:rxdart/rxdart.dart';
 
-/// Minimal nonce replenishment widget for use in wallet creation flow.
-/// Shows aggregate progress with auto-advance capability.
-class MinimalNonceReplenishWidget extends StatelessWidget {
-  final Stream<NonceReplenishState> stream;
-  final VoidCallback? onComplete;
-  final VoidCallback? onAbort;
-  final bool autoAdvance;
+sealed class NonceReplenishTerminal {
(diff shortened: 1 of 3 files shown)


LLFourn:
Force-pushed a squashed, reworked version. One thing worth calling out about the change in value model:

**The original was fake-smooth.** The display ran its own asymptotic timer (`1 - 1/(1+t)`) toward a hardcoded 0.98 ceiling; real progress updates only adjusted the timer's speed. The 0.98 cap was the tell — the display had to be bounded because it was decoupled from truth. The circle kept creeping forward regardless of whether any real work was happening. A stalled device would still look like it was progressing.

**The new version just smooths real updates.** Display is driven by an 800ms tween toward the most recently observed real progress. If the stream is bursty, the circle pauses between updates. That pause is information: it tells the user "no real event in the last second." If it looks janky, that's a signal worth having — something is slow or stuck. A UI that fabricates motion during silence removes the user's ability to distinguish *working* from *stuck*, and it's the first thing to mislead you when debugging a failure.

**In principle:** a progress indicator is a claim about the state of the system. Interpolating between real events is honest — it's a best-guess between samples. Inventing motion during silence is lying. The janky-but-honest version loses some polish in the happy path; it earns it back everywhere else — in support, in debugging, in trust. The default should be truth.

Kept all of @aphanley's visual work — gradient ring, pulse, tip glow, completion flourish. Only the prediction machinery underneath got stripped.

aphanley:
Looks good, i prefer lying but seems good

LLFourn (approved):

PR #455, opened 2026-04-01
Add stack_check tool and CI check

LLFourn:
## Summary
- New `stack_check` CLI tool under `tools/stack_check/` for analyzing device firmware stack usage
- Parses ELF `.stack_sizes` section to show per-function stack frame sizes ranked by size
- `types` subcommand shows type sizes via `-Zprint-type-sizes`
- `--max-pct` flag fails if largest frame exceeds a percentage of available stack
- CI step added to device firmware build job (runs `just stack-check` which defaults to `--max-pct 25`)
- Moved `widget_simulator` into `tools/` directory

## Test plan
- [x] CI should **fail** on this PR because the largest frame (`SignPromptPageList::get` at 32.8%) exceeds the 25% threshold
- [x] Follow-up commits will add `#[inline(never)]` annotations to fix this


Ari4ka (approved):

aphanley (approved):
Tested normal device operations manually with linux coordinator

aphanley (approved):

Ari4ka (approved):

PR #457, opened 2026-04-08
[device] Port to esp-hal v1.2.0

LLFourn:
## Summary

Ports the `device` firmware crate from the frostsnap esp-hal v0.22 fork
to **upstream** esp-hal v1.0.0 — `esp-hal = { version = "1", ... }`,
no fork, no vendoring, no patches.

Three commits:

1. `[device] Port to esp-hal v1.0.0` — the bulk migration. Every
   peripheral touch point, the `Peripherals`-by-value model, the
   `Input`/`Output`/`Uart`/`I2c`/`Spi` config-builder pattern, the
   `TrngSource` + `Trng::try_new` split, the `#[entry]`→`#[main]`
   rename, efuse field renames, `software_reset` move, `mipidsi` 0.8 +
   `display-interface-spi` → `mipidsi` 0.10's built-in `SpiInterface`,
   `embedded-hal-bus` 0.1 → 0.3, and the silent display-killing bug
   fixes (timg `Timer::start()`, SPI 80 MHz → 20 MHz, etc.).

2. `[device] Replace timg Timer with esp_hal::time::Instant` — drops
   `timg::Timer` entirely (`DevicePeripherals` / `Resources` /
   `FrostyUi` / `SerialInterface` / factory paths) in favour of
   `esp_hal::time::Instant::now()`, which is backed by `SYSTIMER` and
   initialised by `esp_hal::init`. Removes the `T: timer::Timer` generic
   from `SerialInterface`, the `Box::leak` on the leaked timer handle,
   and the easy-to-forget `Timer::start()` call that v0.22 used to do
   for us.

3. `[device] Re-enable hardware DS signing without esp-hal fork` —
   reinstates a hand-rolled DS register driver against v1's API and
   adds a v1-specific clock-enable + reset step via the public PAC
   (wrapped in `critical_section::with(...)`). Because the DS driver
   lives in our `device` crate and only uses the public (unstable) PAC
   accessors that upstream esp-hal exposes, we don't need a fork to do
   this.

## Why this branch exists / how it relates to Adam's prior work

Two earlier branches were attempted by Adam and pushed to origin:

- [`hal-1.0`](https://github.com/frostsnap/frostsnap/tree/hal-1.0) — top
  commit `8f8c828e Skips fades so UI displays`. Booted, but the display
  never produced visible pixels.
- [`hal-1.0-dev`](https://github.com/frostsnap/frostsnap/tree/hal-1.0-dev)
  — top commit `5caee0da Boots but no display`. Same status.

**Both branches still depended on a frostsnap esp-hal fork**, just a
different (newer) one than the v0.22 fork — they pin
`https://github.com/frostsnap/esp-hal` at rev `55d62ee5`, which carries
a substantial in-tree DS driver patch (`esp-hal/src/ds.rs`, ~374 lines)
plus an `esp-ds` certificate utility crate and example, ~1465 added
lines total. Adam's approach was "fork esp-hal and add a proper DS
driver inside it." This branch instead keeps the DS driver in our
`device` crate and gets to upstream esp-hal v1.0.0 from crates.io with
zero patches.

A few decisions are inherited verbatim from `hal-1.0` (the SPI 20 MHz
frequency, the per-bin stack-allocated `FlashStorage`, the
`esp-storage = "0.6"` pin), so credit where due.

## On the v0.22 esp-hal fork we're moving away from

For context: the previous workspace was on the
[frostsnap/esp-hal fork](https://github.com/frostsnap/esp-hal) at rev
`6ecaa2eb` — a separate fork from the one Adam used. That older fork
was **exactly 19 lines** on top of upstream esp-hal v0.22 — a single
commit titled `Add DS peripheral clock reset and enable` which:

1. added a `Peripheral::Ds` arm to esp-hal's internal
   `PeripheralClockControl::enable` / `::reset` so it could write the
   `crypto_ds_clk_en` / `crypto_ds_rst` bits in `perip_clk_en1` /
   `perip_rst_en1`, and
2. made `Hmac::new` call `PeripheralClockControl::reset(Ds);
   PeripheralClockControl::enable(Ds);` after enabling its own HMAC
   clock.

That was the v0.22 fork's entire purpose: there was no public esp-hal
API to enable the DS clock, so the fork monkey-patched HMAC to enable
it as a side-effect. That's why master's `device/src/ds.rs` "just
worked" — it was poking DS registers immediately after `Hmac::new` had
silently enabled the clock for it.

In v1 the equivalent register fields (`crypto_ds_clk_en`,
`crypto_ds_rst`) are reachable from outside esp-hal via
`peripherals::SYSTEM::regs()` (gated `#[instability::unstable]`, which
we already enable), so we can do the enable+pulse-reset directly in
`HardwareDs::new` without forking anything. Same workaround the v0.22
fork did internally, just from outside esp-hal.

## Notable carryover decisions

- **`esp-storage` held back at 0.6** rather than tracking 0.8 (matching
  Adam's branches). 0.7 only shipped ESP32-only fixes; 0.8 added a
  `Flash<'d>` peripheral parameter to `FlashStorage::new` (and a `Drop`
  impl), which would force the bins to plumb the FLASH singleton
  through `DevicePeripherals` or to `unsafe { FLASH::steal() }`.
  Neither buys us anything we want, so we stay on 0.6 and the bins keep
  `let flash = RefCell::new(FlashStorage::new());`.
- **`esp-hal/unstable` is enabled** in the workspace `esp-hal` features.
  Everything we touch (`SYSTEM::regs()`, `DS::register_block()`, the
  PAC bit-fields, etc.) lives behind that gate today.

## Test plan

Build-time:
- [x] `just check-device` clean
- [x] `just build-firmware frontier` builds
- [x] `just build-firmware legacy` builds
- [x] `just build-firmware widget_dev` builds

Hardware (still pending — needs a production-provisioned device):
- [ ] `just env=prod build-firmware-signed frontier && just env=prod flash-firmware`
- [x] `just genuine-check` succeeds (exercises the DS challenge/response path end-to-end)
- [x] Display still draws on real hardware (the three silent display-killing fixes from commit 1)
- [x] UART up/downstream still talk after the `SerialInterface` generic removal
- [x] Touch input still works after the `cst816s` v1 port

DS signing **cannot be unit-tested** — `just genuine-check` against a
real provisioned device is the only way to verify it.

## Stacked on

This PR is stacked on #452 (`dev-factory`). Should be reviewed/merged
after that one.


device/src/peripherals.rs
```diff
@@ -32,36 +30,48 @@ use crate::efuse::EfuseController;
...
+        use mipidsi::{interface::SpiInterface, models::ST7789, options::ColorInversion};
+
+        // NOTE: The old v0.22 code asked for 80 MHz here and it worked. In v1.0
+        // the SPI driver picks a different divider and the resulting clock
+        // exceeds the ST7789's maximum write speed (~62 MHz), so nothing shows
+        // up. 20 MHz is what we settled on after matching the working probe.
+        let spi = Spi::new(
+            $peripherals.SPI2,
+            SpiConfig::default()
+                .with_frequency(Rate::from_mhz(20))
```
musdom, line 51: I tried going back to 80MHz here and it still works. I also tested the screen fill performance:

- 20MHz: 62ms
- 61MHz: 35ms
- 62MHz: 21ms
- 80MHz: 21ms

LLFourn:
## Stack-usage work (the commits after the port itself)

The port initially failed CI's stack analyzer (largest frame must be ≤ 25% of available stack) for two compounding reasons:

1. **Less stack available.** esp-hal 1.x's linker defaults place switch/lookup tables and some constants in DRAM (`place-switch-tables-in-ram = true`; the old fork's layout kept them in flash). Together with generally larger statics, available stack dropped **32316 → 29864 bytes**, which pushed frames the port never touched over the percentage gate.
2. **Some frames genuinely grew.** `Resources::init_production` (bigger display/UI types + `Box::new(Self{..})` building the whole struct on the stack) and `Partitions::load` (the esp-bootloader-esp-idf parser reads the whole partition table through a ~3KB stack buffer).

Rather than trade RAM placement back, the follow-up commits **shrink the frames to fit the new constraints** (keeping the esp-hal defaults, so switch tables stay in DRAM):

- `shrink-init-production-frame` — box `Resources.ui` and outline `FrostyUi::new` / `Partitions::load` so the init functions don't carry everything at once: `init_production` 9504 → 3936, `init_dev` 13072 → 5248.
- `recover-stack-budget` — box the partition-table read buffer: `Partitions::load` 7552 → 4480.
- `device-keygen-module` — split `recv_coordinator_message`'s keygen arms into per-variant `#[inline(never)]` handlers (a single frame otherwise holds the union of every arm's inlined ECC machinery), and move the keygen domain into `device/keygen.rs` following the `device/restoration.rs` / #458 module convention: keygen path 8896 → 5552. Pure code motion — token-identical modulo mechanical path fixes.
- `shrink-numeric-keyboard-frame` — box each keypad row at construction instead of building all four ~2KB rows as locals: `NumericKeyboard::new` 7872 → 4768.
- CI tweak: `just stack-check` analyzes the frontier binary, so the legacy matrix leg was just checking frontier twice; the step now runs once.

### Before the port (master, fork esp-hal)

```
Available stack:  32316 bytes
    1    8912   27.6%  FrostSigner::recv_coordinator_message
    2    7872   24.4%  NumericKeyboard::new
    3    6784   21.0%  WidgetTree::draw
    4    6464   20.0%  DeviceLoop::poll
    5    6336   19.6%  Resources::init_production
    6    6160   19.1%  FrostyUi::set_workflow
   10    4624   14.3%  Partitions::load
```

### After the port, before the fixes

```
Available stack:  29760 bytes
    1    9504   31.9%  Resources::init_production      <- FAIL (>25%)
    2    8896   29.9%  FrostSigner::recv_coordinator_message
    3    7872   26.5%  NumericKeyboard::new
    4    7552   25.4%  Partitions::load
```

### Now (PR head)

```
Available stack:  29864 bytes (CI)
    1    6800   22.8%  WidgetTree::draw                <- largest; PASS
    2    6240   21.0%  FrostyUi::set_workflow
    3    5680   19.1%  EraseDevice::new
    4    5552   18.7%  FrostSigner keygen path
    5    5536   18.6%  FeePage::new
    9    4768   16.0%  NumericKeyboard::new
         4480          Partitions::load
         3936          Resources::init_production
```

Net: the worst frame went from 31.9% (or 27.6% even before the port, on the old layout) to **22.8%**, with the top of the profile much flatter — and the firmware keeps esp-hal's performance-preferred RAM placement.

musdom (approved):

musdom:
Rebased on master

musdom:
Rebased on master and upgraded esp-hal to v1.2.0

PR #462, opened 2026-04-16
[app] Refactor FullscreenActionDialogController around a monotonic device set

LLFourn:
The controller now seeds its action-needed set at construction and only ever shrinks it, with a single `shouldShow` invariant driving visibility and the dialog popping itself from its own Navigator scope. This removes a race where caller-side `Navigator.pop` fought with `_reconcile` (hit on wallet_receive's verify flow) and lets callers drop the lazy add-then-show dance.


LLFourn:
@aphanley can you give this another reivew?

aphanley (approved):
All working during testing

frostsnapp/lib/wallet_create.dart
```diff
@@ -360,15 +368,18 @@ class WalletCreateController extends ChangeNotifier {
               network: form.network,
             )
             .toBehaviorSubject();
-        for (final id in selectedDevices) {
-          _keygenController.addActionNeeded(context, id);
-        }
+        _keygenController?.dispose();
+        final keygenController = _buildKeygenController(
```
LLFourn: I don't like whatever is happening here. Why do we build a new keygen controller after we just disposed of one.

PR #468, opened 2026-04-22
[app] Backup Reminder UX

aphanley:
Moves the backup reminder from a post-creation popup into the moments that matter — when the user opens Receive, and via a persistent banner.             
                                                                                                                                                            
  - Tapping Receive on a wallet with incomplete backups shows a reminder dialog. User can dismiss to proceed, open the backup checklist, or cancel.         
  - "Unfinished backups" banner is amber by default, turns red once the wallet has funds.
  - Tapping the banner opens the backup checklist directly.                                                                                                 
  - No more auto-popup after wallet creation.                                                                                                               
  - On the backup checklist, each share has a circular check icon — outlined when incomplete. Tapping opens a confirmation dialog to mark that share as backed up (for users who already recorded their backup outside the app).

### DIFF
--- a/frostsnapp/lib/backup_workflow.dart
+++ b/frostsnapp/lib/backup_workflow.dart
@@ -11,6 +11,18 @@ import 'package:frostsnap/src/rust/api/device_list.dart';
 import 'package:frostsnap/theme.dart';
 import 'package:glowy_borders/glowy_borders.dart';
 
+extension BackupDeviceX on BackupDevice {
+  /// A share counts as backed up when it has been explicitly marked
+  /// complete, OR when there is no tracking record at all (legacy wallets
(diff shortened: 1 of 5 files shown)


aphanley:
Replaces #440

LLFourn (approved):
Added e5b09477a78ae2eb51ddb6e4b2265e134a8b2bfc

- Centralize the "is backup complete" rule on BackupDevice/BackupRun
  extensions, replacing five inline `complete != false` / `== false`
  checks and the private isBackupDone helper with one documented rule.
- Drop the leftover 24-px leading placeholder from backup-checklist
  rows now that the completion indicator lives in the trailing slot.
- Simplify the "Check Backup on Device" dialog body to a single line
  of secondary guidance, removing the redundant key/wallet info card.

PR #469, opened 2026-04-22
TX Details UX

aphanley:
Needed to merge in tx-details-ux work


LLFourn:
Why is this on  the fullscreen-dialog-refactor branch? @aphanley can you fix that.

PR #470, opened 2026-04-22
[app] Show receive address and share index in tx details

aphanley:
Incoming txs now display the receive address they landed on (with its derivation index) in the details column. The signing progress list prefixes each device with its short share index (e.g. "#1 Alice").


aphanley:
@LLFourn here, sorry it merged into fullscreen dialog refactor after we had already merged that branch to master

LLFourn (approved):

PR #475, opened 2026-04-29
Update for chilldkg API changes from secp256kfun security review

LLFourn:
## Summary

Tracks the chilldkg API breakage on the [security-review omnibus branch](https://github.com/LLFourn/secp256kfun/pull/247) of secp256kfun. Patches in that branch via \`[patch.crates-io]\` and updates the call sites:

- \`certpedpop::Contributor::gen_keygen_input\` now takes \`n_contributors\` (= coord count + device count). Updated both coordinator and device call sites.
- \`Certifier::new\` is fallible (rejects \`agg_input.n_contributors() != contributor-set size\`); coordinator side propagates with \`.expect\` since the count was just enforced by \`Coordinator::new\`.
- \`verify_receive_share_and_certify\` now returns \`(SecretShareReceiver, Sig)\` instead of \`(PairedSecretShare, Sig)\`. The receiver holds the share internally until \`finalize\` runs with the full cert map. \`KeyGenPhase2\` stores the receiver (drops the redundant \`paired_secret_share\` and \`agg_input\` fields); the \`Check\` handler calls \`receiver.finalize\` to release the share, replacing the previous in-line \`Certifier::new\` + \`receive_certificate\` + \`finish\` dance.
- \`KeyGenPhase2::t_of_n()\` is gone; the only caller now derives it from the \`CertifiedKeygen\` returned by \`finalize\`.

## Test plan

- [x] \`cargo test -p frostsnap_core\` passes against the patched secp256kfun branch
- [ ] Smoke-test on real hardware before merge

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3985,9 +3985,9 @@ checksum = "ece8e78b2f38ec51c51f5d475df0a7187ba5111b2a28bdc761ee05b075d40a71"
 
 [[package]]
 name = "schnorr_fun"
-version = "0.12.0"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c711acb599464e57111c072bf6a9a64b51313f4baa776f4a500926dd4339bd32"
+checksum = "b92126e202853794199f04edf02851e17963c95e1d2308dccecf92a6857359d5"
 dependencies = [
  "bech32",
  "bincode",
@@ -4094,9 +4094,9 @@ dependencies = [
 
 [[package]]
 name = "secp256kfun"
-version = "0.12.0"
+version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "17d209307a50f8baee901c4e65ccef0da4185ef2d54834d73d6847369b8fa049"
+checksum = "ca6e7ba8c21335d29d0c9b8f46c3fbd1127b03b0e0b7dc5ebbe3ef28ef820ec6"
 dependencies = [
  "bincode",
  "digest 0.10.7",
--- a/frostsnap_core/src/coordinator.rs
+++ b/frostsnap_core/src/coordinator.rs
@@ -634,24 +634,20 @@ impl FrostCoordinator {
                         let cert_scheme = certpedpop::vrf_cert::VrfCertScheme::<Sha256>::new(
                             crate::message::keygen::VRF_CERT_SCHEME_ID,
                         );
-                        let share_index = state.device_to_share_index.get(&from).ok_or(
-                            Error::coordinator_invalid_message(
+                        let receiver_idx = state
+                            .devices_in_order
+                            .iter()
+                            .position(|d| d == &from)
+                            .ok_or(Error::coordinator_invalid_message(
                                 message_kind,
                                 "got share from device that was not part of keygen",
-                            ),
-                        )?;
-
-                        // Input-generator indices: [0..n_coordinators) are coordinators;
-                        // [n_coordinators..n_coordinators + n_devices) are devices in
-                        // share-index order. Share indices are 1-based, hence the `- 1`.
-                        let n_coordinators = state.coordinator_public_keys.len() as u32;
-                        let input_gen_index = u32::from(*share_index) - 1 + n_coordinators;
+                            ))? as u32;
 
                         state
                             .input_aggregator
                             .add_input(
-                                &schnorr_fun::new_with_deterministic_nonces::<Sha256>(),
-                                input_gen_index,
+                                &schnorr_fun::Schnorr::<Sha256>::verify_only(),
+                                certpedpop::Party::Receiver(receiver_idx),
                                 *response.input,
                             )
                             .map_err(|e| Error::coordinator_invalid_message(message_kind, e))?;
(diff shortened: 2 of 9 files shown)


frostsnap_core/src/message/keygen.rs
```diff
@@ -27,6 +27,9 @@ impl From<Keygen> for CoordinatorToDeviceMessage {
     }
 }
 
+/// Slot ordering follows the order of `devices`; the receiver-contributor at
+/// slot `i` has `ShareIndex = i + 1`. Distinctness of `devices` and
+/// `coordinator_public_keys` is enforced by `certpedpop::Contributor::gen_keygen_input`.
```
LLFourn: This comment is unhelpful. Who cares about slot ordering?

frostsnap_core/src/coordinator.rs
```diff
@@ -641,11 +641,10 @@ impl FrostCoordinator {
...
 
-                        // Input-generator indices: [0..n_coordinators) are coordinators;
-                        // [n_coordinators..n_coordinators + n_devices) are devices in
-                        // share-index order. Share indices are 1-based, hence the `- 1`.
-                        let n_coordinators = state.coordinator_public_keys.len() as u32;
-                        let input_gen_index = u32::from(*share_index) - 1 + n_coordinators;
+                        // Input-generator indices: [0..n_devices) are devices in
+                        // share-index order; [n_devices..) are coordinators (aux).
+                        // Share indices are 1-based, hence the `- 1`.
+                        let input_gen_index = u32::from(*share_index) - 1;
```
LLFourn: This comment is referencing internal details of chilldkg that aren't relevant.

frostsnap_core/src/coordinator.rs
```diff
@@ -663,28 +661,28 @@ impl FrostCoordinator {
...
-                                cert_scheme,
-                                agg_input.clone(),
-                                &state.coordinator_public_keys,
-                            );
-
+                            let mut certifier = certpedpop::Certifier::new(cert_scheme, verified);
                             certifier
                                 .receive_certificate(state.my_keypair.public_key(), sig)
-                                .expect("will be able to verify our own certificate");
+                                .expect("self certificate must verify");
```
LLFourn: comment churn

frostsnap_core/src/device.rs
```diff
@@ -382,45 +375,34 @@ impl<S: NonceStreamSlot + core::fmt::Debug> FrostSigner<S> {
             }
             KeyGen(keygen_msg) => match keygen_msg {
                 self::Keygen::Begin(begin) => {
-                    let device_to_share_index = begin.device_to_share_index();
-                    if !device_to_share_index.contains_key(&self.device_id()) {
+                    let Some(my_slot) = begin.devices.iter().position(|d| d == &self.device_id())
```
LLFourn: before we used to have a proper error here to indicate that this device was not part of keygen. We should keep that.

frostsnap_core/src/device.rs
```diff
@@ -452,15 +434,18 @@ impl<S: NonceStreamSlot + core::fmt::Debug> FrostSigner<S> {
...
+                        .map_err(|e| {
+                            Error::signer_invalid_message(
+                                &message,
+                                format!("Failed to verify and receive share: {e}"),
+                            )
+                        })?;
 
-                    //XXX: We check the fingerprint so that a (mildly) malicious
-                    // coordinator cannot create key generations without the
-                    // fingerprint.
```
LLFourn, line 462: This comment should not have been removed. It was explaining why we do something.

frostsnap_core/src/device.rs
```diff
@@ -142,8 +141,9 @@ impl EncryptedSecretShare {
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct KeyGenPhase1 {
-    pub device_to_share_index: BTreeMap<DeviceId, NonZeroU32>,
-    pub input_state: certpedpop::Contributor,
+    /// Devices in slot order. The device at position `i` has `ShareIndex = i + 1`.
```
LLFourn: "slot order" is not a concept that exists in frostsnap. I suggest removing this comment.

Cargo.lock
```diff
@@ -3909,7 +3909,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
```
LLFourn: I wonder why this had to be downgraded

nickfarrow (approved):
Looks great, haven't tested yet

LLFourn:
Changes since the review:

- Removed dead `coordinator_public_keys` fields from `KeyGenPhase1`, `KeyGenPhase2`, and `KeyGenWaitingForResponses` — chilldkg's `Contributor` already retains them.
- Removed `Phase1.devices` and `Phase2.devices` on the device side — chilldkg's `Contributor`/`SecretShareReceiver` already holds the receiver keys (which IS the DeviceId payload).
- Added `verified_agg_input.n_receivers() == devices_in_order.len()` assertion above the per-element `assert_eq!` at the keygen-finalize zip — silent `zip` truncation on length mismatch now panics loudly.
- `KeyGenPhase3` widened/narrowed dance fixed: `n_receivers: u16` and `t_of_n() -> (u16, u16)`, with a checked `u16::try_from(...).expect(...)` at the single chilldkg boundary; dropped the silent `as u16` cast in `frosty_ui.rs`.
- Replaced `.expect("coordinator inputs are validated by start_keygen")` on `gen_keygen_input` with a returned `ActionError::StateInconsistent`.
- Dropped the noop `let contributer = *state.contributer;` indirection — `Box` auto-derefs for self-consuming methods.
- Switched `input_aggregator.add_input(...)` to use `Schnorr::<Sha256>::verify_only()` instead of `new_with_deterministic_nonces` — `add_input` only verifies a PoP, no nonce gen needed.
- Amended commit message to drop the stale `[patch.crates-io]` bullet (the published 0.13.0 already includes the chilldkg-api-tightening commits).
- Cleaned `Cargo.lock` to remove the incidental `windows-sys 0.59 → 0.52` downgrade by replaying with a targeted `cargo update -p schnorr_fun -p secp256kfun`; lockfile diff is now exactly the two crate bumps.
- Rebased on `origin/master` to pick up the merged "drop redundant access_structure_ref" change.

aphanley:
Manually tested, looks good to me

PR #477, opened 2026-05-11
[app] Prompt for screen lock instead of crashing when Android Keystore is unbound

LLFourn:
## Summary
- Android Keystore key is created with `setUserAuthenticationRequired(true)`. On phones without a screen lock the key can't be created, and the resulting opaque `PlatformException` bubbles unhandled out of every signing / backup / recovery flow — the app effectively breaks for that user.
- Detect the no-lock-screen case proactively in `SecureKeyManager.kt` and surface a typed `NO_LOCK_SCREEN` error.
- `SecureKeyProvider.getEncryptionKey()` catches it once globally and shows a blocking dialog (Cancel / Try again / Open Settings) that retries on the user's gesture. Open Settings deep-links to `Settings.ACTION_SECURITY_SETTINGS`.
- Recovery-mode listener in `main.dart` narrows its catch by error code, logs unexpected codes at error level, and surfaces a snackbar so a user-cancelled prompt doesn't fail silently.

## Test plan
- [x] Install on an Android device **without** a screen lock; trigger any flow that calls `getEncryptionKey()` (e.g. signing, backup) and confirm the "Screen Lock Required" dialog appears.
- [x] Tap **Open Settings** — confirms it deep-links to the Security settings page; dialog stays open.
- [x] Set a PIN, return to app, tap **Try again** — flow completes.
- [x] Tap **Cancel** — confirms a snackbar appears (recovery-mode path) or the call site's existing error handling fires (other paths) instead of an unhandled exception.
- [x] Plug in a Frostsnap recovery-mode device at app launch with no PIN set — dialog still appears (not silently swallowed pre-`runApp`).
- [x] Install on an Android device **with** a screen lock — happy path still works, no dialog.

### DIFF
--- a/frostsnapp/android/app/src/main/kotlin/com/frostsnap/SecureKeyManager.kt
+++ b/frostsnapp/android/app/src/main/kotlin/com/frostsnap/SecureKeyManager.kt
@@ -5,6 +5,7 @@ import android.app.KeyguardManager
 import android.content.Context
 import android.content.Intent
 import android.os.Build
+import android.provider.Settings
 import android.security.keystore.KeyGenParameterSpec
 import android.security.keystore.KeyProperties
 import android.security.keystore.UserNotAuthenticatedException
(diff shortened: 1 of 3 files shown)


musdom (approved):

musdom:
Tested ACK. Pixel 7 Pro GrapheneOS with no screen lock. Tested wallet recovery, sign message and check backup flows.

aphanley:
Yep, worked well for me as well on Pixel 6a. There is an issue where if the screen lock key has been deleted by the user (they removed their pin or swipe pattern, and then set a new one) then we can't access the key and everything silently fails (and the user is probably going to forget they did this and not associate it with not being able to carry out signing or backup actions for their existing wallet) I think we should add a notification to the user if this happens, but we don't have to do it now, it's pretty niche.

Just changing pin or switching from pin to swipe pattern is unaffected

aphanley:
force pushed no edit to add signature since commit was unsigned

PR #478, opened 2026-05-12
[just] Fix erase-device by writing 0xFF instead of using flash stub

LLFourn:
## Summary
- `espflash erase-parts` requires the RAM flash stub, which times out on our devices (secure-boot enabled). Replace with `espflash write-bin --no-stub` writing 0xFF across the partition, matching how `flash-firmware` already wipes `otadata`.
- DRY up the shared path: `flash-firmware` now delegates the `otadata` wipe to `erase-device`, and `device/blank-otadata.bin` is gone.
- Look up partition addresses/sizes by feeding `device/partitions.bin` through `espflash partition-table --to-csv`, so we get normalized hex offsets/sizes — no more parsing `16K`/`512K` unit suffixes by hand.

## Test plan
- [ ] `just erase` (default `nvs`) wipes the nvs partition without timing out
- [ ] `just erase otadata` works
- [ ] `just flash` still flashes firmware after wiping otadata
- [ ] `just flash --table` prints the expected addresses/components
- [ ] Sanity-check on Linux (the `LC_ALL=C tr` is portable, but worth confirming)


aphanley (approved):
Fixes the issue for me

aphanley:
force pushed no edit since commit was unsigned

PR #482, opened 2026-05-15
[app,coord] Persist device name at keygen/restoration commit boundary

LLFourn:
## The bug

Devices in a signer list can render as `<unknown>` even though the user named them — most visibly in the Send-tx review dialog. Bundled side fix: in the inline keygen name field, typing a space after a few characters silently dropped the space.

## Why it matters

The coord only persists a device's name when the device sends `SetName` upstream. That message is emitted from the device's commit handlers (`save_pending_device_name` at keygen finalize and at backup save/consolidate). If the cable is yanked in the in-flight window — common enough during test cycles — the coord never sees the name. Every UI surface that reads `coord.get_device_name(id)` then shows `<unknown>` until the device is replugged. Devices that don't come back stay `<unknown>` forever.

The SPC bug had the same shape on the input side: typing-time trim plus a rebuild-driven controller resync were fighting each other and the user's space keystroke lost.

## The fix

Persist the user-typed name on the coord side at the same Dart→Rust call that drives the device's commit — in `finalize_keygen` and the two `tell_device_to_*_physical_backup` wrappers — between the outbound message and the blocking wait. The eventual `SetName` is still received and processed; since both paths converge on the same string, the existing reactive handler short-circuits.

Names flow through verbatim. Dart trims at every Dart→Rust boundary (including the per-keystroke Preview); Rust trusts whatever it receives. Charset and length are enforced by the Dart `TextField`s (`maxLength` and `nameInputFormatter`). Keeping the trim on one side only avoids the "save 'Server ' on the device but 'Server' on the coord and then watch them fight" failure mode.

The SPC bug is fixed by deleting the rebuild force-sync of the keygen name field's controller — the per-keystroke trim was harmless on its own; the resync was the actual culprit.

## Test plan

- [ ] Keygen a 2-of-3 wallet, name each device, yank the cable right after pressing the keygen confirm. Send-tx dialog should show all three names without any replug.
- [ ] In the inline keygen name field, type "Server" then space then "R" — should appear as "Server R".
- [ ] Standard upgrade → keygen → send → receive → erase → restore loop ends with named signers from the start.

### DIFF
--- a/frostsnap_coordinator/src/keygen.rs
+++ b/frostsnap_coordinator/src/keygen.rs
@@ -75,6 +75,10 @@ impl KeyGen {
     pub fn keygen_id(&self) -> KeygenId {
         self.state.keygen_id
     }
+
+    pub fn devices(&self) -> &[DeviceId] {
+        &self.state.devices
+    }
 }
 
 impl UiProtocol for KeyGen {
--- a/frostsnapp/lib/device_setup.dart
+++ b/frostsnapp/lib/device_setup.dart
@@ -34,7 +34,7 @@ class DeviceNameFieldState extends State<DeviceNameField> {
     if (name != null) {
       _controller.text = name;
     }
-    coord.updateNamePreview(id: widget.id, name: _controller.text);
+    coord.updateNamePreview(id: widget.id, name: _controller.text.trim());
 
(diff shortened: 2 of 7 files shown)


aphanley (approved):
LGTM

PR #487, opened 2026-05-21
Fix release signing, env, lockfile, and ci firmware verification

nickfarrow:
@musdom hit `INSTALL_FAILED_UPDATE_INCOMPATIBLE` upgrading v0.2.0 → v0.3.0 of the direct APK. Investigation surfaced a few related bugs.

- **Direct APKs signed with ephemeral keys.** Empty `signingConfigs.debug {}` made CI auto-generate a new keystore per run. This PR loads a stable keystore (same as releases `v0.1.0` -- unsure about v0.2.x) into `ANDROID_KEYSTORE` secret.
- **APKs tagged `-modified` in About.** `flutter pub get` was rewriting `pubspec.lock` mid-build and dirtying the tree. Added `--enforce-lockfile`.
- **Releases bundled the dev genuine cert pubkey.** `release.yml` was building with `env=dev`. Pass `env: prod` through platform actions.
- **v0.3.0 firmware not in `KNOWN_FIRMWARE_VERSIONS`** went unnoticed at release time. Add `verify-firmware --require-known-version` + a release.yml job that gates platform builds.

Existing users on v0.2.x direct APKs may need to uninstall+reinstall once to move to the new stable signing identity.

Sadly we can't use the Google play store key here. Meaning that our APK users will have to uninstall before they can install from the playstore. Google only allows us to have an "upload key" and google then strips the signature and adds their own prior to releasing on playstore.


LLFourn (commented):
Solid PR — clear bug list, each item maps cleanly to one of the four fixes, and CI is already green on the new pin. A few thoughts, none blocking:

**1. The `signingConfigs.debug {}` block is now load-bearing but reads as a no-op.**

`frostsnapp/android/app/build.gradle:66-70` is an empty block that *also* happens to receive a stable production keystore via the CI step at `set-up-app-build`/`build-android`. Anyone reading the gradle file later (or trying to reproduce a release locally) won't see the connection — the comment just says "Standard debug keystore - automatically available." Two options that would make this less surprising:

- Add a comment in the gradle block explaining that the file at `~/.android/debug.keystore` is materialized from the `ANDROID_KEYSTORE` CI secret for release builds, and that local builds will produce APKs signed with the developer's local debug keystore (and therefore won't be upgrade-compatible with CI releases).
- Or, more invasive: rename the block to something like `directRelease`, set explicit `storeFile`/`keyAlias` properties driven by env vars, and have CI inject those — so the file is self-documenting.

Either is fine; the comment is the smaller intervention.

**2. The keystore-restore step is a silent fallback if `ANDROID_KEYSTORE` is missing.**

`if: env.ANDROID_KEYSTORE != ''` in `build-android/action.yml` is the right gate for non-release builds (where the secret is intentionally absent), but it means a release run with the secret accidentally unset would happily produce an APK signed with a fresh auto-generated keystore — exactly the bug this PR is fixing. Worth a hard fail in `release.yml` itself before invoking the build action, e.g.:

```yaml
- name: Require release signing key
  env:
    ANDROID_KEYSTORE: ${{ secrets.ANDROID_KEYSTORE }}
  run: test -n "$ANDROID_KEYSTORE" || { echo "ANDROID_KEYSTORE secret is required for releases"; exit 1; }
```

Costs nothing; closes the "next time the secret rotates and someone forgets to update it" loop.

**3. `verify-firmware --require-known-version` is the right shape.**

Gating the platform builds on `needs: ["verify-firmware"]` means a release whose firmware digest hasn't been registered in `KNOWN_FIRMWARE_VERSIONS` can't ship the platform builds at all. Nice. The slight wart is `cargo run -p frostsnap_factory --release` for a digest lookup — multi-minute build for a millisecond check. Not worth fixing now, but if the verify-firmware step starts dominating release time, a prebuilt `frostsnap_factory` binary cached across releases would pay back quickly.

**4. `pubspec.lock` downgrades look correct but are easy to misread.**

`characters` 1.4.1→1.4.0, `matcher` 0.12.19→0.12.17, etc. are a side effect of pinning Flutter back to 3.38.5 (dart 3.8.x) from whatever 3.9.x-era SDK was generating the lockfile locally. The lockfile is now consistent with what CI will reproduce — that's the whole point. Worth noting in the PR description in case a reviewer pattern-matches "downgrade = regression."

**5. Existing-user migration.**

The PR description already calls out that v0.2.x direct-APK users will need to uninstall+reinstall once. Two suggestions:
- Pin that in the release notes for whatever v0.3.1 / next-direct-APK release ships this; otherwise users will hit `INSTALL_FAILED_UPDATE_INCOMPATIBLE` again.
- If we want to keep the door open to *not* breaking signing identity again in the future, the next time the keystore rotates we'd want at least a one-release deprecation path. Hard to do retroactively, but worth a comment somewhere ("the keystore identity is stable; don't rotate without a migration plan").

LGTM modulo the above. The keystore-secret hard-fail (#2) feels like the only one I'd actually want to see before merge.

LLFourn (approved):

PR #491, opened 2026-05-26
Enforce cargokit lockfile

nickfarrow:
Cargokit's `pubspec.lock` was unenforced. Its ~50 transitive Dart deps run at build time.

- `set-up-app-build`: Run `dart pub get --enforce-lockfile` against the committed cargokit lockfile.
- Post-build: `git diff --exit-code` to catch anything that drifted during the build itself.

Same pattern as the existing `Cargo.lock not dirty` check in `test.yml`.

### DIFF
--- a/.github/actions/build-android/action.yml
+++ b/.github/actions/build-android/action.yml
@@ -53,6 +53,6 @@ runs:
     - name: Build appbundle
       shell: bash
       run: just env=${{ inputs.env }} build appbundle --release --flavor playstore
-    - name: Verify cargokit lockfile not drifted during build
+    - name: Verify no checked-out files drifted during build
       shell: bash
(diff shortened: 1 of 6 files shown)


.github/actions/build-android/action.yml
```diff
@@ -53,3 +53,6 @@ runs:
     - name: Build appbundle
       shell: bash
       run: just env=${{ inputs.env }} build appbundle --release --flavor playstore
+    - name: Verify cargokit lockfile not drifted during build
+      shell: bash
+      run: git diff --exit-code frostsnapp/rust_builder/cargokit/build_tool/pubspec.lock
```
LLFourn: This should probably just be `git diff --exit-code`. No checkout out files should change during build not just this one.

LLFourn (approved):
left one nit

PR #494, opened 2026-06-01
Verify firmware secure-boot key at build time for local desktop builds

LLFourn:
## Why

Local rebuilds of the desktop app silently embed whatever bytes happen to be at
`target/riscv32imc-unknown-none-elf/release/{FROSTSNAP_ENV}-frontier.bin` and call them "firmware".
Nothing checks that those bytes were signed by the secure-boot key the build is claiming to be for.

Preparing v0.3.0 I tripped over this for real. I had the prod-signed
`firmware.bin` from the release on disk and ran `just env=dev build macos --release` —
because `env=dev` is the default of the `just` recipe. The build happily produced an
`.app` that:

- **Embedded the prod-signed firmware bytes** (fine — those bytes are the right
  release firmware).
- **Embedded the dev genuine-cert key** (NOT fine — a dev-keyed coordinator
  cannot verify a real prod-signed device's genuine challenge response).
- **Disabled the runtime firmware-digest whitelist** (`option_env!("FROSTSNAP_ENV") == Some("prod")`
  is the gate, and we were not prod).

The build did not complain. The artifact looked normal. Only a human noticing
the `-modified` suffix in the version string caught it. I then re-built with
`env=prod` and re-uploaded.

CI is fine — `release.yml` already has a `verify-firmware` job that runs
`frostsnap_factory verify-firmware --require-known-version` against the release
firmware, and every platform `build-*` job lists `needs: ["verify-firmware"]`.
The gap is entirely in **local** rebuilds, which is exactly the surface that
produces locally-signed DMGs.

## What this PR does

Adds two checks to `frostsnapp/rust/build.rs` on the conventional
`BUNDLE_FIRMWARE=1` path:

1. **Signature self-consistency.** The bytes at `target/.../{env}-frontier.bin` parse as a
   valid Secure Boot v2 image and the embedded RSA-PSS signature verifies against the
   embedded modulus. Same check the factory CLI's `verify-firmware` runs.

2. **Key binding.** The embedded RSA modulus equals the expected env-specific
   secure-boot public key:
   - **prod**: read from a newly-committed
     `frostsnap_factory/bootloader/prod/secure-boot-public-key.pem` (SubjectPublicKeyInfo
     form). This is the single named source of truth for the prod key. The value was
     extracted once from the signature block of the existing signed prod bootloader,
     so the PEM and the bootloader binary agree on the modulus today.
   - **dev**: derived from the already-committed
     `frostsnap_factory/bootloader/dev/secure-boot-key.pem` via
     `public_key_from_private_pem`. Deliberately no separate dev pubkey file —
     single source of truth per env.

If either check fails the build panics with a clear error pointing at the
exact file on disk, including a hint that the FROSTSNAP_ENV likely doesn't
match the firmware.

Explicit-path `BUNDLE_FIRMWARE=<path>` (used by `just legacy-run` for unsigned
legacy firmware) deliberately skips both checks — the "I told you exactly where
to read" semantic is preserved so legacy support keeps working.

## Mechanical bits

- `secure_boot.rs` is extracted out of `frostsnap_factory` (which depends on
  `mysql`, `clap`, `tracing-subscriber` — none of which belong in a build script)
  into a new minimal crate `frostsnap_secure_boot` (deps: `rsa`, `sha2`, `crc`,
  `rand_core`).
- `sign_firmware` is made generic over `&mut R: RngCore + CryptoRng` so the
  crate doesn't need `rand` either. Callers pass their own RNG.
- The verify API is split:
  - `verify_firmware(bytes, &expected_pk) -> Result<(), VerifyError>` — the
    safer headline API, what `build.rs` actually calls.
  - `verify_and_extract_pk(bytes) -> Result<RsaPublicKey, VerifyError>` — the
    lower-level extract-only form, kept for bootstrap cases. Renamed from the
    misleading-sounding `verify_firmware` to make it explicit that the function
    on its own only proves self-consistency.
- `public_key_from_pem(pem)` added alongside the existing
  `public_key_from_private_pem` so `build.rs` can read PEMs without taking a
  direct `rsa` dep.

## Commit structure

Two commits, each a self-contained step:

1. `[secure_boot] Catch wrong-env firmware bundled into local desktop builds` —
   the extraction + verify API + build.rs key-binding, with the prod expected
   key bootstrapped from the signed prod bootloader at build time.
2. `[secure_boot] Commit prod secure-boot pubkey, drop bootloader-bootstrap` —
   commit the PEM as the explicit source of truth, switch build.rs to read it
   directly via `public_key_from_pem`, drop the bootloader-bootstrap path.

## Test plan

Build matrix verified locally:

| FROSTSNAP_ENV | firmware bytes | result |
|---|---|---|
| `prod` | prod-signed (v0.3.0 release firmware) | builds |
| `prod` | dev-signed | panics: "embedded RSA key is not the expected key" |
| `dev` | dev-signed | builds |
| `dev` | prod-signed (the v0.3.0 mistake) | panics: "embedded RSA key is not the expected key" |
| any  | `BUNDLE_FIRMWARE=<path>` (explicit form) | check skipped (legacy-run) |

Unit tests:
- [x] `cargo test -p frostsnap_secure_boot` — 5 pass (4 existing + new `public_key_pem_roundtrip` covering the PEM helpers)
- [x] `cargo check -p frostsnap_factory` — clean after `secure_boot` extraction
- [x] `cargo check -p rust_lib_frostsnapp` — clean
- [x] Full `just env=prod build macos --release` (the original mac signing pipeline) — produces a working signed `.app`


LLFourn:
Superseded by #498, which unifies this build-time check with #495's CLI categorisation into a single source of truth (one `secure-boot-pubkey.pem` per env consulted by both build.rs and the CLI). Closing.

PR #496, opened 2026-06-05
[embedded] fix nonce index bug in AbSlot crash recovery

nickfarrow:
`AbSlot` stores data across two flash slots (A and B), tagging each write with an ever-increasing index and always reading back the slot with the **higher** index. After an interrupted write the two slots can sit at *different* indexes, and the recovery logic that decides the next index was buggy.

At present, `current_slot_and_index` read slot **A** twice instead of A and B, so it never saw B's index.

**Example** -- slots `A=3, B=4` (B is newer):
- Buggy: only looks at A → thinks newest is `3` → next write uses `4`, **reusing the index B already has**.
- Fixed: looks at both → newest is `4` → next write uses `5`.

It also returned "empty" if either slot was mid-rewrite (each slot is erased before being rewritten), causing the generation index to regress to 0. Both are fixed by reading both slots and treating an empty slot as older than any written one.

**Why this never leads to nonce reuse:** the nonce store on top of `AbSlot` only releases signing shares *after* persisting the new index and reading it back to confirm. A failed write that triggers this bug never gets that confirmation, so shares for an unpersisted nonce were never handed out.

Added a failed-write regression test (both slot orientations) and a crash-injection property test asserting the committed index never goes backwards across power loss at any point of a write.

### DIFF
--- a/frostsnap_embedded/src/ab_write.rs
+++ b/frostsnap_embedded/src/ab_write.rs
@@ -37,7 +37,7 @@ impl<'a, S: NorFlash> AbSlot<'a, S> {
         T: bincode::Encode,
     {
         let (next_slot, next_index) = match self.current_slot_and_index() {
-            Some((current_slot, current_index, _)) => {
+            Some((current_slot, current_index)) => {
                 let next_slot = (current_slot + 1) % 2;
                 let next_index = current_index + 1;
                 if next_index == u32::MAX {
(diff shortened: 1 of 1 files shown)


LLFourn (commented):
_Superseded by the approval below._

LLFourn (approved):
The core index-reuse fix is correct: collapsing the read and write paths onto a single `current_slot_and_index()` removes the duplicated slot-selection that let them disagree — the root cause of index reuse. #497's inlining of the branch is behaviourally identical (checked across the empty/equal/ordered cases) and reads better. The three unit tests directly exercise the recovery branches (torn write and empty slot, both orientations) and pass; dropping the proptest is fine — it over-claimed its crash coverage and the targeted tests cover the logic that was actually buggy.

One follow-up, not blocking this PR: `AbSlot` is not robust over single-word NOR writes — a torn write can leave a valid-looking index over an incomplete payload, and `read()` trusts it with no fallback. Contained for nonces today (index read independently so monotonicity survives; `device_nonces.rs:298` read-back-confirms before releasing shares, so no reuse), but worth a design pass. Tracked in #499.

PR #497, opened 2026-06-09
Tighten AbSlot cleanup: inline if-branch, drop proptest

LLFourn:
Two small follow-ups on #496.

## Inline the if-branch

`current_slot_and_index` was computing the slot first and then deriving the index via `a.max(b)?`. Returning directly from each branch lets `?` do the work — same behavior, one less intermediate to follow:

```rust
if b_index > a_index {
    Some((1, b_index?))
} else {
    Some((0, a_index?))
}
```

## Drop the crash-budget proptest

I think the proptest is overdoing it for this primitive, even though the consequences of a bug here are real (nonce reuse → leaked share — definitely not casual).

- The recovery logic has six total cases (`{None,Some} × {None,Some}`, plus the two index-orderings within `Some×Some`). The two surgical unit tests already cover both buggy orientations directly — I confirmed by reinjecting the original two-line bug into the new fix and watching both unit tests fail (and the proptest fail too, but redundantly).
- What the proptest additionally exercises is durability across mid-write crashes — but that property rides on `Slot::write` being effectively atomic for small payloads (one buffered `flash.write` call), which is a property of the layer below, not of the recovery logic this PR is fixing. So we're testing two different invariants in one test and the framing blurs what the suite is actually asserting.
- For a state machine this small, exhaustive case enumeration in unit tests reads more honestly than randomized crash budgets that may or may not hit every interesting point.

`CrashFlash` itself is a nice primitive and would earn its keep the moment we add real crash-tolerance branching elsewhere — but I'd rather bring it back from git history then than carry it as unused infrastructure now.

## Test plan
- [x] `cargo test -p frostsnap_embedded --lib ab_write` passes (3 tests)
- [x] Reinjected the original `slots[0].read_index()?` twice / `?` short-circuit bug into the fixed code on this branch → both remaining unit tests fail at both orientations

### DIFF
--- a/frostsnap_embedded/src/ab_write.rs
+++ b/frostsnap_embedded/src/ab_write.rs
@@ -70,11 +70,14 @@ impl<'a, S: NorFlash> AbSlot<'a, S> {
     fn current_slot_and_index(&self) -> Option<(usize, u32)> {
         let a_index = self.slots[0].read_index();
         let b_index = self.slots[1].read_index();
-        // An empty slot (`None`) must rank below any written one, otherwise a
-        // crash that leaves one slot mid-erase would make us forget the index
-        // held in the other and reuse it.
-        let current_slot = if b_index > a_index { 1 } else { 0 };
-        Some((current_slot, a_index.max(b_index)?))
+        // `Option` ranks `None` below every `Some`, so an empty slot is older
+        // than any written one — otherwise a crash mid-erase would make us
+        // forget the index held in the surviving slot and reuse it.
+        if b_index > a_index {
+            Some((1, b_index?))
+        } else {
+            Some((0, a_index?))
+        }
     }
 }
 
@@ -134,8 +137,6 @@ mod test {
     use super::*;
     use crate::test::TestNorFlash;
     use core::cell::RefCell;
-    use embedded_storage::nor_flash::{self, ReadNorFlash};
-    use proptest::{collection, prelude::*};
 
     /// The highest index actually committed to flash, or `None` if both slots
     /// are empty. `Option<u32>` orders `None` below every `Some`, so this is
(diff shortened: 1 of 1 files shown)


frostsnap_embedded/src/ab_write.rs
```diff
@@ -70,11 +70,14 @@ impl<'a, S: NorFlash> AbSlot<'a, S> {
...
-        // An empty slot (`None`) must rank below any written one, otherwise a
-        // crash that leaves one slot mid-erase would make us forget the index
-        // held in the other and reuse it.
-        let current_slot = if b_index > a_index { 1 } else { 0 };
-        Some((current_slot, a_index.max(b_index)?))
+        // `Option` ranks `None` below every `Some`, so an empty slot is older
+        // than any written one — otherwise a crash mid-erase would make us
+        // forget the index held in the surviving slot and reuse it.
+        if b_index > a_index {
+            Some((1, b_index?))
```
LLFourn, line 77: `b_index?` here can never short-circuit: `b_index > a_index` is only true when `b_index` is `Some` (`Some > None`, or `Some(y) > Some(x)`), so the `?` in this arm is dead — only the `else` arm (`a_index?`, the `None`×`None` case) can actually yield `None`.

The symmetric `b_index?` / `a_index?` reads cleanly, so I would lean toward keeping it as-is. Flagging only because it momentarily implies `b_index` could be `None` here when it cannot. Non-blocking — taste call.

LLFourn (commented):
Reviewed the inline refactor and the proptest removal.

- The inlined `current_slot_and_index` is behavior-preserving — checked all
  six `(a_index, b_index)` cases against the previous `current_slot` +
  `a_index.max(b_index)?` form.
- Ran `cargo test -p frostsnap_embedded --lib ab_write` against the head sha:
  3 tests pass.
- Agree with dropping the crash-budget proptest. Recovery is a pure function
  of the two `Option<u32>` slot indexes, and the two parametrized unit tests
  enumerate both orientations of every distinct post-crash state, so the
  removal doesn't drop coverage of the recovery logic this PR is about.

One non-blocking note inline on the inlined branch.

PR #495, opened 2026-06-02
Improve verify-firmware CLI functionality

nickfarrow:
`verify-firmware` verified each image's signature against the public key embedded in that same image. So any self-signed file would pass with exit zero.

To make this tool more useful for release management and firmware visibility, this PR compares the image's key digest against Frostsnap's pinned keys in `secure_boot_key/`

- PROD / DEV key → `✅ Verified`, exit 0
- unknown key → `⚠️ SIGNED BY UNKNOWN KEY`, exit 1
- not signed / corrupt → `⚠️ NOT SIGNED` / `⚠️ CORRUPT`, exit 1

PR #498, opened 2026-06-09
Verify secure-boot signer at build time and in verify-firmware CLI (supersedes #494, #495)

LLFourn:
Unified follow-up that supersedes #494 (LLFourn — build-time check) and #495 (nickfarrow — CLI categorisation). Both targeted the same gap (verifying *who* signed a Secure Boot v2 image, not just \"somebody\") but landed in different surfaces, with duplicated key files and incompatibly reshaped APIs. This collapses them to a single source of truth.

## Two user-visible fixes (one PR each before, both here now)

1. **Build-time** (#494's fix): `frostsnapp/rust/build.rs` now verifies the firmware bytes at \`target/.../{env}-frontier.bin\` are signed by the env's expected secure-boot key. Reproduces the v0.3.0 incident — \`just env=dev build macos --release\` with a prod-signed firmware on disk — as a build-time panic instead of a silent broken \`.app\`.

2. **CLI** (#495's fix): \`frostsnap_factory verify-firmware\` now categorises into four named outcomes with proper exit codes:
   - \`✅ Verified — signed by Frostsnap PROD key\` / \`DEV key\` (exit 0)
   - \`⚠️  SIGNED BY UNKNOWN KEY\` (exit 1)
   - \`⚠️  NOT SIGNED\` / \`⚠️  SIGNATURE INVALID / CORRUPT\` (exit 1)

   Malformed firmware bodies are surfaced best-effort so a corrupt body can't hide the signer verdict.

## Single source of truth per env

\`frostsnap_factory/bootloader/{env}/secure-boot-key.pem\` is the one file build.rs *and* the CLI consult. Dev's was already there (private key); prod's is added by this PR (SubjectPublicKeyInfo public key — matching private key stays offline). Same filename per env, content type is a parsing detail handled by \`secure_boot_pubkey_from_pem\`.

This drops, vs. #494 and #495:
- #494's separate \`bootloader/prod/secure-boot-public-key.pem\` filename.
- #495's parallel \`secure_boot_key/prod/public_key.pem\` directory.
- #495's committed \`key_digest.hex\` cache files (digests are derived at startup via \`compute_key_digest\`, so the PEM is the only authority).

The .gitignore line that blocked the path is removed — the file is now legitimately committable; accidental private-key commits get caught at review.

## Crate extraction (from #494)

\`secure_boot.rs\` moves out of \`frostsnap_factory\` (which depends on mysql, clap, tracing-subscriber — none of which belong in a build script) into a new \`frostsnap_secure_boot\` crate. \`sign_firmware\` takes its RNG by argument so the crate stays \`rand\`-free (only \`rand_core\`).

## API surface (mostly from #495, with two additions)

- \`verify_firmware(bytes) -> Result<Verified, VerifyError>\` where \`Verified { public_key, key_digest, image_digest }\`. One entry point, carries everything both callers need.
- \`Signer { Prod, Dev, Unknown }\` + \`classify_signer(key_digest, prod, dev)\`. Pinned digests passed in rather than read from disk, so this crate imposes no filesystem layout on its callers.
- \`compute_key_digest(&RsaPublicKey)\` — derives the ESP32 eFuse key digest from a pubkey using the same \`block[36..812]\` layout the verifier reads. Lets callers pin a signer from a PEM at startup, no committed digest file needed.
- \`secure_boot_pubkey_from_pem(pem)\` — accepts either PKCS1/PKCS8 private (dev) or SubjectPublicKeyInfo public (prod), returns \`RsaPublicKey\` either way.
- \`VerifyError::is_not_signed()\` distinguishes \"not a Secure Boot image at all\" from \"signed but corrupt.\"

## Test plan

- [x] \`cargo test -p frostsnap_secure_boot\` → 7 pass (sign/verify aligned and unaligned, PEM loader accepts both forms, compute_key_digest matches verify output, classify_signer routing, two espsecure fixture tests)
- [x] \`cargo test -p frostsnap_factory\` → passes
- [x] \`cargo run -p frostsnap_factory -- verify-firmware --input target/.../dev-frontier.bin\` → \`✅ Verified — signed by Frostsnap DEV key (not production!)\`
- [x] \`cargo run -p frostsnap_factory -- verify-firmware --input frostsnap_factory/bootloader/prod/signed-bootloader.bin\` → \`✅ Verified — signed by Frostsnap PROD key\`. Proves the committed \`bootloader/prod/secure-boot-key.pem\` produces the same key digest as the prod-signed bootloader's embedded signing key — i.e. **the new PEM matches the actual prod key**.

Build.rs wrong-env panic was the central scenario in #494's test matrix; the verification logic is preserved unchanged here.

## Closes
Closes #494, closes #495.

Co-authored-by: Nick Farrow <nick@nickfarrow.com>


frostsnap_factory/bootloader/prod/secure-boot-pubkey.pem
```diff
@@ -0,0 +1,11 @@
+-----BEGIN PUBLIC KEY-----
```
LLFourn, line 1: **Architecture: `secure-boot-key.pem` is overloaded to mean two different keys, and they coincide only for dev.**

The path is consumed two ways:
- **Verify** (build.rs, CLI) needs the **public** key — committed here, correct.
- **Sign** (`justfile:10` `secure_boot_key`, used at `justfile:130` by `sign-firmware` / `sign-bootloader` / `build-firmware-signed` / `full-provision`) needs the **private** key.

For **dev** that is one file (committed private key, pubkey derived). For **prod** they are necessarily different artifacts — the signing key is offline, the verification key is committed. Making the prod path a public key for verification is right, but it was done by collapsing the separate `secure-boot-public-key.pem` filename from #494 into this shared name, and that collapse is what creates the conflict.

Two concrete consequences, both in scope here:
- `frostsnap_factory/PROVISIONING.md` is now wrong: line 30 calls `secure-boot-key.pem` the *RSA-3072 private key for signing* and line 150 marks prod as *Gitignored (production secret)*. This PR makes both false, so it should fix them.
- The prod signing recipes pass `-k` = this public PEM, so `sign-firmware` fails with `UnexpectedTypeLabel` (expecting an RSA PRIVATE KEY) — repro courtesy codex. This is **not** a regression of any clean-checkout flow (the prod key was always offline / gitignored, so prod signing never ran from a vanilla checkout) — but it forces the real signing ceremony to overlay the private key onto a now-**tracked** path, which dirties a committed file and makes the catastrophic private-key commit easy.

**Suggested fix — two paths, resolves all of the above at once:** keep the private signing key at its own offline / gitignored path; give the committed public verification key a distinct, obviously-public filename (restore the #494 `secure-boot-public-key.pem` split or similar). A path that can only ever hold a public key makes the private-key footgun impossible by construction, removes the verify-vs-sign collision, and lets PROVISIONING.md tell the truth again.

frostsnap_secure_boot/src/lib.rs
```diff
@@ -225,26 +352,61 @@ mod tests {
...
+        let prod = [0x11u8; 32];
+        let dev = [0x22u8; 32];
+        assert_eq!(classify_signer(&prod, &prod, &dev), Signer::Prod);
+        assert_eq!(classify_signer(&dev, &prod, &dev), Signer::Dev);
+        assert_eq!(classify_signer(&[0xFFu8; 32], &prod, &dev), Signer::Unknown);
+    }
+
     #[test]
     fn matches_espsecure_format() {
         let reference = match std::fs::read("tests/fixtures/test-firmware-signed.bin") {
```
LLFourn: These two fixture tests are the only cross-check that the hand-rolled block format matches `espsecure.py`, the canonical ESP32 tool. They `return` (pass) when the fixtures are missing. They are committed now, so they run today — but a silent skip means if a fixture ever stops being present in CI you lose your only external oracle with zero signal: a green suite that no longer validates the thing it exists to validate.

Make absence loud: `include_bytes!` the fixtures so a missing one is a compile error, or at minimum `panic!` instead of `return` in the None arm.

LLFourn (commented):
Reviewed the full diff first-hand (verifier crypto, build.rs enforcement, CLI
categorization, committed key material) and ran the suite.

**This is a clean, correct unification of #494 + #495.** Verified: the crypto chain is
sound (firmware → image_digest → block check → RSA-PSS verify against the embedded key;
`key_digest = SHA256(block[36..812])`, tied to `compute_key_digest` by test); the
"self-consistently signed" vs "signed by *us*" distinction is modeled correctly and
enforced at both consumers; crate extraction is the right move; `cargo test -p
frostsnap_secure_boot` → 7 pass incl. both espsecure fixtures.

**One blocker, inline:** the "single filename" unification overloads
`secure-boot-key.pem` to mean both the committed **public verification key** and the
offline **private signing key** — which only coincide for dev. For prod they are
different artifacts, so the unification (a) makes PROVISIONING.md false, and (b) points
the prod signing recipes at a public PEM. Recommend restoring distinct paths (a separate
`secure-boot-public-key.pem` for the committed verification key, à la #494), updating the
justfile recipes + PROVISIONING.md. A public-only filename also makes the catastrophic
private-key commit impossible by construction.

Plus one non-blocking nit: make the espsecure fixture tests fail loudly on a missing
fixture rather than silently skipping.

musdom (approved):
Tested by building and verifying a signed dev firmware

PR #504, opened 2026-06-19
Add warning dialog for 1-of-N thresholds

nickfarrow:


LLFourn (approved):
Approved. I checked the implementation path and exercised the flow through the app simulator.

Evidence:

1-of-N threshold warning

CI mirror: https://github.com/LLFourn/frostsnap/pull/6 - green.

The warning is in the right place architecturally: `_beginThresholdKeygen` is the single choke point for keygen, including the "Generate keys" path that skips `next()`, so the prompt covers the risky wallet creation path without duplicating state. The condition `form.threshold == 1 && totalDevices > 1` targets 1-of-N wallets and leaves 1-of-1 alone.

The dismissal behavior is also right for a security prompt: barrier/back cancel via `?? false`, the safe "Reconsider" action is emphasized, and "Continue" explicitly proceeds.

Non-blocking nit: the copy says a single device/backup "will instantly be able to access all the bitcoin". That is directionally right for spend authority, but "instantly" slightly overstates it because spending still needs the coordinator/app and wallet descriptor. Not a blocker.

LLFourn:
@LLFourn need to rebase this too.

PR #502, opened 2026-06-18
Device naming improvements

nickfarrow:
* During conference demos, both @aphanley and I independently noticed users attempting to tap the pencil icon to select the field to enter device names (frustratingly unresponsive). d3d1f437b1ca8d2420f0cf560fc2ca8bb8d280ac makes it so tapping the icon activates the field. 
* I noticed that I often went to capitalize the first letter of device/wallet names for aesthetics, which is slow and difficult when demoing on peculiar orientations. 3a8a9bb6ffa1a3a6714546d4ba6167d423e39e5b

PR #505, opened 2026-06-20
Unique device names and no unauthenticated device renaming

nickfarrow:
* a8478f5a137e4e77e90aff089b3693683c847cb4 enforces unique device names for a wallet during keygen
* 39f8530f4e38c1e50ee31570bd1921f678220b01 retires the old device renaming code, but still keeps the message around so that this enum variant is not used by future messages (otherwise a new message would confused deserialization on old devices)

### DIFF
--- a/device/src/esp32_run.rs
+++ b/device/src/esp32_run.rs
@@ -504,13 +504,8 @@ impl<'a> DeviceLoop<'a> {
                             new_name: preview_name.clone(),
                         });
                     }
-                    frostsnap_comms::NameCommand::Prompt(new_name) => {
-                        self.ui
-                            .set_workflow(ui::Workflow::prompt(ui::Prompt::NewName {
-                                old_name: self.name.clone(),
-                                new_name: new_name.clone(),
-                            }));
-                    }
+                    // Retired rename command, kept only to reserve its wire slot.
+                    frostsnap_comms::NameCommand::_Prompt(_) => {}
                 },
                 CoordinatorSendBody::Core(core_message) => {
                     if matches!(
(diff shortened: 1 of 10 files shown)


LLFourn:
## Review + fix — duplicate device names (CI-fork integration triage)

Reviewed this ultra-carefully as part of our CI-fork integration triage (it touches the wire protocol + device firmware). The protocol/device side looks sound; on the **app** side we found — and fixed — one real bug in the new duplicate-name check.

**The bug:** `duplicateNamedDeviceIds` counted collisions over every retained entry in `_form.deviceNames`, but that map keeps a device's name after it disconnects while only *connected* devices participate in keygen. So: name device A "Cold" → unplug A → reuse "Cold" on a remaining device → a false *"Name already used in this wallet"* blocks Continue.

**Before (bug):**
before

**The fix — folded into this branch (head `8c495bf9`):** I rebased this PR onto current `master` (resolving the conflict with merged #502's device-naming changes, dropping the now-retired rename dialog) and applied the fix: `duplicateNamedDeviceIds` now counts collisions only over the live participants (`_deviceList.devices`). Two currently-connected devices sharing a name are still both flagged. All three commits are signed (@nickfarrow's two re-signed as maintainer, `%G? = G`).

**After (fixed):**
after

**Behavior note (intentionally not optimized):** `deviceNames` retains a name across disconnects and isn't transferred when reused, so one edge case remains by design — name A "foo" → unplug A → name B "foo" → **plug A back in** → both are connected and "foo", so both are flagged and you rename one. That's a genuine two-connected-devices conflict, so flagging it is correct; we don't make B silently "steal" A's name.

**Validation:** new regression test (5/5), `just lint-app` clean, Rust ordinary-libs suite green, and full CI (analyze, ordinary tests, all app builds, device firmware) green on the fork mirror `LLFourn/frostsnap#8`. The rebase clears the CI failures the pre-rebase branch had against merged #502.

LLFourn (approved):

PR #507, opened 2026-06-22
Centralize bottom safe-area handling (fix buttons under the Android nav bar)

LLFourn:
## Problem

On Android with edge-to-edge rendering, bottom action buttons can render *under* the system navigation bar (the gesture pill, or a transparent full-height 3-button bar) on some screens — making them hard or impossible to tap.

## Root cause

Safe-area insets were each **leaf screen's** responsibility — 22 hand-written `SafeArea`/`SliverSafeArea` wraps scattered across `lib/`. A cross-cutting concern done by per-leaf copy-paste is easy to forget, and the forgotten ones are the bug.

## Fix — give each *surface* a single inset owner

Every screen reaches the user through one of a few surfaces; make each own its bottom inset, so a leaf can't forget:

- **`MaybeFullscreenDialog`** (compact) wraps its child in `SafeArea(top: false)` — one owner for every fullscreen dialog (incl. the **Backup Keys** page via `wallet.dart`, and the recovery flow).
- **`showAppBottomSheet`** bakes the bottom inset into sheet content, since `showModalBottomSheet(useSafeArea: true)` is `SafeArea(bottom: false)` and never insets the bottom. `device_action` and both `settings` sheets migrate onto it.
- **`BottomActionBar`** owns custom bottom bars (a `Scaffold.bottomNavigationBar` child is *not* auto-inset by Flutter); the wallet toolbar uses it.
- `DialogContentWithActions` insets its own footer too.
- Removed two now-redundant per-leaf wraps; kept the full-`SafeArea` ones that also own the top edge.

`SafeArea` is idempotent, so the migration was safe and incremental.

## Guarantee it can't regress

`just lint-app` now runs `check-safe-area-surfaces`, which fails if `showModalBottomSheet(`/`Dialog.fullscreen(` (incl. typed `<T>`) appears outside the surface helpers. A new screen physically can't reintroduce the bug.

## ⚠️ Testing — visual verification still needed on Android 15

I could **not** get a faithful visual repro: the available test device (moto g04, Android 14) draws 3-button nav **opaque** (the OS insets the window), so the overlap doesn't manifest there, and its gesture-pill inset is too small to be dramatic. The fix is verified **by construction** — `SafeArea` consumes `MediaQuery.padding.bottom`, exactly the inset that's non-zero in every edge-to-edge / transparent-bar config — and on the Backup Keys code path (shown via `MaybeFullscreenDialog`, now surface-inset).

**Please verify visually on Android 15** (edge-to-edge enforced, nav bar transparent in both modes): recovery-flow footers, the device-action sheet, and the backup checklist with several devices.

Design/plan doc: `.clank/plans/fix-missing-safe-areas.md`.

### DIFF
--- a/frostsnapp/lib/device_action.dart
+++ b/frostsnapp/lib/device_action.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:frostsnap/theme.dart';
 import 'dart:async';
 
 Future<T?> showDeviceActionDialog<T>({
@@ -27,9 +28,8 @@ Future<T?> showDeviceActionDialog<T>({
       });
 
   final theme = Theme.of(context);
-  final result = await showModalBottomSheet<T>(
-    context: context,
-    isScrollControlled: true,
+  final result = await showAppBottomSheet<T>(
+    context,
     isDismissible: false,
     backgroundColor: Colors.transparent,
     builder: (BuildContext dialogContext_) {
--- a/frostsnapp/lib/dialog_content_with_actions.dart
+++ b/frostsnapp/lib/dialog_content_with_actions.dart
@@ -33,14 +33,17 @@ class DialogContentWithActions extends StatelessWidget {
           ),
         ),
         if (!isCompact) const Divider(height: 0),
-        Padding(
-          padding: const EdgeInsets.all(16.0),
-          child: OverflowBar(
-            alignment: actionsAlignment,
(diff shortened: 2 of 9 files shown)


nickfarrow:
I was able to reproduce the original bug on master, the backup information screen buttons being hidden behind the android navigation menu. And confirm this fixes the problem.

nickfarrow (approved):

PR #508, opened 2026-06-24
Electrum: fix disable/startup behavior + signet servers; refactor config to a watch subject

LLFourn:
Two commits.

## 1. Fix electrum: respect disabled state, interrupt connections, fix signet servers

With both electrum servers disabled, the coordinator kept connecting to them — on a fresh start with a persisted-disabled config, and after toggling them off at runtime.

Root causes:
- the connection handler's enabled state was never initialized from persisted settings (it defaulted to "all")
- the connect attempt and reconnect backoff didn't poll the control channel, so a disable could never interrupt them

Changes:
- initialize the handler's enabled state from persisted settings at startup → restart-with-disabled never attempts a connection
- race the connect attempt and reconnect backoff against control messages → disabling (or a url change / reconnect) interrupts an in-flight or retrying attempt instead of churning through every server
- attribute connection failures to their server url at the `Conn::new` boundary, so the url travels in the error value, not just the log line
- replace the broken signet defaults: primary `tcp://signet.musdomworks.com:50001`, backup `ssl://mempool.space:60602` (the old defaults were down / rate-limited)

## 2. Refactor electrum config into a single `tokio::sync::watch` subject

The fix above left the settings-derived desired state (enabled + primary/backup urls) duplicated: authoritative in `Settings`, replicated in `StatusTracker`, kept in sync by a startup snapshot plus `SetEnabled`/`SetUrls` messages.

Model it as one `tokio::sync::watch<ElectrumConfig>` shared api→handler: the api writes it, the handler queries it (`should_connect` / `try_connect` read `config_rx.borrow()`) and reacts to its change signal (`config_rx.changed()` preempts idle/connect/backoff/connected and re-emits status with fresh urls). Deletes `Message::SetEnabled`, `Message::SetUrls`, the startup snapshot, and the `StatusTracker` replicas; `ChangeUrlReq` becomes probe-only with the api updating the config on success — removing the duplication that made the startup ordering race possible.

## Testing

- `cargo check` + `cargo test` on `frostsnap_coordinator` and `rust_lib_frostsnapp`
- an ignored network test verifies a signet server connects through the real `Conn::new` (rustls + TOFU + genesis check)
- manually verified on signet: connects via the primary, disabling both servers stops/interrupts attempts immediately, restart-with-disabled never connects, re-enabling / changing a url reconnects

### DIFF
--- a/frostsnap_coordinator/Cargo.toml
+++ b/frostsnap_coordinator/Cargo.toml
@@ -27,7 +27,7 @@ bdk_electrum_streaming = { version = "0.5.2" }
 bdk_coin_select = { version = "0.4.1" }
 
 
-tokio = { version = "1.45", features = ["net", "rt", "time", "macros"] }
+tokio = { version = "1.45", features = ["net", "rt", "time", "macros", "sync"] }
 rustls = { version = "0.23", default-features = false, features = ["tls12", "ring"] }
 rustls-pki-types = { version = "1" }
 tokio-rustls = { version = "0.26", default-features = false, features = ["tls12", "ring"] }
--- a/frostsnap_coordinator/src/bitcoin/chain_sync.rs
+++ b/frostsnap_coordinator/src/bitcoin/chain_sync.rs
@@ -31,6 +31,7 @@ use std::{
     time::Duration,
 };
 use tokio::io::AsyncWriteExt;
+use tokio::sync::watch;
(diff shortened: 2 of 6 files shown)


musdom:
when I switch to mempool.space as the signet primary, it goes into a retry loop and never fall back to backup server

musdom (approved):

PR #511, opened 2026-07-10
Android: empty-key fallback on hardware key-creation failure + delete-and-recover on key mismatch

LLFourn:
Three fixes to Android hardware-key handling, all keeping the app working instead of throwing.

Old-phone empty-key fallback. On old keymasters (e.g. Pixel 2 XL / Android 11) the Android Keystore
can fail to generate the app's encryption key even in the TEE, which previously threw and broke keygen
so the wallet couldn't be created at all. SecureKeyManager now reports KEY_CREATION_FAILED, and
AndroidSecureKeyProvider.getOrCreateKey falls back to an empty encryption key so keygen still works.

Unavailable/wrong key on an existing wallet -> delete-and-recover. A compact guard (existingWalletKey
in wallet_key_mismatch.dart) at the operation call sites (sign tx/message, show/check backup, send,
restore-into-existing) returns null and shows a "Wallet needs recovery" dialog, routing into the
existing delete flow -- instead of an opaque failure or a false "set up a screen lock" prompt (a new
lock would mint a different key that cannot decrypt the old data).

Recovery-mode auto-exit uses the same guard. When a device is plugged in with a physical backup still to
consolidate, the app takes it back out of recovery mode in the background. That path was the last
existing-wallet caller still using `SecureKeyProvider.getEncryptionKey()` — the key-ESTABLISHING surface — so
it skipped the empty-key probe every other existing-wallet operation does, and could hand a non-decrypting
key to `consolidate_pending_physical_backups`, whose `root_shared_key(..).expect("invariant")` then panicked
while the caller held the coordinator mutex, poisoning it for every later call. The re-plug event now names
the wallet the device has pending work for, so the auto-exit goes through `existingWalletKey` like everything
else: it probes the empty key first, and when the key still can't unlock that wallet — a cleared screen
lock, for instance — it shows the "Wallet needs recovery" dialog. Because this fires from a background
listener with no wallet in scope, that dialog has no Delete button and points at the wallet's settings
instead; the device is left in recovery mode. `consolidate_pending_physical_backups` returns a typed `CannotDecrypt` instead of
asserting an invariant it cannot know — `encryption_key` is caller-supplied — and the caller logs it and
leaves the device in recovery mode. (Folds in what was PR #521, now closed.)

This PR adds no unit tests -- see Coverage below for what is and is not exercised.

## Demos — the delete-and-recover UX (headlines 2 and 3)

These gifs show **only** the existing-wallet recovery path: an existing wallet whose stored encryption
key becomes unavailable (here, deleted) can no longer decrypt its data, so each operation routes into
"Wallet needs recovery" → delete-and-recover. They do **not** exercise the empty-key fallback (headline
1) — that path applies at keygen, produces no dialog, and is covered by the `key_empty_key` fsim driver (see Coverage).

### The full recovery flow

The recording deletes the app's stored encryption key for an **existing** wallet, so it can no longer
decrypt that wallet's data:

**delete the encryption key** → "Show secret backup" → "Wallet needs recovery" → **Delete wallet** →
hold-to-delete → recover from the still-connected device → working wallet.

full delete-and-recover workflow

### Where the dialog fires (per call site)

show-backup
sign-message
send

### The re-plug auto-exit (headline 3)

A device carrying a pending physical-backup consolidation, plugged back in after the phone's screen
lock was cleared.

consolidation with the encryption key cleared

Reaching this state needs a real physical backup: create a 1-of-1 wallet, have the device display its
backup, delete the wallet, then restore it by typing those 25 words on a fresh device — **with that
device unplugged when the restore is finished**. That last part is what makes the path reachable at
all, because `finish_restoring` consolidates inline for every device that happens to be plugged in:

```rust
for device_id in needs_consolidation {
    // NOTE: This will only work for the devices that are plugged in otherwise it's a noop
    self.exit_recovery_mode(device_id, encryption_key);
}
```

using the key the restore flow is already holding. A device present at that moment leaves nothing
pending, so the deferred re-plug auto-exit only ever runs for devices that were absent.

Then the screen lock is cleared (`locksettings clear`, invalidating the auth-bound Keystore key) and
the device plugged back in. The coordinator logs either side of that registration are the actual
evidence: `DeviceNeedsConsolidation` is mutated once at restore and never cleared, and after the
plug-back-in there is no `Consolidate` message, no "asking device to exit recovery mode", no
`FinishedConsolidation`, and no panic — so no poisoned coordinator mutex.

## Coverage

This PR contains **no unit tests**. Evidence is end-to-end, or review only where stated.

| Path | Evidence |
|---|---|
| Headline 1 — forced native `KEY_CREATION_FAILED` → Dart empty-key fallback → **successful keygen + a working (decryptable) wallet** | fork `key_empty_key` e2e driver (fsim; stays on the fork) |
| Headline 2 — guard routing when the app key is **unavailable/deleted**, at the demonstrated call sites | the recovery gifs above (end-to-end UX) |
| Ordinary top-level duplicate restoration across correct / wrong / unavailable key states | fork `key_mismatch` e2e driver (fsim; stays on the fork) |
| Headline 3 — re-plug auto-exit routed through the same guard, with the app key **unavailable** | fork `key_consolidation_pin_clear` e2e driver (fsim; stays on the fork) + the gif above. Proves the re-plug reaches `existingWalletKey`, that it refuses the unavailable (and non-decrypting empty-key) case, shows the dialog, and never calls `exitRecoveryMode` — so the panic path is not taken |
| Headline 3 — `consolidate_pending_physical_backups` returning typed `CannotDecrypt` instead of asserting | **none** — code review only. The Dart guard short-circuits first, so this branch is unreachable from the driver above (see below) |

**Still not exercised by any automated test:** the **native Kotlin** catch itself — `SecureKeyManager.kt`
catching a real hardware `createKey()` exception. The `key_empty_key` e2e uses the debug hook, which
returns the `KEY_CREATION_FAILED` **error result** directly (proving the Dart fallback and keygen), but
does not make the real keymaster throw. Also unexercised: the **wrong-but-present key** case — a key that exists but cannot decrypt the wallet. The gifs delete the key or clear the lock screen (making it unavailable), and `key_mismatch` deliberately never invokes a decrypting operation, so nothing here demonstrates that path — including for the headline 3 auto-exit, which is exercised only in its unavailable-key form.

The typed `ConsolidateError::CannotDecrypt` is likewise unexercised, and structurally so rather than by
omission: the guard returns null before `exitRecoveryMode` is called, so
`consolidate_pending_physical_backups` is never reached with a bad key from this path. It is
defence-in-depth for a caller that supplies one anyway — which is the point, since `encryption_key` is
caller-supplied and the previous `expect("invariant")` panicked while holding the coordinator mutex.
Reviewing it as unexercised-by-construction rather than untested is the accurate reading.

Likewise StrongBox→TEE fallback and app-key stability across reboots. These stay real-device /
code-review coverage.


LLFourn:
ci is not passing can you fix this @LLFourn

nickfarrow:
Added two commits:

- `2cdb8d27` Probe the empty key before showing delete-and-recover. The fallback isn't saved anywhere, so if the Keystore later returns a real key, the wallet made under the empty key looks unrecoverable. The dialog would tell the user to delete a wallet that can still be opened. Try the empty key first.
- `3fb9d457` In circumstances where the dialog has no "Delete wallet" button (no WalletContext), the text now says where to find delete.

Should we have a one-time dialogue popup that the wallet may not have strong data protection on the mobile phone?
- `exitRecoveryMode` in main.dart still uses the old key path. On a wrong key it silently does nothing and the device stays in recovery mode.

frostsnap_core/src/coordinator/restoration.rs
```diff
@@ -979,12 +984,17 @@ impl FrostCoordinator {
...
-                .expect("invariant");
-            let frost_key = self
-                .get_frost_key(consolidation.access_structure_ref.key_id)
-                .expect("invariant");
+                .root_shared_key(access_structure_ref, encryption_key)
+                .ok_or(ConsolidateError::CannotDecrypt {
+                    access_structure_ref,
+                })?;
+            let frost_key = self.get_frost_key(access_structure_ref.key_id).ok_or(
+                ConsolidateError::CannotDecrypt {
```
nickfarrow, line 994: This error is a missing wallet not a decryption error, could motivate its own variant or at least an error message that doesn't blame the key.

frostsnapp/lib/main.dart
```diff
@@ -91,31 +90,17 @@ Future<void> main() async {
...
-                showErrorSnackbar(
-                  ctx,
-                  expected
-                      ? "Couldn't take device out of recovery mode: screen lock required."
-                      : "Couldn't take device out of recovery mode: ${e.message ?? e.code}",
-                );
-              }
-              return;
-            }
+            final encryptionKey = await existingWalletKey(
```
nickfarrow, line 99: `existingWalletKey` can still throw `PlatformException` (e.g. `KEY_ERROR`), only `NO_LOCK_SCREEN` becomes `WalletKeyUnavailable`. So these would now be unhandled exceptions where the old code caught and logged everything. I think we should should catch and log/show them.

nickfarrow:
I see tests are now in `fsim`? I think fine for now but we should follow up with proper testing coverage of this feature. Also noting the empty-key probe path from `256ac614` isn't covered by the `fsim` drivers.


You implemented my prior comment of warning "wallet isn't hardware-protected" (`f4a28679d656b9db0ae155af008749efa48b33ac`) but left it out here. Curious as to what you thought? or if accidentally dropped.

nickfarrow (approved):
Tested on Pixel 9 GrapheneOS: removed the screen lock on an existing wallet, got the recovery dialog. Same again after re-adding a PIN (wrong key path). Delete and recover worked, backup viewable after. New wallet with no lock correctly prompts to set one up, cancel doesn't crash, setting one progresses.

I also forced `KEY_CREATION_FAILED` via the debug flag, keygen fell back to the empty key (logcat confirmed) and the wallet works. Cleared the flag, restarted, backup still shows: the probe rescues the wallet once the Keystore returns a real key.

One note from testing: the probe rescues silently, no log. A log line when it fires would make these phones diagnosable in the wild.

Approving with my comments above still open.

LLFourn:
Closing this in favour of #533, which is signed, green and ready for review.

**This is a change of approach, not a rejection of the work here.** #511 kept the strict
`KeyGenParameterSpec` and added an **empty-key fallback** for when key creation failed; #533 relaxes the
spec instead, which removes the device-compatibility failure that fallback existed for — so the fallback
goes rather than being fixed. The *other* half of this PR's problem does not go away and #533 still
handles it deliberately: legacy auth-bound keys that get invalidated, or replaced by a live-but-wrong key,
still need per-wallet recovery. Once the fallback inverted like that, rewriting this PR would have been a
bigger and less honest diff than opening a fresh one.

@nickfarrow — your review here shaped what shipped, and two things in particular:

- Your **Pixel 9 testing** caught behaviour we would not have seen otherwise.
- Your review found **two real bugs**, and they end up in different places — worth being exact rather than
  saying "both fixed":
  - `824c4c23` (point at wallet settings when the dialog has no delete button) — **carried into #533**; the
    context-less branch is there in `wallet_key_mismatch.dart`.
  - `256ac614` (probe the empty key before advising delete-and-recover) — **eliminated structurally, not
    carried**. It fixed the non-persisted empty-key fallback; #533 deletes that fallback outright, so there
    is no longer an empty key to probe. The bug is gone because its cause is gone, which is a better
    outcome than porting the fix but a different one, and I would rather say so than let it read as
    incorporated.
  On provenance: both are commits on *this* branch and #533 is a single fresh commit on `master`, so they
  are not ancestors of it. You have `Co-authored-by` credit on the commit.

Your question about warning users that the wallet isn't hardware-protected (`f4a28679`) is **answered by
#533 rather than deferred**, which is worth stating explicitly since "we'll look at it later" is the usual
fate of that kind of question. It was about this PR's **fallback to an empty key** — if the Keystore could
not give us a key, #511 carried on with an empty one, so the wallet was silently unprotected and nobody
was told. #533 deletes that fallback: there is now either a Keystore key or an error, never a silent
downgrade, so the case you were warning about cannot arise.

The narrower version does survive, and #533's description says so rather than glossing it: we request an
`AndroidKeyStore` key but do not ask for StrongBox or check `isInsideSecureHardware`, so on a device whose
Keystore is software-backed the key is Keystore-protected rather than hardware-protected. That is
unchanged by either PR and is worth its own discussion if you think it warrants user-facing wording.

#533 also carries the evidence this PR never got: fsim recordings of all five call sites where a
live-but-wrong key now routes to recovery, plus the negative case showing a newly created wallet works
normally in the same session while the pre-existing one does not.

PR #512, opened 2026-07-13
[app] Fix duplicate-name false positive when a freed name is reused after disconnect

LLFourn:
A fix for the duplicate-device-name check in #505, targeted at that PR's branch so it can be folded in.

### The bug
`WalletCreateController.duplicateNamedDeviceIds` counts name collisions over every retained entry in `_form.deviceNames`. That map keeps a device's typed name **after it disconnects**, while only the currently connected devices participate in keygen. So: name device A "Cold" → unplug A → reuse "Cold" on a remaining device → a false *"Name already used in this wallet"* blocks Continue even though only one connected device is named "Cold".

**Before (bug):**
before

### The fix
Delegate to a pure `duplicateNamedDeviceIdsAmong(participants, names)` that counts collisions **only over the live participants** (`_deviceList.devices`). Two currently-connected devices sharing a name are still both flagged (unchanged) — only names held by *disconnected* devices stop producing a false collision.

**After (fixed):**
after

### Behavior note — reconnecting the original device (intentionally not optimized)
`deviceNames` retains a name across disconnects and it's **not** transferred when another device reuses it. So one edge case remains by design: name A "foo" → unplug A → name B "foo" (fine) → **plug A back in** → A and B are now both connected and both "foo", so both are flagged and you resolve it by renaming one. That's a genuine two-connected-devices-same-name conflict — flagging it (what this feature is for) is the correct outcome, so we deliberately don't make B "steal" the name from A.

### Validation
- New `frostsnapp/test/wallet_create_duplicate_name_test.dart` (5 cases: the reuse-after-disconnect false positive, the true-positive, case/trim/empty). `flutter test` → 5/5.
- `just lint-app` clean.
- Full CI (analyze, ordinary tests, all app builds, device firmware) ran green on the fork mirror `LLFourn/frostsnap#8`.

### DIFF
--- a/device/src/esp32_run.rs
+++ b/device/src/esp32_run.rs
@@ -504,13 +504,8 @@ impl<'a> DeviceLoop<'a> {
                             new_name: preview_name.clone(),
                         });
                     }
-                    frostsnap_comms::NameCommand::Prompt(new_name) => {
-                        self.ui
-                            .set_workflow(ui::Workflow::prompt(ui::Prompt::NewName {
-                                old_name: self.name.clone(),
-                                new_name: new_name.clone(),
-                            }));
-                    }
+                    // Retired rename command, kept only to reserve its wire slot.
+                    frostsnap_comms::NameCommand::_Prompt(_) => {}
                 },
                 CoordinatorSendBody::Core(core_message) => {
                     if matches!(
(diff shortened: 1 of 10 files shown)


LLFourn:
Closing this — the fix is now folded **directly into #505's branch** instead.

A dependent PR onto `device-name-uniqueness` could never go green: that branch was itself conflicting/stale against merged #502 ("Device naming improvements"), so its own device-firmware + Rust-test jobs failed regardless of this fix. So I rebased #505 onto current `master` (resolving the #502 device-naming conflict by dropping the retired rename dialog), applied this fix on top, signed all three commits, and pushed to `device-name-uniqueness`. The fix + regression test now live on #505 directly (head `8c495bf9`).

PR #514, opened 2026-07-14
[app] Show a 'saving' state before the unplug prompt at keygen

LLFourn:
Gives the device a brief moment to settle after keygen before showing the "unplug devices" prompt — reducing the risk of a device being unplugged before its share write has landed.

Note: this is a fixed short delay, **not** an acknowledged-persistence guarantee — the app queues Finalize and waits ~1s; it does not receive a device persistence ack. If a hard completion guarantee is wanted, an explicit ack would be the way to get it.

Validated: builds green on our CI fork (Flutter analyze + app builds).

### DIFF
--- a/frostsnapp/lib/wallet_create.dart
+++ b/frostsnapp/lib/wallet_create.dart
@@ -990,97 +990,163 @@ class _WalletCreatePageState extends State<WalletCreatePage> {
 
           if (!state.allAcks) continue;
 
-          final keygenCodeMatches =
-              await showDialog<bool>(
-                context: context,
-                barrierDismissible: false,
-                builder: (context) {
-                  final theme = Theme.of(context);
-                  return BackdropFilter(
-                    filter: blurFilter,
-                    child: AlertDialog(
-                      title: Text('Final check'),
-                      content: Column(
-                        mainAxisSize: MainAxisSize.min,
-                        crossAxisAlignment: CrossAxisAlignment.stretch,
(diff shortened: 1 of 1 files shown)


nickfarrow:
If we have the saving state as a page step then it is a bit nicer since we no longer show the "Select threshold" screen in the background behind the waiting dialog. See https://github.com/frostsnap/frostsnap/pull/517

LLFourn:
This is just a one second delay. it doesn't need any a new dialog. it should just show a spinner in the finish button that triggered the delay.

LLFourn:
Implemented the in-button-spinner direction — no dialog. PR-to-the-PR into this branch: #518.

in-button spinner

The "Final check → **Yes**" button spins in place during the ~1s finalize, then closes; the separate `_KeygenSavingDialog` is gone (net −33 lines). Safety details for the async finalize path (PopScope so back can't dismiss mid-save, re-entrancy guard, fixed spinner footprint, single failure owner / no unsafe retry) are in #518. `flutter analyze` clean on both bases; the video is the success path only (1-of-1 sim).

Supersedes the page-step approach — @nickfarrow, flagging so #517 isn't left waiting.

To land: merge #518 into this branch and reconstruct as the signed commit (dropping the "so the write completes first" wording), rebased onto current master.

nickfarrow (approved):

nickfarrow:
Merged inbound PR and squashed

Issue #503, opened 2026-06-19
Transaction ordering issue (MAC)


happytobeunhinged:
Sent 0.00025000 to wallet
then sent 0.00010308 off to another wallet
sent it back and received 0.00009556 back to wallet

The balance is correct but the order of transactions is wrong and would be confusing if you had a lot.

![Image](https://github.com/user-attachments/assets/b7f545a3-097d-4ccf-a1fa-f8be59f113ca)

![Image](https://github.com/user-attachments/assets/a521e82f-f8dc-47ad-9761-4b871869b34a)

PR #516, opened 2026-07-16
Fix transaction list ordering

nickfarrow:
Transactions were sometimes showing up in the wrong order (#503).

We were ordering the list by bdk's `list_ordered_canonical_txs` (which sorts by spend-depth, not time) and just reversing it. So a fresh unconfirmed receive, being a "root" in the spend graph, sank below older confirmed txs.

Changed to sort pending first, then confirmed by most recent block, and only fall back to spend order (child above parent) to break ties.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/wallet.rs
+++ b/frostsnap_coordinator/src/bitcoin/wallet.rs
@@ -326,7 +326,10 @@ impl CoordSuperWallet {
 
     pub fn list_transactions(&mut self, master_appkey: MasterAppkey) -> Vec<Transaction> {
         self.lazily_initialize_key(master_appkey);
-        let mut txs = self
+        // bdk's canonical order is topological (spend-depth), not by time, so reverse it to
+        // get child-before-parent, then sort newest-first by chain position (pending first,
+        // then confirmed by height). Stable sort keeps the child-before-parent tiebreak.
+        let mut canonical = self
             .tx_graph
             .graph()
             .list_ordered_canonical_txs(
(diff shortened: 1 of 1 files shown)


LLFourn (approved):
Reviewed and approving as a pragmatic fix.

Verified against `bdk_chain-0.23.3`: the stable `sort_by_key(Reverse(chain_position))` over the reverse of bdk's topological canonical order gives unconfirmed-first, then confirmed newest-block-first, with equal positions resolving child-above-parent via the stable tiebreak. This fixes the #503 interleaving — the old `txs.reverse()` on bdk's spend-depth (topological) order could drop a fresh unconfirmed tx into the middle of the list.

One honest caveat worth recording: the ordering is chain-position-primary and keeps topology only as an equal-key tiebreak, so it can still show a parent above its child in a rare unconfirmed edge. `ChainPosition::Ord` orders unconfirmed txs by `first_seen`/`last_seen` (with `None` sorting last), so an unconfirmed ancestor with `first_seen = None` sorts above a `first_seen = Some` child under `Reverse`, and the differing keys mean the stable topological tiebreak can't repair it.

That's not a defect in this PR — a fully DAG-consistent newest-first display order is implementable (Kahn's over the reversed DAG with a `Reverse(ChainPosition)` priority queue, i.e. the same machinery #2219 adds in the parent-before-child direction), and ideally lives upstream in bdk. Filed bitcoindevkit/bdk#2247 for a display-oriented canonical ordering; cf. bitcoindevkit/bdk#2201 / #2219.

For a small wallet-side fix this is a clear improvement over the status quo. LGTM.

PR #518, opened 2026-07-26
keygen: in-button spinner instead of the saving dialog (for #514)

LLFourn:
Implements the in-button spinner LLFourn asked for on #514 — no separate "Saving wallet to devices" dialog. Targets #514's branch; base it in and reconstruct as the signed commit when adopting (this is an unsigned experiment branch).

**Change** (`wallet_create.dart`, net −33 lines): on the "Final check → **Yes**" button, spin the Yes button in place while `getEncryptionKey → finalizeKeygen → ~1s settle` run inline, then close. Deletes the standalone `_KeygenSavingDialog` and the `_KeygenFinalizeResult`/`Success`/`Failure` classes; the dialog returns `AccessStructureRef?` directly.

**Safety (keygen-finalize path):**
- `PopScope(canPop: !saving)` so system-back can't dismiss the route mid-finalize; `barrierDismissible: false` kept.
- Re-entrancy guard — a second Yes is a no-op; both buttons disabled once saving.
- Fixed-footprint spinner (`Opacity` + `Stack`, 16px / strokeWidth 2) so the button doesn't reflow.
- `mounted` / `context.mounted` checked after every await before any pop/snackbar/setState.
- Single failure owner, exactly once: error → pop `null` + the existing `Failed to finalize keygen: $e` snackbar; the outer `asRef == null` uniformly resets keygen state for both No and failure — no retry path left open.
- Kept the `Semantics(liveRegion, label: 'Saving wallet to devices')` announcement.

**Video** (Android fsim, 1-of-1 keygen):

in-button spinner

Final check → tap **Yes** → "No" greys out and the "Yes" label becomes an in-place spinner (~1s) → wallet home. **Success path only** — the failure/snackbar path isn't recorded.

`flutter analyze`: **clean** on both the CI-fork base and #514's head `57378723`.

**Note for reviewers:** the existing e2e drivers `keygen_drive.dart` / `keygen_2of3_drive.dart` `waitFor('Saving wallet to devices')` still pass (that string is kept as the `Semantics` live-region label), but the separate saving *dialog* no longer appears — worth a driver comment.

Supersedes the page-step approach in #517.

### DIFF
--- a/frostsnapp/lib/wallet_create.dart
+++ b/frostsnapp/lib/wallet_create.dart
@@ -990,114 +990,163 @@ class _WalletCreatePageState extends State<WalletCreatePage> {
 
           if (!state.allAcks) continue;
 
-          final keygenCodeMatches =
-              await showDialog<bool>(
-                context: context,
-                barrierDismissible: false,
-                builder: (context) {
-                  final theme = Theme.of(context);
-                  return BackdropFilter(
-                    filter: blurFilter,
-                    child: AlertDialog(
-                      title: Text('Final check'),
-                      content: Column(
-                        mainAxisSize: MainAxisSize.min,
-                        crossAxisAlignment: CrossAxisAlignment.stretch,
-                        spacing: 16,
-                        children: [
-                          Text('Do all devices show this code?'),
-                          Card.filled(
-                            child: Center(
-                              child: Padding(
-                                padding: EdgeInsets.symmetric(
-                                  vertical: 12,
-                                  horizontal: 16,
-                                ),
-                                child: Column(
-                                  mainAxisSize: MainAxisSize.min,
-                                  children: [
-                                    Text(
-                                      '${form.threshold}-of-${form.selectedDevices.length}',
-                                      style: theme.textTheme.labelLarge,
-                                    ),
-                                    Text(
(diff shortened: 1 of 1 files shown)


LLFourn:
In the video there's a slight jump when "yes" is pressed and replaced with the spinner because of the change in height. Stabalize that please.

LLFourn:
Fixed — no height jump now (`ee5497b6`).

no jump

There were two independent size changes on Yes→spinner, both now eliminated **by construction** (no hardcoded heights):

1. **Button** — the spinner is a `Positioned.fill` overlay in the `Stack`, so the always-laid-out (transparent) `Text('Yes')` is the only size-determining child; the button's width *and* height stay locked to the label.
2. **Dialog** — the `saving` `Semantics` announcement had been a conditional child of the `spacing: 16` content `Column`, so flipping it in added a 16px gap and grew (re-centered) the whole `AlertDialog`. It's now a `Positioned.fill` overlay contributing zero layout, so the card is identical whether saving or not.

Frame-by-frame, the dialog card edges, "Final check" title, code card, "No" baseline, and actions-row Y are pixel-identical between the pre-tap and during-spinner frames. `flutter analyze` clean.

PR #519, opened 2026-07-26
test(coord): fixed verifier time for TOFU fixture tests (fix Test-ordinary CI time-bomb)

LLFourn:
`frostsnap_coordinator/tests/certs/test1.der` + `test2.der` expired **2026-07-25**, so **`Test ordinary libraries` goes red on every PR after that date** — each PR's last green is a pre-expiry run.

**Why it breaks at expiry:** the `tofu::verifier` tests called `verify_server_cert(.., now = UnixTime::now())`. Once `now` passes the fixtures' `notAfter`, the base `WebPkiServerVerifier` returns a cert-expired error instead of the unknown-issuer path, so `TofuCertVerifier` no longer captures a `take_tofu_error` — `test_tofu_first_use_capture` and `test_certificate_confusion_attack_prevention` then assert `Some` and get `None`.

**Fix:** these tests exercise TOFU handling of an *otherwise-valid* self-signed cert, so the verifier's `now` is **test input, not a real clock**. Added a fixed `test_now()` (2026-01-01, inside the fixtures' 2025-07-25…2026-07-25 validity) and replaced all five `UnixTime::now()` sites with it. Removes the wall-clock dependence entirely — the tests never expire again, no cert-renewal treadmill. **Certificates unchanged.** All 7 `bitcoin::tofu::verifier` tests pass in CI shape (`--release --all-features --locked`).

If cert-expiry behaviour ever needs a contract, it belongs in its own test with its own chosen time.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/tofu/verifier.rs
+++ b/frostsnap_coordinator/src/bitcoin/tofu/verifier.rs
@@ -283,6 +283,18 @@ mod tests {
     const TEST_CERT1: &[u8] = include_bytes!("../../../tests/certs/test1.der");
     const TEST_CERT2: &[u8] = include_bytes!("../../../tests/certs/test2.der");
 
+    /// A fixed verification time inside the `test1.der`/`test2.der` validity window
+    /// (2025-07-25 .. 2026-07-25). These tests exercise TOFU handling of an
+    /// *otherwise-valid* self-signed certificate, so the verifier's `now` is a test
+    /// input and must not depend on the wall clock — otherwise the fixtures "expire"
+    /// and the base verifier returns cert-expired instead of the unknown-issuer path
+    /// the TOFU capture relies on. (Certificate-expiry behaviour, if ever needed,
+    /// belongs in its own test with its own chosen time.)
+    fn test_now() -> UnixTime {
+        // 2026-01-01T00:00:00Z — inside the test fixtures' validity window
+        UnixTime::since_unix_epoch(std::time::Duration::from_secs(1_767_225_600))
+    }
+
     #[test]
     fn test_certificate_fingerprint_format() {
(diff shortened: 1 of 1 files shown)


LLFourn:
Switched approach after review (force-pushed `e5ac7e20` → `1a74794b`): rather than regenerate the certs with a far-future expiry (v1), the tests now use a fixed `test_now()` inside the fixtures' validity instead of `UnixTime::now()`. These tests are about TOFU-of-a-valid-cert, not expiry, so the clock is test input — this kills the time-dependence permanently rather than renewing it. Certs unchanged; 7/7 verifier tests green in CI shape.

nickfarrow (approved):

frostsnap_coordinator/src/bitcoin/tofu/verifier.rs
```diff
@@ -283,6 +283,18 @@ mod tests {
     const TEST_CERT1: &[u8] = include_bytes!("../../../tests/certs/test1.der");
     const TEST_CERT2: &[u8] = include_bytes!("../../../tests/certs/test2.der");
 
+    /// A fixed verification time inside the `test1.der`/`test2.der` validity window
+    /// (2025-07-25 .. 2026-07-25). These tests exercise TOFU handling of an
+    /// *otherwise-valid* self-signed certificate, so the verifier's `now` is a test
+    /// input and must not depend on the wall clock — otherwise the fixtures "expire"
+    /// and the base verifier returns cert-expired instead of the unknown-issuer path
+    /// the TOFU capture relies on. (Certificate-expiry behaviour, if ever needed,
+    /// belongs in its own test with its own chosen time.)
```
LLFourn, line 292: this comment is alittle verbose. Too much info. It's obvious why you would fix it.

nickfarrow: Want to reduce this further or ready to merge?

PR #521, opened 2026-07-29
exitRecoveryMode: route the re-plug auto-exit through the guarded key path + fail-safe consolidation

LLFourn:
Follow-on to #511 that closes the `exitRecoveryMode` gap raised in review, and hardens the shared consolidation primitive so a key mismatch can never panic. Opened against the `keystore-empty-key-upstream` branch so it stacks on the empty-key probe — happy to squash/rebase however you prefer.

## Why

`exit_recovery_mode` needs the app decryption key, and every caller supplies it from a wallet-aware context — **except** the `main.dart` device-list `RecoveryMode` listener (the deferred auto-exit that fires when a device with pending consolidations re-plugs). It reached into the global `SecureKeyProvider.getEncryptionKey()` with no wallet context, so it:

1. used the *establishing* key variant (which can mint a fresh key) for what is really an existing-wallet decrypt;
2. could **panic** via `root_shared_key(..).expect("invariant")` inside `consolidate_pending_physical_backups` on a key that can't decrypt; and
3. never went through the `existingWalletKey` guard, so it missed the empty-key probe this PR adds.

Because a wallet can be sealed under the empty-key fallback *or* a hardware key, one device can hold pending consolidations across different key-eras — so no single key is guaranteed to decrypt the whole set, and "check the first, then loop the rest" still panics on a later entry.

## What

- **core** — `consolidate_pending_physical_backups` preflights every pending consolidation and returns a typed `ConsolidateError` instead of `.expect`: all-or-nothing, **zero messages emitted** on any undecryptable ref. `exit_recovery_mode` treats the error as a logged no-op (sends nothing, does not flip recovery mode off), so `recover_share`/`finish_restoring` are fail-safe too. `has_backups_that_need_to_be_consolidated -> bool` becomes `backups_that_need_to_be_consolidated -> Vec<PendingConsolidation>`.
- **event** — the `RecoveryMode` device-list change now carries the full **unique** set of pending access-structure refs (the wallet identity the emitter already holds). The re-plug site supplies it; the mid-flow and exit transitions carry an empty set.
- **app** — the listener routes through a new `exitRecoveryModeForDevice`, which acquires a candidate via `existingWalletKey` and verifies it decrypts **every** pending ref before auto-exiting. On a whole-set mismatch it sends nothing and surfaces the "wallet needs recovery" dialog.

## Tests

- **core:** a mixed-key regression test — one device recovering two wallets sealed under different keys — asserts each key yields `Err(CannotDecrypt { other wallet })` with no messages and no panic (this case previously panicked).
- **app:** `exitRecoveryModeForDevice` unit tests (empty set → no-op; all-decrypt → exits; first-ok/later-fail → no exit + dialog; unavailable → no exit). `flutter analyze` clean, `flutter test` green.

Behavior note: the whole-set-mismatch dialog is the default and is cleanly swappable (an `onMismatch` callback) if you'd rather the background reaction stay silent.

### DIFF
--- a/frostsnap_core/src/coordinator/restoration.rs
+++ b/frostsnap_core/src/coordinator/restoration.rs
@@ -958,18 +958,35 @@ impl FrostCoordinator {
         }
     }
 
-    pub fn has_backups_that_need_to_be_consolidated(&self, device_id: DeviceId) -> bool {
+    /// The pending physical backups `device_id` still needs to consolidate — one
+    /// per share it recovered by entering a physical backup during restoration.
+    /// The unique wallet work set is `.map(|c| c.access_structure_ref)` deduped.
+    pub fn backups_that_need_to_be_consolidated(
+        &self,
+        device_id: DeviceId,
+    ) -> Vec<PendingConsolidation> {
         self.restoration
             .pending_physical_consolidations
             .iter()
-            .any(|consolidation| consolidation.device_id == device_id)
+            .filter(|consolidation| consolidation.device_id == device_id)
+            .cloned()
+            .collect()
     }
 
+    /// Builds the messages telling `device_id` to consolidate its pending physical
+    /// backups, or fails without emitting anything.
+    ///
+    /// A single device can hold pending consolidations across access structures
+    /// from different global-key eras, so `encryption_key` is not guaranteed to
+    /// decrypt every wallet in the set. Each ref is decrypt-checked while the batch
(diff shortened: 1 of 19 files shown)


LLFourn:
Folded into #511: its `keystore-empty-key-upstream` branch is now rebased onto current master (`614f2791`) and carries this exitRecoveryMode fix together with the empty-key work — a single consolidated PR. Closing this as merged-in-spirit.

PR #528, opened 2026-08-05
Name the testnet version in the UI

nickfarrow:
Stating "testnet" alone doesn't say whether the wallet is on Testnet3 or Testnet4, show "Testnet3" wherever the network is displayed.

### DIFF
--- a/frostsnapp/lib/bitcoin_network_ext.dart
+++ b/frostsnapp/lib/bitcoin_network_ext.dart
@@ -1,0 +1,16 @@
+import 'package:frostsnap/src/rust/api/bitcoin.dart';
+
+extension BitcoinNetworkDisplayExt on BitcoinNetwork {
+  /// User-facing network name. Unlike [name], this says which testnet version
+  /// the wallet runs on ("testnet" is Testnet3, not Testnet4). [name] is the
+  /// identity string that round-trips through [BitcoinNetwork.fromString], so
+  /// it can't be changed for display purposes.
+  String get displayName => switch (name()) {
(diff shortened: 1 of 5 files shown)


LLFourn:
Checked the split holds across the whole app. Outside `displayName` itself, `name()` survives at seven
source locations (nine calls — the two comparisons below have one on each side), and every one is an
identity use rather than a display one —

- `settings.dart:77,95` — equality comparisons
- `settings.dart:1168,1176` — the chooser's current `value:` and the per-item `value:`
- `wallet_create.dart:1411,1420` — segment `value` and `selected`

  (those four have to round-trip through `BitcoinNetwork.fromString`)
- `wallet.dart:1209` — `getBlockExplorer`'s URL routing

Every site that previously rendered a raw network name now goes through `displayName`; the one deliberate
exception is `settings.dart`, which still special-cases mainnet as "Bitcoin (BTC)". So the extension's doc
comment is accurate about why `name()` can't simply change, and nothing is left half-converted.

One adjacent, pre-existing bug this brushes against — **not** for this PR, but it's real rather than
theoretical: `SUPPORTED_NETWORKS` is `[Bitcoin, Signet, Testnet, Regtest]`, so regtest is selectable, but
`getBlockExplorer` only has arms for signet/testnet4/testnet and falls through to `_ => .../signet/`. A
regtest wallet therefore links to the **signet** explorer today, right beside its now-correct label. (The
mirror image is also there: `testnet4` has an explorer arm but isn't in `SUPPORTED_NETWORKS`.)

LLFourn (approved):

PR #529, opened 2026-08-06
Block firmware downgrades on the device

LLFourn:
A malicious or modified coordinator can skip the app's `check_upgrade_eligibility`, so downgrade protection is enforced where the image is actually accepted: **on the device**, just before `switch_partition` commits the newly written slot.

## Device enforcement

All four image-driven post-download checks — the image parses, the written digest matches the announced one, the Secure Boot v2 signature verifies (when enabled), and the firmware-only digest is not a superseded release — fold into one decision (`decide_upgrade` in `device/src/ota.rs`) returning `Commit` or `Refuse(reason)` with fixed precedence. Every refusal takes the same path: active slot unchanged, and the device parks on the error screen (warning icon + "Upgrade failed" + reason + "Unplug to try again") until power cycle — the running firmware boots again on replug. Resuming the normal loop instead would leave the coordinator and chain mid-way through an upgrade conversation that never concludes. This replaced three prior `panic!` paths (parse unwrap, digest mismatch, secure boot), whose handler loops forever. `decide_upgrade` drives the flash/SHA/RSA peripherals directly, so it has no host tests — it is compile-verified for both boards; the version data under it is unit-tested in `frostsnap_comms` (superseded-set bounds, v0.0.1's dual digest conventions) and anchored against the real v0.0.1 release binary in the coordinator's existing test.

## Release table

`frostsnap_comms::firmware_version` is the one shared module: `VersionNumber`, `EARLIEST_ACCEPTABLE` (0.4.0 — devices built from this tree refuse anything older, i.e. all previous firmware), and two private digest tables (body digests per release, plus v0.0.1's one-off full-image digest — that era announced it) reached only through accessors that name the convention they mean: `versions_before(v)` (superseded set, body digests) and the `VersionNumber::from_digest` constructor (either convention). Business logic stays in the consuming crates: the device's downgrade gate is a one-line filter over `versions_before` in `ota.rs`; the coordinator keeps using `VersionNumber::from_digest` as before, now backed by the shared list.

A release's digest is appended to the shared list when it should become active — after its artifact is built and signed, since an image can never contain its own digest and doesn't need to. Release CI keeps `--require-known-version` as the check that registration happened.

## Blocked digests

Firmware-only digests, reproduced from the verified PROD-signed release artifacts and pinned by tests against the real v0.0.1 binary:

| version | body digest |
|---|---|
| v0.0.1 | `8f45ae6b72c241a20798acbd3c6d3e54071cae73e335df1785f2d485a915da4c` |
| v0.1.0 | `5ff7bd280b96d645b2c739a6b91bfc6c27197e213302770f1a7180678ca4f720` |
| v0.2.0 | `e432d313c7698b1e8843b10ba95efa8c28e66a5723b966c56c156687d09d16e0` |
| v0.3.0 | `6273dc08ca7c805fa70ac8feeb98c62f7b6fcb337fb1cd8412f24f0d6dda51f7` |

v0.0.1's known full-image digest (`57161f80…`) resolves to its version for lookups but can never appear in the blockable set — by construction and by test.

## Release ritual

Cutting vN: bump `EARLIEST_ACCEPTABLE`, build + sign the artifact, then append its digest to the shared list when it should become active. Interim by design until versions are embedded in firmware binaries.

## Measurements

- Net flash cost of the whole device-side feature on `frontier`: **+2254 bytes** (the removed panic strings largely pay for the decision logic + per-reason refusal screen).

### DIFF
--- a/device/src/frosty_ui.rs
+++ b/device/src/frosty_ui.rs
@@ -245,6 +245,9 @@ impl<'a> UserInteraction for FrostyUi<'a> {
                         FirmwareUpgradeProgress::downloading(progress)
                     }
                     FirmwareUpgradeStatus::Passive => FirmwareUpgradeProgress::passive(),
+                    FirmwareUpgradeStatus::Rejected { reason } => {
+                        FirmwareUpgradeProgress::rejected(&reason.to_string())
+                    }
                 });
 
                 WidgetTree::FirmwareUpgradeProgress { widget, status }
--- a/device/src/frosty_ui.rs
+++ b/device/src/frosty_ui.rs
@@ -19,6 +19,16 @@ use crate::{
     UpstreamConnectionState,
 };
 
+fn refuse_reason_text(reason: crate::ota::RefuseReason) -> &'static str {
+    use crate::ota::RefuseReason;
+    match reason {
+        RefuseReason::UnparseableFirmware => "Firmware invalid",
+        RefuseReason::DigestMismatch => "Download corrupted",
+        RefuseReason::SecureBootInvalid => "Signature invalid",
+        RefuseReason::DowngradeBlocked => "Downgrade blocked",
+    }
+}
+
 // Type alias for the display to match factory
 type DeviceDisplay<'a> = mipidsi::Display<
     display_interface_spi::SPIInterface<
@@ -357,6 +367,9 @@ impl<'a> UserInteraction for FrostyUi<'a> {
                         FirmwareUpgradeProgress::downloading(progress)
                     }
                     FirmwareUpgradeStatus::Passive => FirmwareUpgradeProgress::passive(),
+                    FirmwareUpgradeStatus::Rejected { reason } => {
+                        FirmwareUpgradeProgress::rejected(refuse_reason_text(reason))
+                    }
                 });
 
                 WidgetTree::FirmwareUpgradeProgress { widget, status }
--- a/device/src/ota.rs
+++ b/device/src/ota.rs
@@ -12,10 +12,78 @@ use esp_hal::time::Duration;
 use esp_hal::timer;
 use esp_hal::Blocking;
 use frostsnap_comms::{
-    CommsMisc, DeviceSendBody, Sha256Digest, BAUDRATE, FIRMWARE_NEXT_CHUNK_READY_SIGNAL,
-    FIRMWARE_UPGRADE_CHUNK_LEN,
+    firmware_version, CommsMisc, DeviceSendBody, Sha256Digest, BAUDRATE,
+    FIRMWARE_NEXT_CHUNK_READY_SIGNAL, FIRMWARE_UPGRADE_CHUNK_LEN,
 };
 use nb::block;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum UpgradeVerdict {
+    Commit,
+    Refuse(RefuseReason),
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum RefuseReason {
+    UnparseableFirmware,
+    DigestMismatch,
+    SecureBootInvalid,
+    DowngradeBlocked,
+}
+
+impl core::fmt::Display for RefuseReason {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.write_str(match self {
+            RefuseReason::UnparseableFirmware => "Firmware invalid",
+            RefuseReason::DigestMismatch => "Download corrupted",
+            RefuseReason::SecureBootInvalid => "Signature invalid",
+            RefuseReason::DowngradeBlocked => "Downgrade blocked",
+        })
+    }
+}
+
+/// Decide whether to commit the downloaded upgrade sitting in `partition`,
+/// against the digest the coordinator announced. Everything examined here is
+/// attacker-controlled, so every failure is a verdict, never a panic. When
+/// more than one thing is wrong the reason follows the check order below,
+/// most fundamental first; that only picks the reason shown, never whether it
+/// commits.
+fn decide_upgrade(
+    partition: &EspFlashPartition<'_>,
+    announced_digest: &Sha256Digest,
+    sha: &mut Sha<'_>,
+    rsa: &mut Rsa<Blocking>,
+) -> UpgradeVerdict {
+    let Ok((firmware_size, _)) = partition.firmware_size() else {
+        return UpgradeVerdict::Refuse(RefuseReason::UnparseableFirmware);
+    };
+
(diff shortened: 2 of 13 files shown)


device/src/frosty_ui.rs
```diff
@@ -19,6 +19,16 @@ use crate::{
     UpstreamConnectionState,
 };
 
+fn refuse_reason_text(reason: frostsnap_comms::firmware_upgrade::RefuseReason) -> &'static str {
```
LLFourn: put a Display impl on RefuseReaosn.

device/src/ota.rs
```diff
@@ -12,11 +12,69 @@ use esp_hal::time::Duration;
...
+        return UpgradeVerdict::Refuse(RefuseReason::UnparseableFirmware);
+    };
+
+    let body_digest = partition.sha256_digest(sha, Some(firmware_size));
+    let full_digest = partition.sha256_digest(sha, Some(firmware_and_signature_block_size));
+
+    // A match on either digest convention identifies the intended image (the
+    // full digest is what v0.0.1-era coordinators announce); SHA256 makes an
+    // accidental match impossible.
+    if body_digest != *announced_digest && full_digest != *announced_digest {
```
LLFourn: since we are now stopping downgrades in this PR wouldn't it make sense to simplfy this to only the body digest.

frostsnap_comms/src/firmware_version.rs
```diff
@@ -0,0 +1,162 @@
...
+/// v0.0.1-era devices announced the digest over the FULL image (body plus
+/// signature block), so that one release is also identified by this digest.
+const V0_0_1_FULL_IMAGE_DIGEST: Sha256Digest = Sha256Digest(hex!(
+    "57161f80b41413b1053e272f9c3da8d16ecfce44793345be69f7fe03d93f4eb0"
+));
+
+/// Releases strictly older than `version`, as (body digest, version) — the
+/// superseded set when `version` is [`EARLIEST_ACCEPTABLE`].
+pub fn versions_before(
+    version: VersionNumber,
```
LLFourn: this should be a method on VersionNumber

frostsnap_widgets/src/firmware_upgrade.rs
```diff
@@ -192,6 +211,44 @@ impl FirmwareUpgradeProgress {
...
+            "Unplug to try again".to_string(),
+            Gray4TextStyle::new(&NOTO_SANS_17_REGULAR, PALETTE.text_secondary),
+        );
+
+        let mut column = Column::new((icon, title, status, instruction))
+            .with_main_axis_alignment(MainAxisAlignment::Center)
+            .with_cross_axis_alignment(CrossAxisAlignment::Center);
+        column.set_gap(0, 12);
+        column.set_gap(1, 6);
+        column.set_gap(2, 16);
```
LLFourn: style note, you should use the builder pattern here where you `.push(item).gap(..)`. Makes it much easier to read.

frostsnap_widgets/src/demo_widget.rs
```diff
@@ -583,6 +583,74 @@ macro_rules! demo_widget {
                 let widget = FirmwareUpgradeProgress::passive();
                 $run_macro!(widget);
             }
+            "firmware_upgrade_rejected" => {
```
LLFourn: we can probably drop the demo here. I don't predict we'll need it much!

nickfarrow:
The version comparison is not achieving anything since refusal needs both table membership and `version < EARLIEST_ACCEPTABLE`, but the table only ever holds released versions and `EARLIEST_ACCEPTABLE` is (always?) above all of them. Test `every_known_version_is_below_earliest_acceptable` asserts exactly this. So the gate is essentially a denylist, not a version floor.

Fine in interim until esp-hal 1.0, `Ord` makes sense going forward. But the naming and the doc comments are misleading. A coordinator can `ota` any prod-signed image that is not in the table, regardless of age.

There may be a test missing (still on fsim again?): description says `decide_upgrade` is unit-tested for every reason.

aphanley:
Manually tested on dev secure-boot hardware using a malicious-coordinator harness that bypasses the app-side checks and drives the device directly. Confirmed the device refuses each failure case with the correct on-screen reason — unparseable firmware, digest mismatch, invalid Secure Boot signature, and downgrade blocked — with downgrade-block verified against the real archived v0.0.1–v0.3.0 release binaries and their registered digests.

Commit is to fix on screen artifact where the progress bar isn't fully cleared when the upgrade failed screen is displayed

LGTM!

frostsnap_comms/src/firmware_version.rs
```diff
@@ -0,0 +1,163 @@
...
+            .find(|(d, _)| d == digest)
+            .map(|(_, v)| *v)
+    }
+
+    /// Releases strictly older than `self`, as (body digest, version) — the
+    /// superseded set when `self` is [`EARLIEST_ACCEPTABLE`].
+    pub fn versions_before(self) -> impl Iterator<Item = &'static (Sha256Digest, VersionNumber)> {
+        KNOWN_FIRMWARE_VERSIONS
+            .iter()
+            .filter(move |(_, v)| *v < self)
```
nickfarrow, line 52: The device can only know about firmware existing prior to itself, whenever there is unknown firmware the device accepts it as new (with no ability to compare in-firmware versions until esp-hal 1.0). 

Hence this `<` check does not add anything , and arguably `EARLIEST_ACCEPTABLE` is a misnomer since what we have just resembles a disallowed list. Perhaps some of this fits neatly into esp hal 1.0 once we have it.

LLFourn: The `< ` does add something: It implements the method correctly. The fact that it is always called with the latest one and so is really just returning the full version set is not important to the implementation.

LLFourn:
> The version comparison is not achieving anything since refusal needs both table membership and `version < EARLIEST_ACCEPTABLE`, but the table only ever holds released versions and `EARLIEST_ACCEPTABLE` is (always?) above all of them. Test `every_known_version_is_below_earliest_acceptable` asserts exactly this. So the gate is essentially a denylist, not a version floor.

It is at the moment but earliest acceptable does not always have to be above all of them. It acts as a denylist right now yes.
 
> Fine in interim until esp-hal 1.0, `Ord` makes sense going forward. But the naming and the doc comments are misleading. A coordinator can `ota` any prod-signed image that is not in the table, regardless of age.

The table lists all the prod signed images. If there are other malcious images being signed then no system can protect against that. The malicious key holder sign's verion 99 and it will always be valid. The scope of the verison system is to protect against downgrade to honestly signed images. It does this. 

> There may be a test missing (still on fsim again?): description says `decide_upgrade` is unit-tested for every reason.

Thanks this was stale from an earlier revision.

aphanley (approved):

LLFourn:
#534 closed in favor of a slightly simpler fix added here.

PR #530, opened 2026-08-06
Add security check explainer text and Learn More dialog

### LINKED ISSUE #352: Hint or learn more explainer link on the security check screen
Users often ask "what's this?" on the security check screen.

Maybe some short text:

```
This check guarantees every device (including this phone) has contributed randomness and acted honestly during key generation.

<Learn More>
```

7-falseparfait:
fixes #352

updates the existing explainer text on the security check step to match the wording proposed in the issue... and adds a "learn more" button that opens a small in-app dialog with more detail.

deliberately kept in-app rather than linking out (url_launcher)... per @LLFourn's feedback on the issue, this screen is shown while the user is actively comparing a security code across devices, and navigating away mid-comparison is a ux risk. the new dialog reuses the same backdropfilter + alertdialog pattern already used elsewhere in this file (the 1-of-n wallet warning i believe)... for visual consistency.

i wasn't able to visually test the rendered dialog... reaching this screen requires the full keygen flow with connected physical devices..happy to do a follow-up visual check if that's useful before merging, or if either of you can eyeball it against a real device that'd be very much appreciated.

### DIFF
--- a/frostsnapp/lib/wallet_create.dart
+++ b/frostsnapp/lib/wallet_create.dart
@@ -284,6 +284,7 @@ class WalletCreateController extends ChangeNotifier {
         return BackdropFilter(
           filter: blurFilter,
           child: AlertDialog(
+            scrollable: true,
             title: const Text('Security check'),
             content: DefaultTextStyle(
               style: theme.textTheme.bodyMedium!.copyWith(
--- a/frostsnapp/lib/wallet_create.dart
+++ b/frostsnapp/lib/wallet_create.dart
@@ -201,11 +201,17 @@ class WalletCreateController extends ChangeNotifier {
                 ),
               ),
               Text(
-                'The security check code confirms that all devices have behaved honestly during key generation.',
+                'This check guarantees every device (including this phone) '
+                'has contributed randomness and acted honestly during key '
+                'generation.',
                 textAlign: TextAlign.center,
                 style: theme.textTheme.bodySmall?.copyWith(
                   color: theme.colorScheme.onSurfaceVariant,
                 ),
+              ),
+              TextButton(
+                onPressed: () => _showSecurityCheckInfo(context),
+                child: const Text('Learn more'),
               ),
             ],
           );
@@ -264,6 +270,45 @@ class WalletCreateController extends ChangeNotifier {
 
   void _onCancel() async {
     await coord.cancelProtocol();
+  }
+
+  /// Shows an in-app explainer for the security check. Deliberately an
+  /// in-app dialog rather than an external link (url_launcher): this screen is
+  /// shown while the user is actively comparing a code across devices, and
+  /// navigating away mid-comparison is a bad UX and a security risk.
+  void _showSecurityCheckInfo(BuildContext context) {
+    showDialog<void>(
+      context: context,
+      builder: (dialogContext) {
+        final theme = Theme.of(dialogContext);
+        return BackdropFilter(
+          filter: blurFilter,
+          child: AlertDialog(
+            title: const Text('Security check'),
+            content: DefaultTextStyle(
+              style: theme.textTheme.bodyMedium!.copyWith(
+                color: theme.colorScheme.onSurfaceVariant,
+              ),
+              child: const Text(
+                'Every device contributing to this wallet generates its own '
+                'random data during key generation. The security check '
+                'displays a short code derived from that combined randomness.\n\n'
+                'If the code is identical on every device, it confirms every '
+                'device (including this phone) contributed randomness and no '
+                'device behaved dishonestly or was swapped during setup.\n\n'
+                'If the codes differ, cancel and try again.',
+              ),
+            ),
+            actions: [
+              TextButton(
+                onPressed: () => Navigator.of(dialogContext).pop(),
+                child: const Text('Got it'),
+              ),
+            ],
+          ),
+        );
+      },
+    );
   }
 
   Future<void> resetDeviceNames(Iterable<ConnectedDevice> devices) async {
--- a/frostsnapp/lib/wallet_create.dart
+++ b/frostsnapp/lib/wallet_create.dart
@@ -200,11 +200,17 @@ class WalletCreateController extends ChangeNotifier {
                 ),
               ),
               Text(
-                'The security check code confirms that all devices have behaved honestly during key generation.',
+                'This check guarantees every device (including this phone) '
+                'has contributed randomness and acted honestly during key '
+                'generation.',
                 textAlign: TextAlign.center,
                 style: theme.textTheme.bodySmall?.copyWith(
                   color: theme.colorScheme.onSurfaceVariant,
                 ),
+              ),
+              TextButton(
+                onPressed: () => _showSecurityCheckInfo(context),
+                child: const Text('Learn more'),
               ),
             ],
           );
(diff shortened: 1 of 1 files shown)


LLFourn:
You asked for a visual check — here it is. Drove the full keygen flow in the app-sim with your branch
applied (`bad6f8da` cherry-picked onto our CI-fork master, integrated `d7c0258f`), captured on an Android
emulator with native `screenrecord`.

**It works.** "Choose threshold" → Generate keys → Security Check (code `1c09 b637`, matching on both device
screens) → your new explainer copy → tap **Learn more** → dialog → **Got it** → back on Security Check
unchanged. Contrast is good and the `BackdropFilter` matches the pattern already used elsewhere in that file.

learn more dialog

**One real bug, and it's the reason visual testing was worth doing.** At font scale 1.8 the third paragraph —
*"If the codes differ, cancel and try again."* — silently disappears. No scroll, no ellipsis, no overflow
stripe; I swiped twice on the content and it does not move:

truncation at font scale 1.8

`content:` is a bare `Text` and `AlertDialog` defaults to `scrollable: false`, so the content gets a hard
height constraint and the text is clipped. It's fine at 1.5 and 1.3, and this was on a tall 914 dp screen —
a shorter phone will truncate at a lower scale. The paragraph that vanishes is the only actionable
instruction in the dialog, which makes it the worst one to lose.

One word fixes it:

```dart
AlertDialog(
  scrollable: true,
  ...
)
```

`flutter analyze` is clean on the integrated tree. Note this PR currently has no CI checks running at all, so
that was the only signal either way.

7-falseparfait:
wow...this is reaally helpful and i appreciate it so much and also thank you for taking the time to actually drive it through the emulator,  that's exactly the kind of visual verification i flagged i couldn't do myself.

good catch on the scrollable clipping, that's a real bug and an important one to catch, given it's the one actionable line in the dialog. i will push the fix now (scrollable: true)

7-falseparfait:
pushed the fix... added `scrollable: true` to the `AlertDialog`.

for anyone following along: at larger accessibility font scales, `AlertDialog` 's default `scrollable: false` gives the content a fixed height budget, so text that doesn't fit gets silently clipped rather than scrolling or showing an overflow indicator. that's exactly what was happening at scale 1.8; 
- `scrollable: true` lets the content scroll instead, so nothing gets lost regardless of font scale.

thanks again for catching this; wouldn't have found it without the emulator test.

7-falseparfait:
heads up, one thing to take note of... pushing the `scrollable: true` fix seems to have triggered CI to actually run for the first time on this pr (you'd noted earlier it had no checks running).

two things showed up:

- build device firmware (frontier) was cancelled... likely just GitHub killing the in-progress run when my second commit landed.

- build device firmware (legacy) failed on a stack-check: a 30.3% stack frame in `Frostsigner::recv_coordinator_message` on the embedded device firmware.

this pr only touches `frostsnapp/lib/wallet_create.dart`... so this looks like pre-existing firmware stack pressure unrelated to this change. happy to rebase onto the latest master if that would help confirm.

7-falseparfait:
Rebased onto the latest master, and all CI checks are now green. The rebase cleared the unrelated firmware stack failure. ready for your final approving review whenever you have a moment, @LLFourn. Thanks again for the emulator visual check earlier, really appreciate it.

frostsnapp/lib/wallet_create.dart
```diff
@@ -266,6 +272,46 @@ class WalletCreateController extends ChangeNotifier {
...
+                color: theme.colorScheme.onSurfaceVariant,
+              ),
+              child: const Text(
+                'Every device contributing to this wallet generates its own '
+                'random data during key generation. The security check '
+                'displays a short code derived from that combined randomness.\n\n'
+                'If the code is identical on every device, it confirms every '
+                'device (including this phone) contributed randomness and no '
+                'device behaved dishonestly or was swapped during setup.\n\n'
+                'If the codes differ, cancel and try again.',
```
LLFourn: The copy reads well. Two things in it overclaim or mislead, and since this dialog is where people come to decide whether to trust the check, it's worth getting them exactly right:

1. **"no device … was swapped during setup"** — the check doesn't tell you anything about device identity. It shows that every device and this phone ended up with the same key-generation transcript, so nobody (including the phone relaying the messages) could have altered or dropped a device's contribution. A device swapped for a different, honest one would still produce matching codes.
2. **"If the codes differ, cancel and try again."** — a mismatch shouldn't happen by accident; it means something interfered with the setup. Telling the user to retry straight away with the same phone and devices is the wrong reflex. Better to stop and get help.

Also, "derived from that combined randomness" isn't quite right: the code is computed over the whole certified key generation, not just the random data. Something like:

```suggestion
                'Every device generates its own secret randomness during key '
                'generation and exchanges it with the others through this '
                'phone. The security check code is a fingerprint of that whole '
                'exchange.\n\n'
                'If the code is identical on every device and on this phone, '
                'they all took part in the same key generation, and every '
                'device\'s contribution is in the final wallet key. Nothing, '
                'including this phone, could have changed or left out a '
                'contribution without the codes differing.\n\n'
                'If any code is different, cancel. Something interfered with '
                'the setup, so don\'t retry until you\'ve contacted '
                'support@frostsnap.com.',
```

Happy for the wording to be tightened further; the substance to keep is (a) no identity/swap claim and (b) a mismatch means stop, not retry.

frostsnapp/lib/wallet_create.dart
```diff
@@ -266,6 +272,46 @@ class WalletCreateController extends ChangeNotifier {
     await coord.cancelProtocol();
   }
 
+  /// Shows an in-app explainer for the security check. Deliberately an
+  /// in-app dialog rather than an external link (url_launcher): this screen is
+  /// shown while the user is actively comparing a code across devices, and
+  /// navigating away mid-comparison is a bad UX and a security risk.
```
LLFourn: Small one: the reason for choosing a dialog over a link is exactly what belongs here, thanks for writing it down. The first sentence just restates the function name, though, so it can go:

```suggestion
  /// An in-app dialog rather than an external link: this screen is shown while
  /// the user is comparing a code across devices, and navigating away
  /// mid-comparison is a bad UX and a security risk.
```

frostsnapp/lib/wallet_create.dart
```diff
@@ -266,6 +272,46 @@ class WalletCreateController extends ChangeNotifier {
...
+        return BackdropFilter(
+          filter: blurFilter,
+          child: AlertDialog(
+            scrollable: true,
+            title: const Text('Security check'),
+            content: DefaultTextStyle(
+              style: theme.textTheme.bodyMedium!.copyWith(
+                color: theme.colorScheme.onSurfaceVariant,
+              ),
+              child: const Text(
```
LLFourn: `AlertDialog` has a `contentTextStyle` parameter for this, which drops the extra `DefaultTextStyle` wrapper and the `!` on `bodyMedium`:

```suggestion
            contentTextStyle: theme.textTheme.bodyMedium?.copyWith(
              color: theme.colorScheme.onSurfaceVariant,
            ),
            content: const Text(
```

(with the matching closing `)` for the `DefaultTextStyle` removed below.)

frostsnapp/lib/wallet_create.dart
```diff
@@ -201,12 +201,18 @@ class WalletCreateController extends ChangeNotifier {
                 ),
               ),
               Text(
-                'The security check code confirms that all devices have behaved honestly during key generation.',
+                'This check guarantees every device (including this phone) '
+                'has contributed randomness and acted honestly during key '
+                'generation.',
```
LLFourn: I know this is word-for-word what #352 proposed, so this one's on us rather than you. Looking at it next to the dialog, though, "guarantees … acted honestly" claims more than the check can show: a matching code proves everyone took part in the same key generation with nothing altered in transit. It doesn't prove any particular device behaved honestly internally. Something closer to what it actually shows:

```suggestion
                'Matching codes show that every device (including this phone) '
                'took part in the same key generation, and that no one altered '
                'any device\'s contribution.',
```

frostsnapp/lib/wallet_create.dart
```diff
@@ -201,12 +201,18 @@ class WalletCreateController extends ChangeNotifier {
...
                 textAlign: TextAlign.center,
                 style: theme.textTheme.bodySmall?.copyWith(
                   color: theme.colorScheme.onSurfaceVariant,
                 ),
               ),
+              TextButton(
+                onPressed: () => _showSecurityCheckInfo(context),
+                child: const Text('Learn more'),
+              ),
```
LLFourn: This button exposes a latent bug in `showFullscreenActionDialog` that nothing hit before, because nothing opened a dialog on top of the Security Check screen. Your change isn't wrong, but I think it has to be fixed before this can merge.

When the last device confirms the code (or the devices disconnect), `controller.shouldShow` goes false and the fullscreen dialog pops itself. The pop is latched (`popScheduled = true`) and only runs if the route `isCurrent` (`device_action_fullscreen_dialog.dart:288-303`). If Learn more is open at that moment, the pop is skipped and never retried. Tapping "Got it" then leaves a stale Security Check screen, and the keygen flow in `_beginThresholdKeygen` stays stuck in `removeActionNeeded` → `awaitDismissed`, waiting on a dialog future that never completes, so "Final check" never appears. That's quite likely to happen for real: a user opens Learn more *because* they're in the middle of comparing codes on the devices.

The fix belongs in the helper rather than here: gate scheduling on the route being current, read in `build`, so the builder re-runs when the overlay closes and the route becomes current again:

```dart
builder: (context, _) {
  final isCurrent = ModalRoute.of(context)?.isCurrent ?? false;
  if (!controller.shouldShow && isCurrent && !popScheduled) {
    ...
```

(`ModalRoute.of` registers a dependency that re-notifies when `isCurrent` changes.) If you'd rather keep this PR to the copy, we're happy to land the helper fix separately first. Just say so.

PR #532, opened 2026-08-07
Stabilize and unblock Frontier stack checks

LLFourn:
## What changed

- pin Rust 1.88.0 as the shared firmware and stack-check toolchain
- make the CI stack gate explicit to the production `frontier` board
- split the two oversized measured frames so the existing 25% limit passes
- remove the floating nightly dependency from the stack-check path

## Why

PR #529 did not itself increase these frames, but floating compiler output made an already-tight stack budget nondeterministic. This keeps the check aligned with the shipped compiler and restores CI without raising or silencing the limit. The firmware downgrade behavior is unchanged.

## Validation

`just stack-check stacks --board frontier --max-pct 25`

Passes with 32,076 bytes available; the largest measured frame is 7,872 bytes (24.5%).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,17 +29,21 @@ jobs:
       - name: Add RISC-V GCC to PATH
         run: echo "PATH=$PATH:${GITHUB_WORKSPACE}/riscv/bin" >> $GITHUB_ENV
 
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: riscv32imc-unknown-none-elf
-          components: rustfmt, clippy
+      - name: Install Rust (pinned firmware + stable lint)
+        working-directory: .
+        run: |
+          rustup toolchain install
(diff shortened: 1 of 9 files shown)


nickfarrow (approved):
The `rust-toolchain.toml` is at the repo root so applies to everything including app builds. I think fine for the sake of bug reproducibility, but commit message describes the opposite.

Non-blocking, as predates this PR: this check only measures the largest single frame, never sums a call chain -- so it can't actually tell us the peak stack. Which means `#[inline(never)]` splits relocate stack rather than remove it. `set_workflow` went 8757B → 3376B, but it now calls `build_erase_device` (6608B), so that path is ~9984B vs 8757B before -- the metric improved a lot, actual peak went up ~1.2KB. The keygen split is roughly flat (8532B → ~8400B) so that one's fine.

Improving the check out of scope here, but I thought worth noting for the future.

`RUSTC_BOOTSTRAP` on pinned stable is great.

PR #533, opened 2026-08-10
[app] Make the Android app key work on phones the strict spec locked out

LLFourn:
Replaces #511.

## The bug

A user on a new Poco F7 (Android 16) could not create a wallet at all — keygen died with:

```
PlatformException(KEY_ERROR, Failed to get or create key: Keystore operation failed, null, null)
```

Fresh phone, first run, nothing else going on. The cause is our own `KeyGenParameterSpec`. We ask the
Keystore for considerably more than we need — user-authentication binding, a lock-screen precondition, a
validity window — and on some devices that combination cannot be satisfied at all. The key can then be
neither created nor used, and the app is unusable.

## The fix

We don't need any of it. The thing being protected is not strictly secret, and encrypting it doesn't
change the threat model; the key exists so wallet data at rest isn't trivially readable. So the spec is
relaxed to the minimum that still yields a hardware-backed key — HMAC-SHA256, 256-bit, SHA-256 digest,
nothing else — and the pre-checks are replaced by handling the exceptions that actually occur.

**Backwards compatibility is the constraint this is built around, and it holds exactly.** `KEY_ALIAS`, the
`HmacSHA256` algorithm and the `"frostsnap-hmac-v0"` input are untouched, so every wallet created under
the old strict spec keeps deriving the same key from the same existing key blob. The relaxation applies
only to keys minted from here on.

## The case the old spec was hiding

Clearing the screen lock **destroys** an auth-bound key. `getOrCreateKey` then quietly mints a fresh,
working, *wrong* one — no exception thrown — and every existing wallet silently fails to decrypt. The
pre-checks never caught this; they failed earlier, for different reasons, which is why it looked handled.

Two things now handle it, at different layers.

**At the Keystore boundary**, a `KeyPermanentlyInvalidatedException` is healed in place: the dead alias is
a tombstone that blocks its own replacement and nothing it encrypted is recoverable from it, so it is
deleted and recreated. That failure therefore never surfaces to the app as an exception — it simply
becomes a working key that happens not to match existing wallets, which is the case below.

**Per wallet**, the app asks the coordinator directly whether that wallet's ciphertext still decrypts, and
resolves to exactly one of three outcomes:

| outcome | when | what the user sees |
|---|---|---|
| resolved | the key decrypts this wallet | proceeds normally |
| needs recovery | key is live but does not decrypt **this** wallet | "Wallet needs recovery" — delete it from the app and recover it from its Frostsnap devices |
| unavailable | wallet/access structure unknown, **or** any other key-acquisition failure | an ordinary error, non-destructive |

The third row is the one worth reviewing closely. **"Delete wallet" is deliberately not attached to a
catch-all.** It is offered only on positive evidence of a wrong key — a decryption that demonstrably
fails against a wallet we know about. An unknown wallet and a transient platform failure both land on the
ordinary-error path instead, because the remedy is destructive and irreversible if the diagnosis is wrong.

Every existing-wallet operation goes through that one boundary — showing a backup, checking a backup,
signing a message, sending, and the re-plug auto-exit — so a dead key can no longer surface as silence or
as a misleading error at any of them.

## Supporting changes

- The pending-consolidation `AccessStructureRef` is threaded through the device-list event, so
  consolidation is driven **per access structure** rather than per device — a device belonging to two
  wallets was previously ambiguous.
- `ConsolidateError` distinguishes a missing wallet from a wrong key.
- The now-unreachable lock-screen prompt and a dead send-signing controller are removed.

## Relationship to #511

This replaces #511 rather than dismissing its review. @nickfarrow's review there found two real bugs and
his Pixel 9 testing caught behaviour we would not have; the fixes for both are **incorporated here**. To
be precise about provenance: this branch is a single fresh commit on top of `master`, so his `256ac614`
and `824c4c23` are not ancestors of it — their changes were carried across, not inherited.

The approach changed — #511 kept the strict spec and added fallbacks around it, whereas this removes the
strictness that caused the failure — so a fresh PR was cleaner than rewriting that one.

@nickfarrow also asked there whether users should be warned that the wallet isn't hardware-protected.
That question was about #511's **fallback to an empty key** — if the Keystore couldn't give us a key,
#511 carried on with an empty one, so the wallet was silently unprotected and nobody was told. **This PR
deletes that fallback**, so the case he was warning about cannot arise: there is now either a Keystore
key or an error, never a silent downgrade. No warning is needed for something that can no longer happen.

The narrower version of his question does survive, and is worth stating rather than glossing: we ask for
an `AndroidKeyStore` key but do not request StrongBox or check `isInsideSecureHardware`, so on a device
whose Keystore is software-backed the key is Keystore-protected rather than hardware-protected. That is
unchanged by this PR and is not something it tries to settle.

## Evidence

Recorded by fsim on Android — real virtual devices, a real regtest chain, the app's own UI. In each, the
app key is deleted mid-session so the wallet's ciphertext no longer decrypts, which is the live-but-wrong
key case that throws no exception.

| path | recording |
|---|---|
| Show secret backup | show backup |
| Sign message | sign message |
| Send (funded regtest wallet) | send |
| Re-plug auto-exit | consolidation |

**The negative direction matters as much as the positive.** After the key is replaced, a *newly created*
wallet opens its backup normally while the *pre-existing* wallet still routes to recovery — in the same
app, same session. That is what distinguishes "the boundary evaluates each wallet" from "the app panics
because the key changed":

negative

And the case the relaxation exists for — clearing the screen lock no longer destroys the key, so the
wallet keeps working:

pin clear survives

## Status

Ready for review. One commit, **signed** (`%G? = G`). All 10 checks are green on the **signed head**
`607ab838` — both here and on the fork mirror
([LLFourn/frostsnap#11](https://github.com/LLFourn/frostsnap/pull/11)).


frostsnap_core/src/coordinator/restoration.rs
```diff
@@ -958,39 +958,54 @@ impl FrostCoordinator {
...
+        )?;
+        let root_shared_key = frost_key
+            .complete_key
+            .root_shared_key(access_structure_ref.access_structure_id, encryption_key)
+            .ok_or(ConsolidateError::CannotDecrypt {
+                access_structure_ref,
+            })?;
+
+        let mut messages = vec![];
+        for consolidation in self
```
LLFourn: This should be unique since a device can only have one share in an access structure. Should just be a `.find` not `.filter`

frostsnap_core/tests/restoration.rs
```diff
@@ -546,3 +552,148 @@ fn delete_share_and_recover_then_sign() {
...
+        .consolidate_pending_physical_backups(device_id, asr, wrong_key)
+        .unwrap_err();
+    assert!(
+        matches!(err, ConsolidateError::CannotDecrypt { access_structure_ref } if access_structure_ref == asr),
+        "an existing wallet that won't decrypt must be CannotDecrypt, got {err:?}"
+    );
+}
+
+#[test]
+fn per_ref_consolidation_uses_each_wallets_own_key() {
```
LLFourn, line 625: Not sure if this test is worth it. The branches we're testing are not that complicated. The other two tests over it.

frostsnap_core/src/coordinator/restoration.rs
LLFourn: Done in `53999aad`.

Agreed, and the type makes the case a bit stronger than "should be unique": `PendingConsolidation` sits in a `BTreeSet` keyed on `(device_id, access_structure_ref, share_index)`, so two entries agreeing on the first two are *structurally* representable even though the domain forbids them. With `.filter` that would have silently emitted a second `Consolidate` for one device/access-structure pair; `.find` puts the one-share invariant at the point that depends on it.

`cargo test -p frostsnap_core --release` green (11 binaries, incl. `consolidate_reports_missing_wallet_without_blaming_the_key` and `consolidate_reports_wrong_key`); `cargo fmt --check` clean.

frostsnapp/lib/wallet_send_controllers.dart
```diff
@@ -336,174 +325,3 @@ class AmountInput extends StatelessWidget {
     );
   }
 }
-
-class DeviceModel {
```
LLFourn, line 340: note this is all just dead code. It was better to delete it than update it to use the new API.

frostsnapp/rust/src/device_list.rs
```diff
@@ -175,11 +181,17 @@ impl DeviceList {
...
+                pending_consolidations: vec![],
             });
         }
     }
 
-    pub fn set_recovery_mode(&mut self, id: DeviceId, recovery_mode: bool) {
+    pub fn set_recovery_mode(
+        &mut self,
+        id: DeviceId,
+        recovery_mode: bool,
```
LLFourn: recovery_mode is not orthogonal to pending_consolidations -- you can only have them if you're in recovery mode right? Perhaps this should become a sum type mirroed into dart.

frostsnap_core/tests/restoration.rs
LLFourn: It does cover something the other two can't, though the cost point is fair.

`consolidate_reports_missing_wallet_without_blaming_the_key` and `consolidate_reports_wrong_key` each build a **single** wallet from one `Run::start_after_keygen_and_nonces`, so between them they cover one error branch each on one wallet.

This one is the only test with **two wallets under different encryption keys on the same device**. It asserts A's key opens A, A's key against B fails as `CannotDecrypt` **for B alone rather than as a batch failure**, and B's own key opens B. That's the invariant the `AccessStructureRef` threading exists for — consolidation driven per access structure, not per device — and no single-wallet test can catch a regression that batches across wallets or reaches for the wrong wallet's key. It's also the case that actually happens in the field: a wallet from before the app key was invalidated sitting beside one from after.

You're right that it's the heaviest setup in the file (a second full keygen against a second `TestEnv`). Happy to drop it if you still think that's not worth the coverage — just say and I'll remove it.

frostsnapp/rust/src/device_list.rs
LLFourn: Good catch — done in `27561f5e`.

```rust
pub enum RecoveryMode {
    Off,
    On { pending_consolidations: Vec<AccessStructureRef> },
}
```

It lives on `ConnectedDevice`, so `DeviceListChange` loses its parallel `pending_consolidations` field entirely — the device carries it — and `set_recovery_mode` takes one value instead of a bool beside a list that could disagree with it. FRB mirrors it as a Dart sealed class.

The tell that you were right: `main.dart` was reconstructing the invariant by hand as a three-part conjunction —

```dart
if (change.kind == DeviceListChangeKind.recoveryMode &&
    change.device.recoveryMode &&
    pendingConsolidations.isNotEmpty)
```

— which is now `recoveryMode is RecoveryMode_On && recoveryMode.pendingConsolidations.isNotEmpty`. The consumer having to re-derive it is usually the sign the type wasn't carrying it.

Left the transition guard as-is per your "ok for now", so a device already in recovery that gains consolidations still emits nothing — it compares on/off, not the whole value. There's a comment at the site naming that as a decision rather than an oversight, since an obviously-too-narrow comparison with no explanation invites someone to "fix" it blind. Widening it to `!=` on the whole enum is a one-liner whenever you want it.

`just lint-app` and `just test-ordinary --release --all-features --locked` both green.

frostsnapp/rust/src/api/device_list.rs
```diff
@@ -33,13 +33,28 @@ pub struct DeviceListState {
...
+/// Whether a device is in recovery mode, and if so what it still needs consolidated.
+/// The two are not independent — pending consolidations only exist in recovery — so they
+/// are one value rather than a bool beside a list that could disagree with it.
+#[derive(Clone, Debug, PartialEq)]
+pub enum RecoveryMode {
+    Off,
+    On {
+        /// Wallets this device already holds physical backups for that still need
+        /// consolidating. The background auto-exit asks for each wallet's key separately.
+        /// Non-empty only on the re-plug that discovers them; the recovery-mode entries
```
LLFourn: this reasoning about why this being empty is a valid state needs to be way more clear. It seems like an architecure issue and this comment should loudnly flag it as such.

frostsnap_core/src/coordinator/restoration.rs
LLFourn: Follow-up on your `Option` suggestion — done in `98fe786c`. `consolidate_pending_physical_backups` now returns `Result<Option<CoordinatorSend>, ConsolidateError>`.

You were right that the `Vec` was an `Option` in costume: it hid the at-most-one the `.find` had just established, leaving callers to infer it from a collection that happens never to exceed one element.

It paid for itself downstream too. The sole production caller had:

```rust
if msgs.is_empty() { return; }
let expected = msgs.len();
...
done += 1;
done >= expected
```

The emptiness check is now an `Ok(None) => return` arm, and `done`/`expected` — a counter that could only ever reach one — collapses to a plain `matches!`. Tests needed no change, since `Run::extend` takes `impl IntoIterator` and `Option` is one.

frostsnapp/rust/src/device_list.rs
LLFourn: On "why not just compare the whole value, what would it cost" — nothing worth keeping the special case for, so it's in `98fe786c`:

```rust
if connected_device.recovery_mode != recovery_mode {
```

Cost: a `Vec` compare of at most one element. Behaviour: the auto-exit now fires in the case it silently dropped; equal values still don't emit so nothing re-fires; the one new emission (`On{[a]}` → `On{[]}`) carries an empty set that Dart's `isNotEmpty` guard ignores.

The thing I should have spotted before you asked: preserving the drop needed a two-`matches!` dance, so the fixed version is *shorter* than the version defending the bug. When keeping a known defect costs more code than fixing it, that's the tell. I'd read your "that's ok for now" as "don't change behaviour" when it meant "don't let it block you".

`just lint-app` and `just test-ordinary --release --all-features --locked` both green.

frostsnapp/android/app/src/main/kotlin/com/frostsnap/SecureKeyManager.kt
```diff
@@ -77,103 +73,113 @@ class SecureKeyManager : FlutterPlugin, MethodCallHandler, ActivityAware, Plugin
...
+    // both entry points classify failures the same way.
+    private fun deliverKeyBytes(result: Result?, allowAuthPrompt: Boolean) {
+        val keyBytes =
             try {
-                keyGenerator.init(buildKeySpec(useStrongBox = true))
-                keyGenerator.generateKey()
-                Log.i(TAG, "Key created successfully with StrongBox backing")
+                getKeyBytes()
+            } catch (e: UserNotAuthenticatedException) {
+                // Only pre-existing keys are auth-bound; newly created ones are not.
```
LLFourn: There should be a comment here that this branch is probably unreachable. THe previous auth parameters was for 64 years so there's no way it would expire and cause this. This is just left here in case.

frostsnapp/rust/src/api/device_list.rs
LLFourn: You're right, and it's a model bug rather than a comment bug — fixed in `3c62d540` instead of documented.

I traced the convention to the call sites and it's total: `coordinator.rs:202` (re-plug) is guarded by `if !pending.is_empty()` so it *always* passes non-empty, and both mid-restoration sites *always* pass `vec![]`. So the empty vec was silently encoding a **second reason for being in recovery** — "a restoration flow owns this device" — wearing the same variant as "re-plugged with work to do".

```rust
pub enum RecoveryMode {
    Off,
    /// A restoration this app is driving put it there. That flow owns the device and takes it
    /// back out; the background auto-exit must leave it alone.
    InFlow,
    /// A re-plugged device turned out to already hold physical backups needing consolidation.
    NeedsConsolidation { pending: Vec<AccessStructureRef> },
}
```

The tell that it was the right cut: Dart's guard lost its `.isNotEmpty` check entirely — `recoveryMode is RecoveryMode_NeedsConsolidation` is now the whole condition. That's the second time in this PR that fixing the type deleted a check the consumer was using to re-derive the invariant by hand.

One residual, flagged rather than explained away: an empty `pending` inside `NeedsConsolidation` is still structurally representable, since there's no non-empty vector type in this tree. The doc says it's nonsense rather than a second meaning, and notes the sole constructor guards it — that's the one gap left in the model.

`just gen`, `just lint-app`, `just test-ordinary --release --all-features --locked` all green.

LLFourn: Reverted — you were right, and `InFlow` is gone as of `a2763d66`:

```rust
pub enum RecoveryMode {
    Off,
    On { pending: Vec<AccessStructureRef> },
}
```

My argument for the variant was that empty "silently encodes a second reason". That was overstated: `pending: []` doesn't code for mid-flow, and nothing consumes the *why* anyway — `signing.rs:128` only asks whether the mode is `Ready`, and the auto-exit only asks whether there's work. So `InFlow` was a state nobody read, and the `.isNotEmpty` I'd called "the consumer re-deriving an invariant" is just the business rule, which belongs there.

What you asked for was that the reasoning be clear, and the comment now says the narrow true thing rather than the confident wrong one I first replaced it with: `pending` is the **deferred** `pending_physical_consolidations` work from *finished* restorations, so empty means only that there's nothing for the background auto-exit to pick up — an active flow can still have consolidation in flight via `tmp_waiting_consolidate`.

frostsnapp/android/app/src/main/kotlin/com/frostsnap/SecureKeyManager.kt
LLFourn: Added in `a2763d66`.

Checked the figure rather than repeating it: the pre-relaxation spec set `setUserAuthenticationValidityDurationSeconds(Integer.MAX_VALUE)` = 2147483647 s, which is **~68 years** rather than 64 — same conclusion, that window never lapses in practice.

The comment now says the branch is kept defensively, that only keys minted under the old strict spec are auth-bound at all, and that keys minted since the relaxation cannot raise it.

aphanley:
Manually tested on my phone to make sure no new errors surfaced, but my phone was not affected by the prior issue this PR is resolving

nickfarrow (approved):

PR #534, opened 2026-08-10
Erase the whole OTA partition before download

nickfarrow:
The erase loop compared the count of erased sectors against `n_sectors() - 1`, and so the last sector of the OTA partition was never erased and kept whatever the previous image left there.

I don't think any of our past firmware has been large enough to write anything to the end partition.

### DIFF
--- a/device/src/ota.rs
+++ b/device/src/ota.rs
@@ -305,21 +305,21 @@ impl FirmwareUpgradeMode<'_> {
                     }
                     State::Erase { seq } => {
                         let mut finished = false;
-                        let last_sector_index = partition.n_sectors() - 1;
+                        let n_sectors = partition.n_sectors();
                         /// So we erase multiple sectors poll (otherwise it's slow).
                         const ERASE_CHUNK_SIZE: usize = 1;
                         for _ in 0..ERASE_CHUNK_SIZE {
                             partition.erase_sector(*seq).expect("must erase sector");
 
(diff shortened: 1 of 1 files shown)


LLFourn:
Thanks — good catch, and your diagnosis is right: comparing the incremented `seq` (a count) against `last_sector_index` stopped the loop a sector early, leaving the last sector of the destination partition holding whatever the previous image wrote.

Taken on `firmware-downgrade-block` in 08fa3eb5 with a `Co-authored-by:` for you, since that branch already rewrites the surrounding code and #529 would have conflicted with this one either way.

Landed it as moving the increment *after* the check rather than changing the bound — that keeps `seq` as the index of the sector just erased, so it stays capped at `last_sector_index` and the existing progress fraction stays bounded at 1.0 without needing the denominator changed too. One line.

One clarification worth recording: the stale-signature-sector consequence fails **closed** — a leftover magic makes `find_signature_sector` point at the wrong sector so a *good* image gets rejected; making a *bad* image pass would still require a valid PROD signature in that sector. So it's a false-rejection hazard rather than a security hole, consistent with your "not reachable today".

Also apologies — I force-pushed `firmware-downgrade-block` while your PR was based on it, which is what put this into CONFLICTING. That was my mistake; I'll stop rewriting that branch now that it has other work stacked on it.

Closing in favour of 08fa3eb5.

PR #541, opened 2026-08-14
[app] Prompt to upgrade a connected device's firmware

LLFourn:
A device could sit on stale firmware indefinitely. The only way to discover an upgrade was the device list's "Upgrade N devices" button, which you had to go looking for.

This floats a prompt over the wallet page when a connected device reports it can be upgraded, offering the existing upgrade flow or a dismissal that lasts the session.

## Placement is the design

The prompt lives in the **wallet page's own widget tree**, not on the root navigator and not in an overlay.

Every fullscreen device dialog is a root-navigator route, so page content sits beneath all of them by construction. The prompt *cannot* cover a device prompt, rather than being ordered not to — there is no z-order rule for a later change to get wrong. An `OverlayEntry` would have done the opposite.

It reads its count from `DeviceActionUpgradeController`, the `ChangeNotifier` whose job that already is, rather than watching the device list a second time. That also means it observes connections continuously, so a device plugged in halfway through signing is noticed like any other.

## An approach that was tried and abandoned

The first attempt gated each action on the firmware question — asking before check-backup, display-backup, verify-address and signing. Worth recording why it didn't survive:

- The coordinator dispatches protocol messages the moment a call is made (`tell_device_to_check_backup` calls `start_protocol` inside its own body), so the question had to be asked *before* every call site sent anything. That makes it a line each site must remember.
- It sampled the device list **once**, at action start. Signing and address verification are exactly the flows where devices arrive one at a time *afterwards*, so the two that most needed it were structurally uncovered.
- Covering them properly needs the app, not the coordinator, to decide when per-device messages go out. Signing already works that way (`connected_but_need_request` is state the app acts on), but `VerifyAddressProtocol::connected` queues its own send, so verify would need a coordinator change to participate at all.

A prompt that watches the device list interposes on nothing and needs none of that.

## Two deliberate choices

**It sits over the page's own `SliverAppBar`, drawer button included.** Dismissing to reach the tray is accepted as reasonable pressure to deal with the upgrade. The spec-correct M3 alternative is a banner, which is *inline* rather than floating — that is what would stop it covering anything — but the app bar it must sit below is a sliver inside each page's `CustomScrollView`, so it costs an insertion per page instead of the single one this needs. A pinned-sliver variant was built and compared before settling here.

**The drawer background changes too**, because it is the same defect a level up: `NavigationDrawer` was passed `colorScheme.surface`, the token the page body already uses, so the tray and the content it slid over were the same colour. That was an override away from Flutter's own M3 default of `surfaceContainerLow`.

## Testing

No new tests. Nothing here can be asserted without the device stack, and mocked device tests would only assert the mock.

`flutter analyze`, `just dart-format-check-app` and the existing 144-test suite are clean. Behaviour was driven by hand against two devices on stale firmware — prompt appears on connect, Upgrade runs the existing flow, it clears itself once the devices report the new digest, dismissal lasts the session, and it is completely hidden behind a fullscreen device dialog.

### DIFF
--- a/frostsnapp/lib/firmware_upgrade_bubble.dart
+++ b/frostsnapp/lib/firmware_upgrade_bubble.dart
@@ -1,0 +1,133 @@
+import 'package:flutter/material.dart';
+import 'package:frostsnap/device_action_upgrade.dart';
+
+/// Session-scoped: dismissing silences it until the app restarts. A notifier
+/// rather than a bare bool so the scope matches the storage — every mounted
+/// prompt reacts, not only the one whose button was pressed.
+final _dismissed = ValueNotifier(false);
+
+/// Belongs in the page's own widget tree rather than on the root navigator or
+/// an overlay — every fullscreen device dialog is a root-navigator route, so
+/// living below them is what keeps this from ever covering a device prompt.
+/// Ordering it explicitly would be a rule to get wrong; this cannot be.
+class FirmwareUpgradeBubble extends StatefulWidget {
+  const FirmwareUpgradeBubble({super.key});
+
+  @override
+  State<FirmwareUpgradeBubble> createState() => _FirmwareUpgradeBubbleState();
+}
+
+class _FirmwareUpgradeBubbleState extends State<FirmwareUpgradeBubble> {
+  static const _maxWidth = 460.0;
+
+  final _upgrades = DeviceActionUpgradeController();
+  bool _upgrading = false;
+
+  @override
+  void dispose() {
+    _upgrades.dispose();
+    super.dispose();
+  }
+
+  Future<void> _upgrade() async {
+    setState(() => _upgrading = true);
+    try {
+      await _upgrades.run(context);
+    } finally {
+      if (mounted) setState(() => _upgrading = false);
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return Positioned(
+      top: 0,
+      left: 0,
+      right: 0,
+      // Deliberately over the page's own `SliverAppBar`, drawer button
+      // included: LLFourn's call, on the grounds that having to dismiss it to
+      // reach the tray is acceptable pressure to deal with the upgrade. Note
+      // it is the drawer button it lands on, so dismissing is the only way
+      // back to the tray while it is up.
+      child: SafeArea(
+        bottom: false,
+        child: Align(
+          alignment: Alignment.topCenter,
+          child: ConstrainedBox(
+            constraints: const BoxConstraints(maxWidth: _maxWidth),
+            child: ListenableBuilder(
+              listenable: Listenable.merge([_upgrades, _dismissed]),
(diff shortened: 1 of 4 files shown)


nickfarrow:
Tested on desktop, works well for me. 

Sometimes prompting when the device can not upgrade (probably never seen by user.. optional):
https://github.com/LLFourn/frostsnap/pull/15

If someone has tested how the dialog sits on mobile, then good to go.

LLFourn:
Tested on mobile — Android emulator, 1080×2400. Answering @nickfarrow's *"if someone has tested how the dialog sits on mobile, then good to go."*

**No code blockers.** The one correction below is to the PR description, not the code.

**Placement holds, and it holds differently per surface:** a partial bottom sheet does **not** cover the pill, while a fullscreen root-navigator route covers it completely and it fades back on cancel. That is what the widget-tree placement buys, rather than a z-order rule a later change could get wrong.

**Appears on connect** — the pill slides in over the app bar.

appears on connect

**Hidden behind a device dialog** — visible above the Receive *sheet*, fully covered by the fullscreen **Verify address on device** route, back again on cancel.

hidden behind a device dialog

**Upgrade, then self-clear** — confirm on device, progress, Upgrade Successful, pill gone. It really clears rather than hiding: afterwards Connected Devices no longer offers `Upgrade N device`, so the device re-announced.

upgrade and self-clear

**Dismissal** — ⚠️ shows the *old* behaviour, superseded by `b9711d78`. At the time of recording a replug did not bring the pill back; it now re-arms on any device arrival. Note the clip does show that dismissal suppresses the prompt without discarding the state — Connected Devices still offers `Upgrade 1 device` while dismissed.

dismissal

**Drawer** — `#171c1e` against the page's `#0f1416`.

drawer colour

**Arrives mid-signing** — signing already under way at `Signatures Needed 1/2` when a stale device is plugged in, and the pill appears. Reading the count from `DeviceActionUpgradeController` means connections are observed continuously; the action-gated attempt sampled the device list once at action start, so signing and address verification — the flows where devices arrive one at a time — were the two least covered.

arrives mid-signing

⚠️ **That arriving device is a bystander**, added after keygen, so it is not one of the two signers and nothing prompts it to sign. The case that actually motivates the feature — a *signer* arriving mid-signing, where its signing prompt and the pill have to coexist — is untested.

## Correction for the PR description

It says the drawer change is no longer passing a colour, letting M3's default apply. `wallet.dart:662` explicitly passes `theme.colorScheme.surfaceContainerLow` (and the same token backs the wide-layout rail at `:693`). Same result, different mechanism.

## Provenance

Recorded on the fork's CI stack plus this PR at `3233de34` and the fork's now-closed #15, so the footage predates `3764b5c6`. That commit changes what `needs_firmware_upgrade()` selects, which only differs for a `CannotUpgrade` device — no scenario here connects one, so nothing above renders differently on the current head. Visual claims are read off extracted frames rather than driver assertions.

nickfarrow (approved):

PR #542, opened 2026-08-15
Make the fee display pure: plan is a value, commit is the only mutation

LLFourn:
## The bug

`TxState::fee()` ran the full `send_to` on every rebuild of the send flow's signer page — and `send_to` fused answering *"what would this cost?"* with **reserving a change address**. Every repaint revealed a change index, persisted the reveal, marked it used, and ran BnB coin selection (~360µs + a sqlite transaction) synchronously on the UI isolate. A normal send burned several indices of which one reached the chain, so consecutive on-chain change outputs were separated by each session's repaint count — feeding the gap toward the sync's lookahead limit, the raw material for visible-but-unspendable funds.

Reproduced end-to-end on regtest (fork-side sim harness): the wallet's first send put its change at internal index 2 instead of 0.

## The fix

The plan becomes an explicit value with an owner instead of hidden builder state:

- **`frostsnap_coordinator/src/bitcoin/send.rs`** (new): `SendPlan` — each input pinned as its keychain outpoint (derivation path and outpoint, its immutable identity) plus the outputs it was selected for and the fee the selection fixed; private fields, no `Clone`, so the fee it reports is the fee the committed transaction pays. `plan_send` reads and reserves nothing; `commit_send` (plan → signable template) is the wallet's single change-allocation point — it re-canonicalizes only the plan's own outpoints, since spendability is the one pinned fact that can go stale, failing only when a planned input is no longer spendable. `send_to` is **deleted** so the fused behavior cannot return.
- Dart holds the plan: confirming the amount calls `BuildTxState::try_finish()`, the review fee row reads `plan.fee()` — a field read that cannot reach the wallet — and "Sign transaction" passes the plan back through `SuperWallet::commit_send`. A dead commit returns the user to re-confirm the amount, building a fresh plan whose fee they see before anything commits.

## Tests

Unit tests pin: planning reserves nothing; commit pays the planned fee from the first change index; consecutive sends use consecutive indices; send-max allocates nothing; a stale plan errors without allocating. End-to-end regtest drives (change-lands-on-first-change-address; full 2-of-3 send) run green on the fork and follow upstream once the sim harness does.

Stacked series 1/3, followed by "Balance and Send Max agree" and "Change reservations are a computed view".

### DIFF
--- a/frostsnap_coordinator/src/bitcoin.rs
+++ b/frostsnap_coordinator/src/bitcoin.rs
@@ -2,6 +2,7 @@ pub mod chain_sync;
 mod handler_state;
 pub mod outgoing;
 pub mod psbt;
+pub mod send;
 pub mod status_tracker;
 pub mod tofu;
 pub mod wallet;
--- a/frostsnap_coordinator/src/bitcoin/send.rs
+++ b/frostsnap_coordinator/src/bitcoin/send.rs
@@ -1,0 +1,561 @@
+//! Building a spend, as the two stages with different rights over the wallet: planning may only
+//! read, committing is the single place a change address comes from. The boundary value
+//! [`SendPlan`] is a finished coin selection that has reserved nothing, so a caller can plan (and
+//! re-plan) as often as it likes — a fee display bound to a plan cannot move the wallet's
+//! keychain indices.
+
+use super::wallet::CoordSuperWallet;
+use anyhow::{anyhow, Result};
+use bdk_chain::{
+    bitcoin::{self, Amount, OutPoint, TxOut},
+    CanonicalizationParams,
+};
+use bdk_coin_select::{
+    metrics, Candidate, ChangePolicy, CoinSelector, DrainWeights, FeeRate, Target, TargetFee,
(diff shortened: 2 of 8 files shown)


LLFourn:
Updated: `commit_send` now carries **no reservation at all** — the RAM-only `mark_used` is dropped rather than inherited. It died on restart regardless while burning cancelled sends' indices in-session; until a committed tx reaches the graph, a repeat send may pick the same change index (reuse between your own in-flight txs — rare and funds-safe). Pinned by `unbroadcast_sends_reuse_the_change_index`; the reservations PR (#544) closes the window properly.

nickfarrow:
Reformatted commit message of f15560f.


e0eb26d: the signer page's fee comes from the plan but its amount was still a live wallet query, so under `Max` a deposit landing while you pick devices showed more than the transaction pays. Now both come from the plan.

Wallet has 1 BTC, you tick Max and confirm, then 5 BTC arrives while you pick signing devices:

|                  | Summary shows | Transaction sends             |
| ---------------- | ------------- | ----------------------------- |
| master           | 6 BTC         | 6 BTC (late deposit swept in) |
| this PR          | 6 BTC         | 1 BTC                         |
| with this commit | 1 BTC         | 1 BTC                         |

nickfarrow (approved):

PR #544, opened 2026-08-15
Change reservations are a computed view of live signing sessions, not wallet state

LLFourn:
**Stacked on #542 + the balance/Send Max PR** (contains their commits until they merge; review the tip commit only).

## The bug

`commit_send` reserved change with `next_unused_spk` + `mark_used` — a RAM-only bit (`keychain_txout`'s changeset cannot carry it) with a two-sided lifecycle nobody completed: one mark call in the tree, zero unmarks. A cancelled send's change index stayed burned until an app restart happened to recycle it; a pending send's reservation died on restart while its fully-signed session survived, letting two live templates pay the same change address.

## The fix

`mark_used` is deleted and nothing stateful replaces it — the reservation *is* the session:

- `commit_send(plan, reserved)` allocates the lowest revealed-unused index outside the caller's view, revealing fresh only when nothing passes (a fresh reveal can't collide: reservations come from committed templates, whose indices sit at or below the frontier).
- The knowledge lives where its facts live: `WireSignTask::reserved_change_indices(master_appkey, account)` — the task knows its contents, the account explicit because internal keychain is really (account, keychain) — and `FrostCoordinator::reserved_change_indices` builds the **active ∪ finished** union beside the session maps that define it (a fully signed session leaves the active map while its tx may still be unbroadcast). The api is one line naming `BitcoinAccount::default()` at the single visible call site.
- Two lifetimes, stated honestly: graph-used indices are durable from indexing; session reservations last until cancel/forget (core has no chain visibility); post-broadcast overlap is harmless.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/send.rs
+++ b/frostsnap_coordinator/src/bitcoin/send.rs
@@ -8,6 +8,7 @@ use super::wallet::CoordSuperWallet;
 use anyhow::{anyhow, Result};
 use bdk_chain::{
     bitcoin::{self, Amount, OutPoint, TxOut},
+    indexer::keychain_txout,
     CanonicalizationParams,
 };
 use bdk_coin_select::{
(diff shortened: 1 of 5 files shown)


LLFourn:
Updated after the #542 change: `mark_used` is already gone there, so this PR's role sharpens — it supplies the *real* reservations (active ∪ finished session view) over the accepted no-reservation baseline, rather than replacing the RAM mark.

nickfarrow (approved):

PR #545, opened 2026-08-16
Force-spend coins stranded past a risky gap

LLFourn:
Stacked on #544 (contains #542–#544; review the tip commit).

A restore discovers history by crawling: from last used index p it derives p+1..=p+20 (BIP44’s gap limit, the strictest convention in common use) and **stalls at the first window with no history**. Any coin above the stall point is invisible to it — including a whole cluster of coins above one wide gap, however dense their own spacing. Letting such coins sit there makes recovery silently lossy.

## Change

- Every outgoing plan consolidates them: `plan_send` **simulates that crawl per keychain** (reach starts at the baseline window; each reachable used index u extends it through u + `RISKY_GAP`; unreachable history extends nothing — recovery is transitive) and **preselects every unreachable utxo in the `CoinSelector`** before the bnb run. Change absorbs the overfunding and lands at a low index — which is what moves the value back inside every scanner's reach. Send-max needs nothing (it already spends everything).
- "Used" is the indexer's `is_used` semantics — spent history resets a restore's gap counter, so it resets ours.
- Defending 20 covers every wider window (bdk’s default 25 included) a fortiori. The threshold sits deliberately below the wallet's own lookahead of 50: a coin past a bigger gap never enters the UTXO set at all, so this is proactive protection for coins the wallet can still see, not rescue for coins it cannot.
- Preselection composes with `bdk_coin_select` 0.4.1 by construction: the selection set holds original candidate indices, the metric's sort only permutes iteration order, and bnb branches only over unselected candidates; the `select_until_target_met` fallback keeps the base selection.

## Tests

Boundary regressions on the existing send fixtures: a gap of 19 forces nothing, exactly 20 forces, a spent-through index chains the windows, and a cluster above one gap is forced whole. Verified end to end on the fork's regtest harness: an ordinary small send sweeps far external and internal coins staged from the exported descriptor, with the change landing at internal index 0.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/send.rs
+++ b/frostsnap_coordinator/src/bitcoin/send.rs
@@ -4,7 +4,7 @@
 //! re-plan) as often as it likes — a fee display bound to a plan cannot move the wallet's
 //! keychain indices.
 
-use super::wallet::CoordSuperWallet;
+use super::wallet::{CoordSuperWallet, KeychainId};
 use anyhow::{anyhow, Result};
 use bdk_chain::{
     bitcoin::{self, Amount, OutPoint, TxOut},
@@ -20,8 +20,16 @@ use frostsnap_core::{
     tweak::{BitcoinAccountKeychain, BitcoinBip32Path},
     MasterAppkey,
 };
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use tracing::{event, Level};
+
+/// A coin a restore with this scan window cannot reach gets force-spent by every plan. From
+/// last used index p a restore derives p+1..=p+window and stalls where no history appears; 20
+/// is BIP44's gap limit, the strictest convention in common use, and defending it covers every
+/// wider window too (bdk's default 25 included) — whatever a 20-window crawl reaches, a wider
+/// one reaches a fortiori. It must stay below the wallet's own lookahead — a coin past a bigger
+/// gap never enters the UTXO set, so there is nothing to force.
+const RISKY_GAP: u32 = 20;
 
 /// A finished coin selection that has reserved nothing.
 ///
@@ -109,22 +117,22 @@ impl CoordSuperWallet {
             target_outputs
         };
 
-        let utxos: Vec<(_, bdk_chain::FullTxOut<_>)> = self
-            .tx_graph
-            .graph()
-            .filter_chain_unspents(
-                self.chain.as_ref(),
-                self.chain.tip().block_id(),
-                CanonicalizationParams::default(),
-                self.tx_graph
-                    .index
-                    .keychain_outpoints_in_range(Self::key_index_range(master_appkey)),
-            )
-            .collect();
+        let (keychain_indices, utxos): (Vec<(KeychainId, u32)>, Vec<bdk_chain::FullTxOut<_>>) =
+            self.tx_graph
+                .graph()
+                .filter_chain_unspents(
+                    self.chain.as_ref(),
+                    self.chain.tip().block_id(),
+                    CanonicalizationParams::default(),
+                    self.tx_graph
+                        .index
+                        .keychain_outpoints_in_range(Self::key_index_range(master_appkey)),
+                )
+                .unzip();
 
         let candidates = utxos
             .iter()
-            .map(|(_path, utxo)| Candidate {
+            .map(|utxo| Candidate {
                 input_count: 1,
                 value: utxo.txout.value.to_sat(),
                 weight: TR_KEYSPEND_TXIN_WEIGHT,
@@ -173,6 +181,9 @@ impl CoordSuperWallet {
         );
 
         let mut cs = CoinSelector::new(&candidates);
+        for position in self.gap_stranded(master_appkey, &keychain_indices) {
+            cs.select(position);
+        }
         let metric = metrics::LowestFee {
             target,
             long_term_feerate: long_term_feerate_guess,
@@ -190,14 +201,16 @@ impl CoordSuperWallet {
         }
 
         let selected = cs
-            .apply_selection(&utxos)
-            .map(|(((_, account_keychain), index), utxo)| {
+            .selected_indices()
+            .iter()
+            .map(|&position| {
+                let ((_, account_keychain), index) = keychain_indices[position];
                 (
                     BitcoinBip32Path {
-                        account_keychain: *account_keychain,
-                        index: *index,
+                        account_keychain,
+                        index,
                     },
-                    utxo.outpoint,
+                    utxos[position].outpoint,
                 )
             })
             .collect();
@@ -214,6 +227,43 @@ impl CoordSuperWallet {
                 .saturating_sub(recipient_value)
                 .saturating_sub(change_value.unwrap_or(0)),
         })
+    }
+
+    /// Positions in `coins` of ones a [`RISKY_GAP`]-window restore cannot discover, found by
+    /// simulating its crawl per keychain: the window starts covering the first RISKY_GAP indices,
+    /// and each REACHABLE used index u extends it through u + RISKY_GAP. Reachability is
+    /// transitive — history sitting above the window extends nothing, so a whole cluster above
+    /// one wide gap is stranded together; measuring only the run below each coin would let
+    /// unreachable history reset the gap and leave the cluster's upper coins stranded forever.
+    /// "Used" is the indexer's is_used semantics — some txout was indexed there, spent or not —
+    /// because spent history feeds a restore's crawl just the same.
+    fn gap_stranded(&self, master_appkey: MasterAppkey, coins: &[(KeychainId, u32)]) -> Vec<usize> {
+        let mut used: BTreeMap<KeychainId, BTreeSet<u32>> = BTreeMap::new();
+        for ((keychain, index), _) in self
+            .tx_graph
+            .index
+            .keychain_outpoints_in_range(Self::key_index_range(master_appkey))
+        {
+            used.entry(keychain).or_default().insert(index);
+        }
+        let mut stranded: BTreeMap<KeychainId, BTreeSet<u32>> = BTreeMap::new();
+        for (keychain, indices) in &used {
+            let unreachable = stranded.entry(*keychain).or_default();
+            let mut reach = RISKY_GAP - 1;
+            for &index in indices {
+                if index <= reach {
+                    reach = index + RISKY_GAP;
+                } else {
+                    unreachable.insert(index);
+                }
+            }
+        }
+        coins
+            .iter()
+            .enumerate()
+            .filter(|(_, (keychain, index))| stranded[keychain].contains(index))
+            .map(|(position, _)| position)
+            .collect()
     }
 
     /// Turn a [`SendPlan`] into a signable template. This is the wallet's single change-address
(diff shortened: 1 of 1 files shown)


frostsnap_coordinator/src/bitcoin/send.rs
```diff
@@ -19,8 +20,17 @@ use frostsnap_core::{
...
+use std::collections::{BTreeMap, BTreeSet};
 use tracing::{event, Level};
 
+/// A coin a restore with this scan window cannot reach gets force-spent by every plan. From
+/// last used index p a restore derives p+1..=p+window and stalls where no history appears; 20
+/// is BIP44's gap limit, the strictest convention in common use, and defending it covers every
+/// wider window too (bdk's default 25 included) — whatever a 20-window crawl reaches, a wider
+/// one reaches a fortiori. It must stay below the wallet's own lookahead — a coin past a bigger
+/// gap never enters the UTXO set, so there is nothing to force.
+const RISKY_GAP: u32 = 20;
```
LLFourn: The doc says this "must stay below the wallet's own lookahead", and that is right — a coin past a bigger
gap never enters the UTXO set, so there would be nothing to force. But the two numbers are independent
literals in different files, `RISKY_GAP` here and `50` in `wallet_persist.rs`, with only this comment
linking them.

If someone narrows the lookahead below 20 later, nothing fails. Stranded coins simply stop being forced,
and recovery goes quietly lossy again — the exact failure this PR exists to prevent, reintroduced
silently by an edit that looks unrelated.

A `const` assertion against the lookahead, or deriving one from the other, would make the relationship
break loudly instead.

frostsnap_coordinator/src/bitcoin/send.rs
```diff
@@ -210,14 +229,59 @@ impl CoordSuperWallet {
...
 
+    /// Positions in `coins` of ones a [`RISKY_GAP`]-window restore cannot discover, found by
+    /// simulating its crawl per keychain: the window starts covering the first RISKY_GAP indices,
+    /// and each REACHABLE used index u extends it through u + RISKY_GAP. Reachability is
+    /// transitive — history sitting above the window extends nothing, so a whole cluster above
+    /// one wide gap is stranded together; measuring only the run below each coin would let
+    /// unreachable history reset the gap and leave the cluster's upper coins stranded forever.
+    /// "Used" is the indexer's is_used semantics — some txout was indexed there, spent or not —
+    /// because spent history feeds a restore's crawl just the same.
+    fn gap_stranded(&self, master_appkey: MasterAppkey, coins: &[(KeychainId, u32)]) -> Vec<usize> {
```
LLFourn: Reach is keyed by `KeychainId` here, and that is load-bearing: external history must not extend internal
reach, or a busy receive keychain would mask stranded change. The implementation gets it right.

Nothing tests it. Every new case funds the external keychain only, because `Fixture::fund` uses
`BitcoinAccountKeychain::external()`. A refactor that flattened the map, or hoisted `reach` out of the
per-keychain loop, would pass the whole suite.

A focused case — used history on one keychain, a coin past a wide gap on the other, asserting the coin is
still forced — would pin it. If it is already covered by an e2e elsewhere, that is worth knowing, but an
e2e that lives outside this repo is not a contract this file can rely on.

LLFourn (commented):
delete

nickfarrow (approved):
Think I found a bug: send max fails when a stranded coin is worth less than it costs to spend.

The max is computed from effective coins only, so uneconomical coins are excluded from it. Then it gets force-selected anyway during this stranded selection.

Ordinary sends near the advertised max go the same way, the user sees "Insufficient Balance" on a sufficient wallet that has stranded coins.

Fix is three lines -- only force a coin that actually can pay for its own rescue:

```rust
if candidates[position].effective_value(target.fee.rate) > 0.0 {
    cs.select(position);
}
```

That keeps forced coins inside the same effective set the spendable figure is built from, so the two can't disagree. It also makes forcing the no-op you describe under send max: send max already spends every effective coin, so the only thing forcing added there was the uneconomic coins that broke it.

Added tests for two things that weren't covered:
- Send max
- Use internal keychain, since the burned change is internal but every existing test funds external.

Please review my changes and reasoning!

nickfarrow:
Max send:

```
  1M sat coin at index 0, 300 sat coin at index 21:
  feerate  1  ok
  feerate 10  ERR Insufficient funds. Missing 280 sats.
  feerate 50  ERR Insufficient funds. Missing 2600 sats.
```

Regarding uneconomical coins: you'd be paying more to `rescue` the coin than the coin is worth, which is worse than losing it at any restore probability. The coin can become rescuable again as soon as fees fall.

frostsnap_coordinator/src/bitcoin/send.rs
```diff
@@ -533,6 +600,155 @@ mod test {
...
+                .iter()
+                .any(|&(_, outpoint)| outpoint == stranded_change),
+            "change past the risky gap must be force-spent too"
+        );
+    }
+
+    /// Send max needs no forcing: its target is every effective coin, so a stranded coin worth
+    /// spending is already in the selection and forcing it changes nothing.
+    #[test]
+    fn send_max_already_sweeps_a_stranded_coin() {
```
LLFourn: I think this one can go, as redundant rather than wrong.

Its assertion doesn't discriminate the feature. I ran two ablations on this branch: with your `effective_value` guard removed it passes, and with the forcing neutered entirely it still passes. Both outcomes are what the docstring predicts — "forcing it changes nothing" — so this isn't a defect in the test so much as a sign that the invariant it states is a property of send max rather than of the sweep. Send max's target is already every effective coin, so a 500,000-sat coin at index 21 is in the selection before forcing is consulted, on master as much as here.

What tips it from redundant-but-harmless to worth removing is that the send-max case which *does* discriminate is already fourteen lines below. `a_stranded_coin_below_its_spend_cost_is_left_alone` is a send-max plan on a 300-sat coin at the same index 21, and it's the single test that fails when the guard comes off. That's the interaction that actually broke — forcing an uneconomical coin, the "Insufficient Balance" on a full wallet — and it's covered. Between that test and `an_ordinary_send_consolidates_gap_stranded_coins` for the forcing path itself, I don't see what the economical send-max case adds.

Suggestion: drop it and keep the docstring, which is a genuinely useful sentence, as a comment on `gap_stranded` — it's worth saying explicitly that the sweep deliberately doesn't special-case send max. Repurposing it to dust isn't the alternative it looks like: that's exactly `a_stranded_coin_below_its_spend_cost_is_left_alone`.

frostsnap_coordinator/src/bitcoin/send.rs
```diff
@@ -533,6 +600,155 @@ mod test {
...
+        f.fund(20, 500_000, 101);
+        assert_eq!(stranded_external_indices(&f), Vec::<u32>::new());
+    }
+
+    /// Recovery is transitive: history a restore cannot reach extends nothing. With coins at 30
+    /// and 45 above one wide gap, BOTH are invisible to a restore stalled at the gap — forcing
+    /// only 30 (whose local run is wide) and not 45 (whose local run is 14) would leave 45
+    /// stranded forever, since 30's spent history keeps 45's local run short after the sweep.
+    #[test]
+    fn coins_clustered_above_one_gap_are_all_forced() {
```
LLFourn: Not a drop — a rename.

The name promises forcing and the body asserts what `gap_stranded` *returns*: `assert_eq!(stranded_external_indices(&f), vec![30, 45])`. That's a unit assertion on the analysis, and it's a good one — the clustering rule is the subtlest part of the policy and deserves pinning.

But "are_all_forced" reads as end-to-end. I neutered the forcing and this test passed, which is correct behaviour for an analysis test and would look like a hole to anyone auditing by name. Something like `gap_stranded_reports_a_whole_cluster` says what it does, and leaves the end-to-end claim to `an_ordinary_send_consolidates_gap_stranded_coins`, which does fail when forcing goes away.

LLFourn (commented):
Two notes on the tests, neither blocking, and one of them is only a name.

I ran two ablations against `2f479519` to see which tests actually discriminate the feature. Removing your `effective_value` guard fails exactly one — `a_stranded_coin_below_its_spend_cost_is_left_alone` — so that one earns its keep. Neutering the forcing entirely fails exactly two, `an_ordinary_send_consolidates_gap_stranded_coins` and `a_stranded_change_coin_is_forced_too`.

`send_max_already_sweeps_a_stranded_coin` survives both. That's exactly what its docstring predicts, so it isn't a broken test — but it means the assertion carries no signal about this feature, and the send-max case that *does* carry signal is `a_stranded_coin_below_its_spend_cost_is_left_alone` sitting right below it. Details inline.

Thanks for `a_stranded_change_coin_is_forced_too`, incidentally — that closes the per-keychain gap from my earlier review, and widening `fund` into `fund_keychain` was the right way to get there. The reason nothing tested the internal keychain was that the fixture couldn't express it.

nickfarrow (approved):
Included feedback in force push:

- Dropped `send_max_already_sweeps_a_stranded_coin`. Kept the sentence as a comment on `gap_stranded`, since it's worth recording that send max isn't special-cased on purpose.
- Renamed `coins_clustered_above_one_gap_are_all_forced` to `gap_stranded_reports_a_whole_cluster`.

Double-checked by breaking the code on purpose, the same two ways you did. Removing my three-line guard and only `a_stranded_coin_below_its_spend_cost_is_left_alone` fails. Remove the sweep entirely and `an_ordinary_send_consolidates_gap_stranded_coins` and `a_stranded_change_coin_is_forced_too` fail.

frostsnap_coordinator/src/bitcoin/send.rs
```diff
@@ -20,9 +20,17 @@ use frostsnap_core::{
...
+use std::collections::{BTreeMap, BTreeSet};
 use tracing::{event, Level};
 
+/// A coin a restore with this scan window cannot reach gets force-spent by every plan. From
+/// last used index p a restore derives p+1..=p+window and stalls where no history appears; 20
+/// is BIP44's gap limit, the strictest convention in common use, and defending it covers every
+/// wider window too (bdk's default 25 included) — whatever a 20-window crawl reaches, a wider
+/// one reaches a fortiori. It must stay below the wallet's own lookahead — a coin past a bigger
+/// gap never enters the UTXO set, so there is nothing to force.
+const RISKY_GAP: u32 = 20;
```
LLFourn: I think we should set this lower to like 10.

LLFourn:
Heads up on a force-push, since it touched your commit. The branch is now two commits, `df218fd0` and `e589cd8f`, both signed.

Your fix is `e589cd8f`. **Author, message and tree are byte-for-byte what you pushed** — same authorship line, same author timestamp, tree still `ba964ef5`. To be exact about what did change, since it is more than the signature: the commit was recreated on a new parent (`df218fd0`, the signed version of the commit underneath it), so it has a new sha, and I am now the committer, with a new committer timestamp and my signature in place of yours.

That wasn't a preference. The commit underneath yours was unsigned and needed signing, and signing a commit rewrites everything above it — there's no way to sign the one without reshaping the other. Afterwards I checked each tree hash and message against the pre-rewrite objects to make sure nothing besides that metadata moved.

Sorry to have replaced your signature on your own commit. If you'd rather it carry yours, replay `e589cd8f` and re-sign and I'll take it.

Thanks for taking both notes on the tests, and for re-running the ablations independently rather than taking my word for them.

PR #546, opened 2026-08-16
Device sign prompt discloses self-spend outputs, and both surfaces name them

LLFourn:
The device sign prompt modelled "what leaves the wallet to strangers". A transaction with no foreign recipient rendered as two pages — **Network Fee → Hold to Sign** — naming nothing at all, and the proportional high-fee guard short-circuited on the empty recipient list, dead in exactly the case where the fee is the whole cost of the transaction. Two commits on master.

## 1. The device discloses the outputs that come back to us

`PromptSignBitcoinTx` carries one **recipients** list in transaction output order, each entry knowing whether the wallet derives it — computed on the device from the `LocalSpk` keychain already in the sign task, so nothing new crosses the wire. A disclosed output of our own gets the same **Send Amount #n** / **To Address #n** pages as any recipient, with **`To Self`** under the amount.

- **Disclosure**: the single change output of an ordinary send stays undisclosed noise. Our own external addresses, change with no foreign recipient, and **two or more** change outputs are disclosed — several change outputs are not a shape this wallet produces on an ordinary send, so hiding them would hide structure the user never saw.
- **The warning is always armed, and measures value at risk rather than value displayed.** Where anything leaves the wallet, the denominator is the foreign total alone: outputs coming back to us are not at risk, and counting them lets a self-payment dilute the ratio — foreign 200k with a 60k fee warns, and adding a 1M output back to ourselves would silence it on a fee that is 30% of what actually leaves. Where nothing leaves, the fee is the only value the transaction consumes, so the self-spend total keeps the rule armed rather than dividing by zero.
- Unowned pages are pixel-identical to before, no new `AnyOf` variant, and stack-check's largest frame is unchanged at 7872 bytes (24.5% of the 25% cap).

## 2. Naming which of our addresses it pays — on both surfaces

`To Self` says the value comes back but not where, and the app said nothing at all until the device did. Both now name the script: the device prints **`Receive #0`** / **`Change #3`** under the address it derives, and Send carries a pill reading **`To self: Receive #0`**. The wording is `BitcoinBip32Path::label`, defined once because both call it. It names the keychain rather than printing its number — on a threshold device a bare `1/3` is indistinguishable from one-of-three signers, a claim about the access structure rather than about where the money went.

Naming it meant fixing what the wallet's lookup returned: `spk_index` gave `Option<u32>`, an index with the keychain thrown away, so receive #5 and change #5 came back indistinguishable. It is now `spk_path -> Option<BitcoinBip32Path>`, with `is_spk_mine` expressed in terms of it and the one caller that genuinely wants a bare index projecting at its own call site.

> **Two things found and deliberately left alone.** `psbt_to_tx_template` still classifies outputs with `Some(&((_, account_keychain), index))` — ignoring the owning appkey, so with two wallets loaded a PSBT paying wallet B's address, signed by wallet A, builds a `LocalSpk` whose derived spk differs from the real output. `spk_path` is exactly the appkey-scoped accessor that site wants. And `Transaction::is_mine` carries the same discarded keychain, which the tx-details screen renders as `Received at #3`, so a payment received at a change address reads there as a receive index.

## The screens

Device-framebuffer recordings; the last is a desktop capture showing the app and the signer together.

| | |
|---|---|
| Pure self-spend | pure self-spend |
| High-fee warning on a self-spend — previously impossible to trigger | high fee |
| Foreign recipient plus our own external address | send plus self |
| Ordinary send, one change output — the control, unchanged | ordinary send |

Pill and device naming the same output identically, ending with a cold-start relaunch where the pill still appears immediately:

pill and device agree

## Verification

`lint-ordinary`, `test-ordinary --release --all-features`, `lint-device`, legacy + frontier firmware, `stack-check --max-pct 25`, `lint-app`. Tests: 6 core policy tests over `user_prompt` (classification, ordering, disclosure), `label()` pinned to its exact wording since that string is the contract between the two surfaces, and `spk_path` pinned to return the whole path for both keychains and `None` for another appkey. The app-side behaviour — pill present for our address, absent for a foreign one, gone again after swapping back, and present on a cold start — is driven end to end against a real regtest wallet in the fork's `self_send_recognition` driver, which is what produced the capture above.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/wallet.rs
+++ b/frostsnap_coordinator/src/bitcoin/wallet.rs
@@ -123,14 +123,23 @@ impl CoordSuperWallet {
     }
 
     pub fn is_spk_mine(&self, master_appkey: MasterAppkey, spk: ScriptBuf) -> bool {
-        self.spk_index(master_appkey, spk).is_some()
+        self.spk_path(master_appkey, spk).is_some()
     }
 
-    pub fn spk_index(&self, master_appkey: MasterAppkey, spk: ScriptBuf) -> Option<u32> {
-        self.tx_graph
-            .index
-            .index_of_spk(spk)
-            .and_then(|((key, _), index)| (*key == master_appkey).then_some(*index))
+    /// The derivation this wallet reaches `spk` by, or `None` if it does not reach it.
+    /// The keychain is half the answer: receive #5 and change #5 are different scripts,
+    /// and an index alone cannot say which one was found.
+    pub fn spk_path(
+        &self,
+        master_appkey: MasterAppkey,
+        spk: ScriptBuf,
+    ) -> Option<BitcoinBip32Path> {
+        let &((key, account_keychain), index) = self.tx_graph.index.index_of_spk(spk)?;
+        (key == master_appkey).then(|| BitcoinBip32Path {
+            account_keychain,
+            index: NormalIndex::new(index)
+                .expect("bdk derived this spk, so its index is a normal child"),
+        })
     }
 
     fn descriptors_for_key(
--- a/frostsnap_core/tests/user_prompt.rs
+++ b/frostsnap_core/tests/user_prompt.rs
@@ -1,0 +1,171 @@
+use bitcoin::{Amount, ScriptBuf, TxOut};
+use frostsnap_core::bitcoin_transaction::{LocalSpk, PromptSignBitcoinTx, TransactionTemplate};
+use frostsnap_core::tweak::{BitcoinBip32Path, NormalIndex};
+use frostsnap_core::MasterAppkey;
+use schnorr_fun::fun::G;
+
+fn local_spk(bip32_path: BitcoinBip32Path) -> LocalSpk {
+    LocalSpk {
+        master_appkey: MasterAppkey::derive_from_rootkey(G.normalize()),
+        bip32_path,
+    }
+}
+
+fn template_with_input(input_sats: u64) -> TransactionTemplate {
+    let mut template = TransactionTemplate::new();
+    template.push_imaginary_owned_input(
+        local_spk(BitcoinBip32Path::external(NormalIndex::new(0).unwrap())),
+        Amount::from_sat(input_sats),
+    );
+    template
+}
+
+fn push_foreign(template: &mut TransactionTemplate, sats: u64) {
+    let spk =
+        ScriptBuf::from_hex("5120a62baa9e7c1aeda63492f2129cc8226a39db1bc05a9c11e45a61cb751a11061d")
+            .unwrap();
+    template.push_foreign_output(TxOut {
+        value: Amount::from_sat(sats),
+        script_pubkey: spk,
+    });
+}
+
+fn push_to_self(template: &mut TransactionTemplate, sats: u64, index: u32) {
+    template.push_owned_output(
+        Amount::from_sat(sats),
+        local_spk(BitcoinBip32Path::external(NormalIndex::new(index).unwrap())),
+    );
+}
+
+fn push_change(template: &mut TransactionTemplate, sats: u64, index: u32) {
+    template.push_owned_output(
+        Amount::from_sat(sats),
+        local_spk(BitcoinBip32Path::internal(NormalIndex::new(index).unwrap())),
+    );
+}
+
+fn prompt(template: &TransactionTemplate) -> PromptSignBitcoinTx {
+    template.user_prompt(bitcoin::Network::Bitcoin)
+}
+
+fn summary(prompt: &PromptSignBitcoinTx) -> Vec<(u64, Option<BitcoinBip32Path>)> {
+    prompt
+        .recipients
+        .iter()
+        .map(|r| (r.amount.to_sat(), r.owned))
+        .collect()
+}
+
+#[test]
+fn ordinary_send_hides_the_single_change_output() {
+    let mut template = template_with_input(100_000);
+    push_foreign(&mut template, 60_000);
+    push_change(&mut template, 39_000, 0);
+
+    let prompt = prompt(&template);
+    assert_eq!(summary(&prompt), vec![(60_000, None)]);
+    assert_eq!(prompt.value_moved(), Amount::from_sat(60_000));
+    assert_eq!(prompt.value_at_risk(), Amount::from_sat(60_000));
+}
+
+#[test]
+fn foreign_only_has_no_owned_recipients() {
+    let mut template = template_with_input(100_000);
+    push_foreign(&mut template, 99_000);
+
+    let prompt = prompt(&template);
+    assert_eq!(summary(&prompt), vec![(99_000, None)]);
+    assert_eq!(prompt.value_moved(), Amount::from_sat(99_000));
+    assert_eq!(prompt.value_at_risk(), Amount::from_sat(99_000));
+}
+
+#[test]
+fn send_that_also_pays_our_own_external_address_discloses_it_in_order() {
+    let mut template = template_with_input(100_000);
+    push_foreign(&mut template, 50_000);
+    push_to_self(&mut template, 30_000, 1);
+    push_change(&mut template, 19_000, 0);
+
+    let prompt = prompt(&template);
+    assert_eq!(
+        summary(&prompt),
+        vec![
+            (50_000, None),
+            (
+                30_000,
+                Some(BitcoinBip32Path::external(NormalIndex::new(1).unwrap()))
+            ),
+        ]
+    );
+    assert_eq!(prompt.value_moved(), Amount::from_sat(80_000));
+    // Disclosed but not at risk: counting the self payment would dilute the
+    // proportional ratio, letting a large enough one silence the fee warning.
+    assert_eq!(prompt.value_at_risk(), Amount::from_sat(50_000));
+}
+
+#[test]
+fn pure_self_spend_discloses_everything_including_change() {
+    let mut template = template_with_input(100_000);
+    push_to_self(&mut template, 70_000, 1);
+    push_change(&mut template, 29_000, 0);
+
+    let prompt = prompt(&template);
+    assert_eq!(
+        summary(&prompt),
+        vec![
+            (
+                70_000,
+                Some(BitcoinBip32Path::external(NormalIndex::new(1).unwrap()))
+            ),
+            (
+                29_000,
+                Some(BitcoinBip32Path::internal(NormalIndex::new(0).unwrap()))
+            ),
+        ]
+    );
+    assert_eq!(prompt.value_moved(), Amount::from_sat(99_000));
+    // Nothing leaves, so the fee is the only value consumed and the self total
+    // is what keeps the proportional rule armed rather than dividing by zero.
+    assert_eq!(prompt.value_at_risk(), Amount::from_sat(99_000));
+}
+
+#[test]
+fn consolidation_to_change_only_is_still_disclosed() {
+    let mut template = template_with_input(100_000);
+    push_change(&mut template, 99_000, 0);
+
+    let prompt = prompt(&template);
+    assert_eq!(
+        summary(&prompt),
+        vec![(
+            99_000,
+            Some(BitcoinBip32Path::internal(NormalIndex::new(0).unwrap()))
+        )]
+    );
+}
+
+#[test]
+fn two_change_outputs_with_a_foreign_recipient_are_disclosed() {
+    let mut template = template_with_input(100_000);
+    push_foreign(&mut template, 50_000);
+    push_change(&mut template, 25_000, 0);
+    push_change(&mut template, 24_000, 1);
+
+    let prompt = prompt(&template);
+    assert_eq!(
+        summary(&prompt),
+        vec![
+            (50_000, None),
+            (
+                25_000,
+                Some(BitcoinBip32Path::internal(NormalIndex::new(0).unwrap()))
+            ),
+            (
+                24_000,
+                Some(BitcoinBip32Path::internal(NormalIndex::new(1).unwrap()))
+            ),
+        ]
+    );
+    assert_eq!(prompt.value_moved(), Amount::from_sat(99_000));
+    assert_eq!(prompt.value_at_risk(), Amount::from_sat(50_000));
+}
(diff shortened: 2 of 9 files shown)


LLFourn:
## Design revision: self-spend outputs are ordinary recipients, marked **OURS**

The aggregated "To This Wallet" page (shown in the description above) is gone. A disclosed self-spend output now gets the same **Send Amount #n** and **To Address #n** pages as any other recipient — same numbering, transaction output order — with a bold confirm-green **`OURS: <keychain>/<index>`** marker under both titles. The derivation shown is the one the device itself attests (a fourth change address reads `OURS: 1/3`), and the green deliberately echoes the hold-to-confirm ring: reassurance, not a warning. Unowned pages are pixel-identical to before; the disclosure policy and the always-armed value-moved fee rule are unchanged, and the page arithmetic loses its self-page special case entirely (`recipients × 2 + warning + fee + confirm`).

Full-resolution mp4s: the `*-v2.mp4` assets on the [release](https://github.com/LLFourn/frostsnap/releases/tag/self-spend-videos).

**Pure self-spend** — every page names the value and the address, both marked OURS:

pure self-spend, OURS pages

**High-fee warning on a self-spend** — still fires via the value-moved rule, now after the OURS recipient pages:

high fee self-spend warning

**Foreign recipient plus a payment to our own external address** — the self output rides in the normal page sequence as recipient #2:

send plus self output

**The app's own Send flow, send-max to our own receive address**:

app send to self

**Ordinary send, one change output** — the control, unchanged:

ordinary send unchanged

LLFourn:
## Design v3: the ownership marker is now a small footnote

Iterating on the comment above — the bold green **OURS** banner competed with the page's main content. It's now the literal **`(to self: <keychain>/<index>)`**, set in the prompt's smallest face in secondary gray, always below the main content: under the BTC line on the amount page, under the address grid on the address page. An annotation, not an announcement; the derivation shown is still the one the device attests. Everything else (model, disclosure policy, fee rule, page order) is unchanged, and unowned pages remain pixel-identical.

Full-resolution mp4s: the `*-v3.mp4` assets on the [release](https://github.com/LLFourn/frostsnap/releases/tag/self-spend-videos).

**Pure self-spend**:

pure self-spend, footnote

**High-fee warning on a self-spend**:

high fee self-spend warning

**Foreign recipient plus a payment to our own external address**:

send plus self output

**The app's own Send flow, send-max to our own receive address**:

app send to self

**Ordinary send, one change output** — the control, unchanged:

ordinary send unchanged

LLFourn:
## The app side, on desktop: pasting your own address into Send

Captioned walkthrough of the whole flow — the wallet's own receive address pasted into Send, and the signer (live in the simulator tray) reviewing it with the `(to self: 0/0)` footnote before hold-to-sign. Reproducible with `./fsim test self_send_recognition`.

desktop capture: own-address recognition in Send

LLFourn:
## A self-payment can silence the high-fee warning

Reviewing this branch turned up one defect, and it is in the part the PR adds rather than the part it
changes. Ours to fix, and the fix is written — this is the intent, with the clip that shows it.

Disclosing owned outputs also put them into the denominator the proportional fee warning measures
against. `value_moved()` sums every **disclosed** recipient, so an output that comes back to us inflates
it and the fee looks proportionally smaller. Master measured against `foreign_recipients` — value
leaving the wallet — and those two agree on an ordinary send, then diverge exactly where a coordinator
has freedom.

`HIGH_FEE_ABSOLUTE_THRESHOLD_SATS` bounds the damage: the absolute rule is `fee_sats > 100_000`, so
anything **over** 0.001 BTC still warns regardless. The window is a fee **at or below** 0.001 BTC that
nonetheless exceeds 5% of what actually leaves — so, small sends.

**200 000 sats to a foreign address, 1 000 000 back to ourselves, 60 000 fee.** That fee is 30% of what
leaves the wallet. Before, the denominator is 1 200 000, `60 000 > 5%` is false, and the prompt goes
straight from the recipients to the fee to hold-to-sign:

no warning on a 30% fee

This is deliberate and pinned, not an oversight —
`fee_warning_denominator_includes_owned_recipients` asserts it. So the fix changes that expectation
rather than adding a missing one.

**The fix makes the rule piecewise.** Where foreign value leaves, that alone is the denominator; outputs
returning to us are not at risk, and counting them lets a self-payment dilute the ratio. Where nothing
leaves, the fee is the only value the transaction consumes, so the self-spend total keeps the warning
armed rather than dividing by zero — `pure_self_spend_can_trigger_the_proportional_fee_warning` stays
passing untouched, which is what ruled out simply using the foreign value everywhere.

`value_moved()` survives for what the prompt itemises. The two were one accessor doing two jobs and the
warning had quietly taken the display one.

**Changing the denominator meant changing the warning's sentence, and rendering caught our first
attempt at that.** "Fee exceeds 5% of the value moved" is accurate on this branch as it stands — it
names the denominator this branch actually uses. Our fix is what makes it stale, and our first
replacement was still ambiguous: every output here is titled *Send Amount #n*, owned ones included, so
"the amount sent" reads as the 1 200 000 total. The sentence now names what the rule divided by, and
`warning_lines()` is a pure function with a test pinning all three branches — the old tests only
asserted that a warning page existed, so wording that named the wrong quantity passed everything.

Fix is 3 files, +110/-21. I will push it to this branch and add the after clip here once its checks
pass.

---

Pushed as `49830884` — 3 files, +110/-21, and the branch is green 10/10.

warning now fires

Same transaction as above. The warning now appears between the recipients and the fee, and reads
"Fee exceeds 5% of the amount leaving" — naming what the rule divided by rather than the total, since
every output here is titled *Send Amount #n*, owned ones included.

`warning_lines()` is a pure function now with a test pinning all three branches (mixed, pure self-spend,
absolute), so wording that names the wrong quantity fails rather than rendering.

LLFourn:
## The app knows too now: a "to self" pill in Send

New commit on the branch (`c354e277`): the Send flow resolves the recipient against the wallet's own addresses (existing `search_for_address` bridge, latest-input-wins with generation-guarded results) and shows a quiet tonal **to self** chip — on the entry field while typing, on the completed Recipient card after advance. Captioned desktop capture below: foreign address → no pill, our own address → pill, swapped back → pill leaves, then the signer confirms with the `(to self: 0/0)` footnote. Driven by `./fsim test self_send_recognition` on the fork.

to self pill, captioned desktop capture

LLFourn:
Pill implementation simplified per review (`c354e277` → `a0b18320`): ownership is now one `#[frb(sync)] is_address_mine` call delegating to the existing `is_spk_mine` — the async `search_for_address` route with its generation guards and cached scan bound is gone, and a synchronous `txState` call at Send-page entry makes descriptor initialization structural (the button enables before the tx-stream task acquires the wallet). The GIF above is unchanged — the pill just lands a frame sooner.

nickfarrow:
PR to introduce some language changes to the device screens, cleaning things up a bit for the user:
https://github.com/LLFourn/frostsnap/pull/22

LLFourn:
## Footnote split across the two pages it belongs on

Latest push (`03eb2d75`…`7b41ba4b`) refines the annotation, and supersedes the GIFs in the earlier comments.

The two pages answer different questions, so the footnote splits: the amount page asks where the value is going and the whole answer is **`To Self`**; *which* of our scripts it lands on moves under the address it derives, and it **names the keychain** — **`Receive #0`**, **`Change #3`** — rather than printing its number. That last part matters on a threshold device: a bare `1/3` is indistinguishable from one-of-three signers, a claim about the access structure rather than about where the money went. The demo's owned recipient is now taproot too, since an output the wallet derives always is.

Also dropped in this push: the string-literal assertions on the warning copy and the page-index shape tests — they restated the branches they called rather than pinning behaviour.

| | |
|---|---|
| Pure self-spend | pure self-spend |
| High-fee warning on a self-spend | high fee |
| Foreign recipient + our own external address | send plus self |
| The app's Send flow, send-max to ourselves | app send to self |

And the desktop capture end to end — pill on paste, pill gone on swap, device agreeing, and a cold-start relaunch where the pill still appears immediately:

to self pill, desktop

nickfarrow:
The rebase dropped `frostsnap_widgets/tests/sign_prompt_pages.rs` (7 tests), which was added by this PR. Maybe it got pulled into your `fsim` again?

LLFourn:
@nickfarrow I dropped them because they were very brittle and useless. They test that the device display shows certain pages one after the other on sign prompt. If we change that then they break. These sort of things always need manual eyeball verification anyway. Just because a page is there doens't mean it's displaying the correct thing.

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -773,4 +773,47 @@ mod test {
             )),
         );
     }
+
+    #[test]
+    fn is_spk_mine_scopes_by_appkey_and_covers_both_keychains() {
```
LLFourn: This is the wrong place to test this. A trivial spk "is_mine:" shouldn't need to consutrct a fake electrum sever.

LLFourn:
## The pill now names what the device names

Fourth commit (`6148ab42`). The two surfaces described the same output with different amounts of information — the screen said `Receive #0`, the pill said only `to self` — so the pill now reads **`To self: Receive #0`**, and the wording lives in `BitcoinBip32Path::label` that both call.

Getting the app that far meant fixing the lookup underneath it: `spk_index` returned `Option<u32>`, an index with the keychain thrown away, so receive #5 and change #5 came back indistinguishable. It is now `spk_path -> Option<BitcoinBip32Path>`, with `is_spk_mine` expressed in terms of it and the one caller that genuinely wants a bare index saying so at the call site.

Same ambiguity still sits in the tx-details screen's `Received at #3`, which reads a bare index out of `Transaction::is_mine` — untouched here, worth a separate look.

pill and device name the output identically

frostsnap_core/tests/user_prompt.rs
```diff
@@ -0,0 +1,162 @@
...
+        summary(&prompt),
+        vec![
+            (50_000, None),
+            (
+                30_000,
+                Some(BitcoinBip32Path::external(NormalIndex::new(1).unwrap()))
+            ),
+        ]
+    );
+    assert_eq!(prompt.value_moved(), Amount::from_sat(80_000));
```
nickfarrow, line 100: We could also test the `value_at_risk` in these tests:
```
// The self-payment is disclosed but is not at risk: counting it would dilute the
// proportional fee ratio and let a coordinator silence the high-fee warning.
assert_eq!(prompt.value_at_risk(), Amount::from_sat(50_000));
```

frostsnapp/lib/wallet_send.dart
```diff
@@ -82,9 +82,49 @@ class _WalletSendPageState extends State<WalletSendPage> {
     sub.start().listen((_) => mounted ? setState(() {}) : null);
 
     addrController = AddressInputController(state);
+    addrController.controller.addListener(_onRecipientTextChanged);
```
nickfarrow, line 85: Clanker is telling me:

The txState call that used to sit just above this line didn't survive the rewrite:
```
widget.superWallet.txState(masterAppkey: widget.masterAppkey);
```
spk_path takes &self, so unlike its neighbours it can't run lazily_initialize_key itself, and nothing else on this page does. An immediate paste asks an uninitialised index and gets no pill.

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -123,14 +123,23 @@ impl CoordSuperWallet {
     }
 
     pub fn is_spk_mine(&self, master_appkey: MasterAppkey, spk: ScriptBuf) -> bool {
```
nickfarrow, line 125: No callers now that `owned_address_label` goes straight to `spk_path`.

nickfarrow (approved):
Looks great, three minor non-blocking nits

frostsnapp/lib/wallet_send.dart
LLFourn: it's always initialzied at this point. But this whole "initialized" thing needs to die.

frostsnap_core/tests/user_prompt.rs
LLFourn: Added, with your exact case as the anchor — the mixed send now asserts `value_at_risk() == 50_000` against `value_moved() == 80_000`, so the two accessors are pinned apart rather than only being described in a commit message. Also covered the pure self-spend (nothing leaves, so the self total keeps the rule armed rather than dividing by zero), the two-change case, and the foreign-only ones.

This is a much better seam than what it replaces: pure, no fixture, and it fails exactly when the denominator regresses.

frostsnap_coordinator/src/bitcoin/wallet.rs
LLFourn: Correct that it has no callers now. LLFourn's call is to leave it, so it stays as-is — I did briefly delete it and have put it back.

frostsnapp/lib/wallet_send.dart
LLFourn: Checked this rather than assuming, because it's the kind of thing that would only show up as an intermittently missing pill.

The initialization has already happened by the time this page exists: `sub_tx_state` calls `tx_state` **synchronously in its own body** (`stream.add(self.tx_state(master_appkey))`), and `tx_state` -> `list_transactions` -> `lazily_initialize_key`. Nothing reaches Send without the wallet being open, which is what mounts that subscription — so there is no window where the index is uninitialised and the page is tappable.

The driver covers the case your comment describes, deliberately: `self_send_recognition` ends with `restartApp()`, re-enters the wallet, goes straight into Send and pastes **without** waiting for balance or any settle signal, then asserts the pill. It passes with the `txState` call removed. If the race were real that scenario is what would catch it.

(The `txState` line was mine and it was doing nothing — as LLFourn says below, the whole "initialized" notion wants to die rather than get another caller propping it up.)

LLFourn:
On the missing `frostsnap_widgets/tests/sign_prompt_pages.rs` — not a rebase casualty, deleted on purpose at LLFourn's request, along with the `warning_lines` string test and (later) a `spk_path` test I had added. The page-shape tests asserted which widget type lands at which index, which the page arithmetic already states; the string tests asserted literals against the same literals.

What replaced them is coverage at seams that can actually fail: the `user_prompt` policy tests in core (classification, ordering, disclosure, and now `value_at_risk` vs `value_moved` per @nickfarrow's suggestion), and the `self_send_recognition` driver for the app path. The one seam still missing a unit test is `spk_path`, and that is deliberate too — reaching it means standing up a database and a chain client to assert a map lookup. The fix there is structural, noted in the commit: the sans-io wallet wants to be its own struct, and `SuperWallet` is meant to disappear once wallet locking gives each wallet its own BDK database.

PR #547, opened 2026-08-16
[coord] Test that a synced coin stays spendable after a restart

LLFourn:
This is @nickfarrow's test from LLFourn/frostsnap#18, carried to this repo now that the fix it pins (#543) has merged here. The regression belongs next to the fix rather than in a fork branch. The commit keeps him as author.

A coin arrives at an address the wallet only knows about because a sync reported it, the app restarts, and the coin must still be spendable. That is the invariant `apply_update` broke by computing the reveal changeset from `reveal_to_target_multi`, using it only for the `changed` flag, and dropping it: the frontier a sync taught us was RAM-only, so after a restart the stored txs re-indexed against a stale window. Balance re-derives from the spk map and healed itself every session; coin selection reads the outpoint set, which does not. Full balance shown, nothing spendable.

The test asserts both layers — that the frontier is on disk, and that send max can still see the coin — so a failure says which one broke.

**One change from #18.** The coin's index is read off `wallet.lookahead()` rather than written as the literal `60`. `60` was correct against today's lookahead of `50`, but the two are independent numbers in different files: raise the lookahead past `60` and the load-time re-index finds the coin unaided, so the test passes whether or not the fix is present. Verified rather than argued — with `changeset.indexer.merge` ablated and `KeychainTxOutIndex::new(50, false)` raised to `100`, the literal version passes and the derived version still fails.

Evidence run locally on `f0e6bdb0`:

- `cargo test -p frostsnap_coordinator bitcoin::send` — 8/8 pass.
- Ablate `changeset.indexer.merge(indexer_changeset)` (`wallet.rs:404`) → fails with *"the frontier the sync established must be on disk"*.
- Ablate it **and** widen the lookahead 50 → 100 → still fails.
- `cargo fmt --check` and `cargo clippy --all-features --tests` clean.

Test-only, so no media.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/send.rs
+++ b/frostsnap_coordinator/src/bitcoin/send.rs
@@ -542,6 +542,103 @@ mod test {
         assert_eq!(paid, planned, "the amount shown is the amount paid");
     }
 
+    /// A reveal the wallet only learned about from a sync must survive a restart, and the coin it
+    /// makes visible must stay spendable. The reveal was the exact changeset `apply_update`
+    /// discarded, so every reload re-indexed stored txs against a too-narrow window and the coin
+    /// showed in the balance but never in coin selection.
+    #[test]
+    fn a_synced_reveal_survives_a_restart_and_its_coin_stays_spendable() {
+        const VALUE: u64 = 1_000_000;
+
+        let db = Arc::new(Mutex::new(rusqlite::Connection::open_in_memory().unwrap()));
+        let master_appkey =
+            MasterAppkey::derive_from_rootkey(Point::random(&mut rand::thread_rng()));
+        let recipient = bitcoin::Address::from_str("bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4")
+            .unwrap()
+            .require_network(NETWORK)
+            .unwrap();
+        let external = (master_appkey, BitcoinAccountKeychain::external());
+        let blocks = [
+            BlockId {
+                height: 0,
+                hash: bitcoin::constants::genesis_block(NETWORK).block_hash(),
+            },
+            BlockId {
+                height: 100,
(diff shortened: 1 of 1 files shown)


LLFourn:
Two things about what this test can establish, raised now because they are cheaper to settle before it merges than to revisit as a follow-up. Neither is a blocker and neither is a merge verdict — the call on landing it is not mine.

**It writes the input whose persistence it tests.** `last_active_indices: [(external, far)]` is hand-built here, and that is exactly the field `apply_update` was dropping. So the test pins "if the chain source reports a frontier ahead of what we persisted, that survives a restart" — it cannot say the chain source *would* report it. The `DerivedSpkTracker` half of that contract is mocked, and a regression there has the same user-visible symptom with this test green.

**It asserts one side of the invariant it is named for.** #543 is "balance and send max agree", and the test only ever asks for send max. So a change that broke the *balance* side would leave this test green while the agreement it is named for was actually broken — not a vacuous pass, an unobserved failure.

Two changes that need nothing new, if you think they are worth it:

1. **Assert the property rather than the scenario** — that every canonical **unspent** owned output which is otherwise spendable is visible to coin selection, not merely to the spk map. (Not "every owned txout": spent, immature and policy-ineligible outputs are legitimately absent, and the assertion has to say so or it fails for the wrong reasons.) Then the check is the invariant itself and can run after any update, not just this hand-built one.
2. **Assert the balance side too**, so "agree" is checked rather than assumed.

The honest fix for the first point is an e2e that drives a real sync, which is not something this repo can express today — so these two are the available improvements, not the complete answer.

One thing worth recording either way: the white-box `last_revealed_index` assertion is **not** what gives the test its teeth. With the fix ablated and that assertion deleted, the behavioural one still fails — *"balance and send max must agree across a restart; send max saw -41"*. So it is a useful diagnostic pointing at which layer broke, not the thing catching the regression.

nickfarrow (approved):

PR #548, opened 2026-08-17
Recover change a gap-limit restore missed and make it spendable again

LLFourn:
A restored wallet walks the gap limit and stops at the first gap wider than its scan window. Change revealed far past the last genuinely used index — as manufactured by the `fee()` defect fixed in #542 — is invisible after restore: not in the balance, not spendable. #545 attacks this from the sending side so the gap never forms; this is the other half, recovery for wallets where it already has.

## How it works

An outgoing transaction is discoverable even when its change is not: `is_mine` chains outputs with prevouts, so owning a spent prevout is enough to hold the tx — its unattributed change script is already sitting in the local graph. Recovery is therefore a **purely local comparison**; the network holds no information we lack, and no candidate script is ever sent to a server.

`CoordSuperWallet::recovery_scan`:

1. Flags canonical txs that spend ours, have more than one output, and pay us nothing (send-max legitimately has no change, so single-output spends are exempt).
2. Derives candidate spks locally out to **1000 indices past each keychain's frontier** (~80ms for the full 2000 on a laptop — which is why there is no blocking scan UI).
3. On a match, reveals to **exactly the highest matched index** — never the budget — through the ordinary `reveal_to_target` path, then re-indexes every stored transaction (bdk indexes a tx once, against the spks derived at that moment, so a reveal strands already-stored outputs without a full re-index).
4. Iterates: a reveal can make an already-stored descendant spend outgoing-and-unreconciled.

Scan progress (compared-to high-water marks, outstanding txids) persists in its **own tables, never in `last_revealed`** — parking progress in the frontier would advance address allocation and manufacture the next restore's gap. Only a real match mutates wallet derivation state, and only up to the matched index.

The scan runs automatically: on every sync update naming a key (deliberately not gated on the changeset — an upgraded wallet whose stranded tx is already persisted applies every sync as a no-op) and fire-and-forget at wallet open, offline included. Repeat scans over an unchanged wallet are one canonicalization pass, zero derivations.

After a far match, the revealed range reaches the electrum server in the same session: `bdk_electrum_streaming` 0.5.3 widens a tracked descriptor's subscription window in place when it is re-tracked with a larger `next_index` (evanlinjin/experiments#8, released for this), so `monitor_keychain` simply re-announces the keychain on the live connection and the new range is subscribed with no reconnect. A unit test pins that upstream contract against future dep bumps.

## The consolidation nudge

A recovered coin still sits at an index a standard restore would miss, so the wallet home gains a banner counting the spendable coins past the risky gap — computed by the same crawl simulation #545's `plan_send` (now merged) force-consolidates by, and counting **only coins a plan would actually rescue** (the relay-floor rule), so one send always clears it. Tapping it opens Send; the copy tells the user their next payment consolidates automatically. It applies equally to organically stranded coins — no recovery scan required.

Transactions the scan budget cannot account for get **no dedicated surface**: the detection predicate cannot distinguish missing change from a legitimately changeless exact-value send (two recipients, no change), so a warning there would mostly cry wolf on wallets missing nothing. In practice the budget also covers the realistic #542 shapes — matches act as stepping stones, extending reach another 1000 past each recovered index.

## Testing

- 9 coordinator tests pin the recovery contract: no-match leaves wallet derivation state byte-identical while progress persists; match reveals exactly to the matched index; every stored tx is re-indexed (a stray output below the new frontier becomes spendable off an unrelated match); restart resumes from the scan table with no work; descendants reconcile both within one scan and across a sync boundary; a no-op update still triggers recovery; the electrum tracker widen contract; the banner summary counts exactly the coins a plan would rescue.
- Demoed end-to-end on regtest against a real bitcoind+electrs: a device-signed tx paying change to the wallet's own change descriptor at index 800 is recovered silently, flagged by the consolidation banner, and swept back on-chain via a send. GIFs in the comments.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -350,9 +350,9 @@ dependencies = [
 
 [[package]]
 name = "bdk_electrum_streaming"
-version = "0.5.2"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d1e86daa7b72258672522b521569386692d51ef70b5a4e879cdd8e6a6caa427f"
+checksum = "a16576b1efa0960f4c6d65d588005889ee9bee600d47540adca62d21003154bc"
 dependencies = [
  "anyhow",
  "bdk_core",
--- a/frostsnap_coordinator/Cargo.toml
+++ b/frostsnap_coordinator/Cargo.toml
@@ -23,7 +23,7 @@ rsa.workspace = true
 serde = { version = "1.0", features = ["derive"] }
 
 bdk_chain = { version = "0.23.3", features = ["rusqlite"] }
-bdk_electrum_streaming = { version = "0.5.2" }
+bdk_electrum_streaming = { version = "0.5.3" }
 bdk_coin_select = { version = "0.4.1" }
 
 
--- a/frostsnap_coordinator/src/bitcoin.rs
+++ b/frostsnap_coordinator/src/bitcoin.rs
@@ -2,8 +2,11 @@ pub mod chain_sync;
 mod handler_state;
 pub mod outgoing;
 pub mod psbt;
+pub mod recovery_scan;
 pub mod send;
 pub mod status_tracker;
+#[cfg(test)]
+pub(crate) mod test_util;
 pub mod tofu;
 pub mod wallet;
 mod wallet_persist;
--- a/frostsnap_coordinator/src/bitcoin/chain_sync.rs
+++ b/frostsnap_coordinator/src/bitcoin/chain_sync.rs
@@ -182,6 +182,9 @@ impl ChainClient {
         block_on(response)?
     }
 
+    /// Track `keychain` through `next_index` plus lookahead. Re-calling with a larger
+    /// `next_index` widens the live subscription window in place (bdk_electrum_streaming
+    /// >= 0.5.3); equal or smaller is a no-op, so the window never narrows.
     pub fn monitor_keychain(&self, keychain: KeychainId, next_index: u32) {
         self.start_client();
         let descriptor = descriptor_for_account_keychain(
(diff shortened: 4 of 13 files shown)


LLFourn:
Demo from the regtest app-sim — a device-signed transaction (built by bitcoind, not the app) pays its change to the wallet's own change descriptor at index 800, far out of range. The scan recovers it silently, the consolidation banner flags the recovered coin as restore-fragile (it survives an app restart — the frame below is from the restarted session), and one send sweeps it back within reach, clearing the nudge. The e2e run also asserts the beyond-budget case (index 2500) surfaces nothing.

recover and consolidate

LLFourn:
Updated: the tracker-rebuild-on-reconnect workaround is gone. `bdk_electrum_streaming` 0.5.3 (evanlinjin/experiments#8, released today) widens a tracked descriptor's subscription window in place when re-tracked with a larger `next_index`, so `monitor_keychain` is a plain passthrough again and a recovery-scan match costs no reconnect — the delta range is subscribed on the live connection. A unit test pins that upstream contract against future dep bumps, and both regtest e2e runs pass on the new mechanism (including the sweep-confirms assertion, which only holds if the widened window actually reached the server).

LLFourn:
Stills of the consolidation nudge, from the e2e run (the drive also asserts the tap behavior):

**The banner** — appears once the recovered coin sits past the risky gap, with the at-risk amount:

consolidation banner

**Tapping it** opens the Send flow directly — any payment consolidates the stranded coin, since `plan_send` force-selects it:

banner tap opens send

LLFourn:
Folded into #550, which already contains this commit as the first of its two — closing here so there's one review surface for the recovery scan and the consolidation remedy it needs.

PR #549, opened 2026-08-17
app: disable look for address button on invalid or wrong-network input

### LINKED ISSUE #252: check address function snafu
You shouldn't be able to "look for address" unless a valid address for the correct network is in the input.

Image

7-falseparfait:
fixes #252

What changed:

1. Added a listener on CheckAddressPage's textInputController that re-validates the input text against the active wallet network using Address.fromString.

2. Disabled the "Look for address" button (onPressed: null) whenever the input is empty, malformed, or for the wrong network.

Added explicit error feedback (Invalid address for this network) directly on the input field when non-empty text fails validation.

### DIFF
--- a/frostsnapp/lib/address.dart
+++ b/frostsnapp/lib/address.dart
@@ -32,18 +32,18 @@ class _CheckAddressPageState extends State<CheckAddressPage> {
     super.dispose();
   }
 
+  // Rebuilds on every change rather than on validity transitions: the error text
+  // reads both the field's emptiness and its validity, and `_isValidAddress` is
+  // false for both, so a validity-only trigger misses empty<->invalid entirely.
   void _onTextChanged() {
     final walletCtx = WalletContext.of(context);
     final text = textInputController.text.trim();
-    final isValid =
-        text.isNotEmpty &&
-        walletCtx != null &&
-        Address.fromString(s: text, network: walletCtx.network) != null;
-    if (isValid != _isValidAddress) {
-      setState(() {
-        _isValidAddress = isValid;
-      });
-    }
+    setState(() {
+      _isValidAddress =
+          text.isNotEmpty &&
+          walletCtx != null &&
+          Address.fromString(s: text, network: walletCtx.network) != null;
+    });
   }
 
   Future<SearchResult> searchAddress() async {
@@ -65,8 +65,8 @@ class _CheckAddressPageState extends State<CheckAddressPage> {
   }
 
   Widget _buildSearchResults(SearchResult? result) {
-    final walletCtx = WalletContext.of(context)!;
     if (result == null) return const SizedBox.shrink();
+    final walletCtx = WalletContext.of(context)!;
 
     final children = <Widget>[
       Text(
--- a/frostsnapp/lib/address.dart
+++ b/frostsnapp/lib/address.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:frostsnap/contexts.dart';
+import 'package:frostsnap/src/rust/api/bitcoin.dart';
 import 'package:frostsnap/src/rust/api/super_wallet.dart';
 import 'package:frostsnap/theme.dart';
 import 'package:frostsnap/wallet_receive.dart';
@@ -16,17 +17,33 @@ class _CheckAddressPageState extends State<CheckAddressPage> {
   Future<SearchResult>? searchFuture;
   int currentDepth = 0;
   int searchSize = 100;
+  bool _isValidAddress = false;
 
   @override
   void initState() {
     super.initState();
     textInputController = TextEditingController();
+    textInputController.addListener(_onTextChanged);
   }
 
   @override
   void dispose() {
     textInputController.dispose();
     super.dispose();
+  }
+
+  void _onTextChanged() {
+    final walletCtx = WalletContext.of(context);
+    final text = textInputController.text.trim();
+    final isValid =
+        text.isNotEmpty &&
+        walletCtx != null &&
+        Address.fromString(s: text, network: walletCtx.network) != null;
+    if (isValid != _isValidAddress) {
+      setState(() {
+        _isValidAddress = isValid;
+      });
+    }
   }
 
   Future<SearchResult> searchAddress() async {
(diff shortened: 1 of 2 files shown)


LLFourn:
Thanks for this — and for #252 being a real papercut worth closing. I integrated it onto our sim stack and drove it on an Android emulator with a regtest wallet. The button half works exactly as advertised; the error text had a problem, and the fix for it went up as 7-falseparfait/frostsnap#1, which you've since merged.

**The button logic is correct.** Empty → valid address enables it; editing one character of a valid address disables it again and shows the error. That control matters, because it proves `errorText` renders fine when a rebuild happens:

controls work

**Typing an invalid address into a fresh field showed no error at all.** Open Check address, tap the field, type `notanaddress` — the button stays correctly disabled, but no message appears and the underline stays blue:

no error on first invalid input

**And once the error did show, clearing the field left it behind.** Delete the entry completely and the field renders empty with `Invalid address for this network` and the red underline still on it:

error persists on an empty field

Both survived dismissing the keyboard, scrolling, re-focusing the field, and typing then deleting another character — they weren't artifacts of a particular interaction order.

The cause is that the rebuild trigger was narrower than the state being rendered. `errorText` derives from two things:

```dart
errorText: textInputController.text.isNotEmpty && !_isValidAddress ? 'Invalid address for this network' : null,
```

but the only thing that re-runs `build` is:

```dart
if (isValid != _isValidAddress) { setState(() { _isValidAddress = isValid; }); }
```

`_isValidAddress` is `false` for *both* "empty" and "invalid", so empty→invalid and invalid→empty each move what should be rendered without moving the trigger. Nothing else rebuilds this page either: `WalletContext.of` uses `getInheritedWidgetOfExactType` and its `updateShouldNotify` is `=> false`, so there's no ambient rebuild to paper over it.

One thing worth calling out, since it's the case in your title: `Address::from_str(s).ok()?.require_network(network).ok()` returns `None` for malformed **and** wrong-network input alike, so a wrong-network address takes the same path as the garbage above. I read that from the code rather than filming it separately — but it means the wrong-network case was affected by the same gap.

---

**The fix — 7-falseparfait/frostsnap#1**, based on your branch so you could review it rather than have your branch written to, and now merged. `_onTextChanged` rebuilds on every text change, so the trigger covers everything the view reads. Widening the guard to track emptiness as well would also work, but it re-encodes the render condition in a second place, which is what broke here.

Both transitions now behave:

error appears and clears

valid enables, extra char disables

It carries a widget test covering both transitions, which fails on `9404ae5d` and passes with the fix — `0 passed, 2 failed` → `2 passed`, on the pinned Flutter 3.38.5 with `--enforce-lockfile`, `flutter analyze` clean.

**Verdict: good from my side.** You merged 7-falseparfait/frostsnap#1, so the head is now `0f07db5f`
and both transitions behave. I checked the branch rather than the merge notice: `9404ae5d..0f07db5f` is
exactly `address.dart` and the regression test (+46/-10), and the tree is identical to the one I tested.
Nothing further from me in this scope.

*(This comment previously read "requesting changes" — accurate when the head was `9404ae5d`, and left
here as an edit rather than a second comment.)*

**Two caveats on that evidence, so it isn't read as more than it is.** That PR had **zero CI check runs** — it targeted a branch on your fork, where the workflows don't run — so the test result above is a local run plus the recordings, not a green upstream build. And `.github/workflows/test.yml` has no `flutter test` job at all, so even on `master` that test wouldn't execute. It's real and it runs locally; it just isn't guarding anything in CI yet. Raising that gap separately — it isn't yours to fix here.

7-falseparfait:
Thanks for testing this on regtest and catching that state transition bug! That makes total sense ....`_isValidAddress` staying `false` across the empty invalid transition prevents `setState()` from firing to update `errorText`. 

Driving the rebuild off text changes directly makes complete sense to keep the render condition and trigger aligned. Looking forward to reviewing your PR!

frostsnapp/lib/address.dart
```diff
@@ -29,6 +32,20 @@ class _CheckAddressPageState extends State<CheckAddressPage> {
...
   }
 
+  void _onTextChanged() {
+    final walletCtx = WalletContext.of(context);
+    final text = textInputController.text.trim();
+    final isValid =
+        text.isNotEmpty &&
+        walletCtx != null &&
+        Address.fromString(s: text, network: walletCtx.network) != null;
+    if (isValid != _isValidAddress) {
```
LLFourn: This guard is the whole of it: the trigger tracks *validity*, but `errorText` below reads **both** the field's emptiness and its validity — and `_isValidAddress` is `false` for "empty" as well as "invalid". So the two transitions between those states change what should be rendered without changing this condition, and no rebuild happens.

Concretely: empty → `notanaddress` is `false → false` (no error ever appears), and invalid → cleared is also `false → false` (the error stays on an empty field). Both reproduce on an emulator; clips are in the summary comment.

`7-falseparfait/frostsnap#1` drops the guard and `setState`s on every text change, so the trigger covers everything the view reads.

7-falseparfait: Thanks for the breakdown.. I Just merged PR #1 to drop the guard.

frostsnapp/lib/address.dart
```diff
@@ -29,6 +32,20 @@ class _CheckAddressPageState extends State<CheckAddressPage> {
     super.dispose();
   }
 
+  // Rebuilds on every change rather than on validity transitions: the error text
+  // reads both the field's emptiness and its validity, and `_isValidAddress` is
+  // false for both, so a validity-only trigger misses empty<->invalid entirely.
+  void _onTextChanged() {
+    final walletCtx = WalletContext.of(context);
+    final text = textInputController.text.trim();
```
LLFourn: The field is validated **trimmed** here, but `searchAddress()` still sends the **raw** text (`addressStr: textInputController.text`, line 57), and the coordinator parses it with a plain `Address::from_str` that doesn't trim (`frostsnap_coordinator/src/bitcoin/wallet.rs:352`). With a trailing newline — easy to get in this `maxLines: 6` field by pressing Enter after pasting, or by copying a whole line — the button is enabled, but the search can never match. The page says `Address not found in first 100 addresses.`, and a few "Search deeper" taps later it tells the user their own address "almost certainly doesn't belong to this wallet".

I checked the parse against bitcoin 0.32: `"bcrt1q…\n"` and `" bcrt1q…"` both fail `Address::from_str`, and the bare address parses. To be fair, this was already broken before your PR, since the old page never trimmed at all. But now the enabled button tells the user the input is fine, so it's worth closing here.

The smallest fix is `addressStr: textInputController.text.trim()`. What I'd prefer is to keep the string you validated and search with that, so the validator and the search can't drift apart again:

```dart
String? _validAddress;

void _onTextChanged() {
  final walletCtx = WalletContext.of(context);
  final text = textInputController.text.trim();
  setState(() {
    _validAddress = walletCtx != null &&
            Address.fromString(s: text, network: walletCtx.network) != null
        ? text
        : null;
  });
}
```

Then `searchAddress` searches `_validAddress`, the button is `onPressed: _validAddress == null ? null : …`, and `errorText` reads `textInputController.text.trim().isNotEmpty && _validAddress == null`. That last change also stops a whitespace-only entry from showing "Invalid address for this network", which it does now because the emptiness check on line 155 uses the untrimmed text.

PR #550, opened 2026-08-18
Let users consolidate coins a restore would miss, and stop losing them to a narrow window

LLFourn:
A coin whose keychain index sits past a long run of unused indices is recovery-fragile: a restore walks the gap-limit crawl and stops at the first gap wider than its lookahead, so anything above that gap is invisible to the restored wallet. #545 attacks this from the sending side, sweeping such coins along with every outgoing payment. This is the other half, and it turned out to be two separate things:

- a user with no payment to make had no way to act on it, and no way to know there was anything to act on;
- and change the `fee()` defect fixed in #542 sent far past the frontier is not just fragile but **invisible** — never attributed, so never in the balance and never spendable.

Two commits, reviewable in order.

## 1. Consolidate coins a restore would miss, without making a payment

A banner on the wallet home says there is something to act on, and its tap fixes it.

The count and total come from `gap_stranded_value`, which reports only coins worth rescuing: unspent, past the risky gap by #545's own crawl simulation, and of positive effective value at 10 sat/vB. A coin the rescue fee would eat is not worth nudging anyone about. The bar is 10 rather than 1 because the banner renders where no feerate is known.

`plan_consolidate` builds the remedy, and it is a shape the wallet could not previously express: **no recipient at all**, every selected coin in, one output back to the wallet's own change keychain. No coin selection — the input set is the point. Fee comes from the fixed set at the requested rate; a set that cannot pay its own way is refused rather than emitting sub-dust. Commit rides the existing `commit_send` unchanged, so the single output is ordinary change allocation.

Two agreement properties, both pinned by tests:

- **The banner and the plan read one source**, so the remedy clears the nudge by construction.
- **A coin effective-negative at the chosen feerate is still rescued** — pinned at a rate where that seam actually opens, since `plan_send` one screen up filters such coins as ordinary selection behaviour.

The review page borrows the send flow's own vocabulary rather than inventing a second one: the same `FeeRatePickerDialog`, the same nonce-aware signer selection, the same chrome. A user arriving from a warning should not have to learn a new screen to act on it. Two of send's summary widgets move to `theme.dart` so both read from one definition instead of drifting. The destination is deliberately not shown — it is change, allocated at commit, and does not exist until then. A rate-dependent plan failure keeps the feerate control on screen, the page's one degree of freedom.

## 2. Derive further than we subscribe, and tell the server when it pays off

The invisible-change half has a smaller cause than it looks. Nothing is missing from the transaction graph: a transaction is ours if we own one of its spent prevouts, so the change script is already sitting there as an unattributed txout. Only the attribution is missing, and `KeychainTxOutIndex` already does attribution — `_index_txout` reveals to whatever index a txout matched and replenishes the derived set past it. It simply was not deriving far enough to recognise the script.

What stopped us asking for a wider window is that **one number was serving two costs that have nothing in common.** Deriving a script is a secp operation against our own keys. Subscribing to one hands it to a server, buys per-block work for every client attached to it, and widens the fingerprint of the wallet. `DerivedSpkTracker` was constructed from `super_wallet.lookahead()`, so the local window could not grow without dragging the subscription window with it.

So they are separate constants now — the indexer derives 1000, the tracker subscribes 50 — and each says which cost is its own. That is the whole recovery mechanism. There is no scan, no heuristic, no new table, no API, and nothing across the frb bridge.

Wallets heal by two routes:

- **an upgraded database**, through the reindex in `lazily_initialize_key`. `insert_descriptor` is the one operation that leaves the index stale — every path that puts transactions *in* indexes them on the way — so that is the only reindex the wallet needs. Its reveals are now persisted rather than dropped: only a txout that actually paid us moves a frontier, and an index that has been paid is used whatever our records say.
- **a restored wallet**, through `apply_update`, as the crawl delivers each spend and it is indexed against the wide window as it lands.

The asymmetry creates one obligation, which is the other half of this commit. The two views no longer agree, so a frontier we reach locally is one the server has never been asked about — recovered change would be spendable while its own spend went unseen. `resync_monitoring` closes that at the two places a frontier moves for a reason the server did not tell us about. It is free to over-call: a tracked window only ever widens, equal-or-smaller is a no-op, and only scripts the tracker does not already hold become subscribe requests, so a wallet with nothing far out generates no traffic. That contract needs `bdk_electrum_streaming` 0.5.3 (evanlinjin/experiments#8, released for this) and two tests pin it against future dep bumps.

**1000 is a parameter, not a boundary.** Nothing records that a database has been searched, so raising it repairs every affected wallet on its next launch — no migration, no flag, nothing to version. What it bounds is reach: runs of used indices spaced under 1000 apart unwind however long they are, since each match extends the window again, and only an isolated gap wider than the constant defeats it.

### Rejected alternatives

Three other shapes were built or specified first, and the reasons they failed are worth recording:

- **A throwaway graph at a wide lookahead, run as a startup migration.** Correct for transactions already stored, useless for a restore — the graph is empty until the crawl fills it — and a live wallet's index is a load-once snapshot, so rows written behind its back stay invisible until the next launch.
- **A wide electrum stop gap.** The tracker is already a gap-limit crawl, so a wider one would work and would resolve chains by itself. It costs roughly 2100 subscriptions per key, a server-side history lookup for each, and hands the server 2100 of our scripts instead of 70 — a permanent privacy and bandwidth cost on every user, to repair a legacy defect.
- **A hook on transaction entry with a shape heuristic.** There is no viable predicate. Send-max already produces a spend of ours with no change of ours, and multiple recipients would make "two or more outputs, none of them ours" an ordinary transaction. Value accounting discriminates better but gains a false-negative class, since small stranded change looks like a plausible fee — trading wasted work for silently missed coins.

## Testing

Coordinator tests, each verified to fail for its own reason rather than assumed to. Setting the lookahead back to 50 fails all five of the new ones; dropping `resync_monitoring` fails only the signal test; dropping the reindex fails only the reopen test.

- far change is attributed and spendable;
- a reopened database rebuilds that attribution without another sync;
- the chain source is told where to look further — asserted at the message level by draining the tracking messages the client queued, which needs no electrum at all;
- a self-payment hides its far change no better than any other transaction, which is the case every rejected heuristic would have skipped;
- the local window is wider than what the server is told.

Plus the consolidation contract: exact coin set, single-output accounting at the requested rate, planning reveals nothing, sub-dust refusal, commit → broadcast clearing the stranded set, and the effective-negative seam.

End-to-end drives on regtest are a follow-up. What they would add beyond the above is only what a fixture cannot fake — a real server delivering the far change instead of a hand-built update, the actual recovery flow, and spending the coin for real.

---

Supersedes #548. #551 is stacked on this one.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/send.rs
+++ b/frostsnap_coordinator/src/bitcoin/send.rs
@@ -61,6 +61,23 @@ impl SendPlan {
         self.fee
     }
 
+    pub fn input_count(&self) -> usize {
+        self.selected.len()
+    }
+
+    /// The total value the plan spends. The plan pins outpoints, not values, so this is the
+    /// accounting identity fee + outputs rather than a wallet lookup — it is what the committed
+    /// transaction consumes even if the wallet's view moves after planning.
+    pub fn input_total(&self) -> u64 {
+        self.fee
+            + self.change_value.unwrap_or(0)
+            + self
+                .recipients
+                .iter()
+                .map(|txo| txo.value.to_sat())
+                .sum::<u64>()
+    }
+
     /// What this plan pays recipient `index`, the value the committed transaction carries.
     ///
     /// For send max this is the maximum as the selection fixed it, not as the wallet reports it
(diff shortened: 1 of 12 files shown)


LLFourn:
From the regtest e2e run (frames captured by the drive that also asserts them):

**The nudge** — a recovered coin at internal index 800, flagged after an app restart:

banner

**Its tap** — the dedicated consolidate page: total in, rate, fee, amount returning, nonce-aware signer selection:

consolidate page

**After one confirmation** — the banner cleared itself; the consolidation renders honestly as a −111 sat send (only the fee left the wallet), and the chain shows 0.74997889 BTC on change index 0:

after consolidation

LLFourn:
Recovery demo — regtest app-sim, real bitcoind + electrs, device-signed. Five steps, ~3.5s each, with a progress bar so it is followable:

1. A restored wallet holding 1 BTC, nothing amiss.
2. A wallet elsewhere on the same key spends it, sending change to change address #800 — far out of range of any restore's crawl.
3. Frostsnap finds that change locally and flags it: a future restore would miss this coin.
4. Tapping the banner offers to move it back within reach — nothing leaves but the fee.
5. Signed and confirmed: the coin is safe again on change index 0, and the warning clears itself.

recover and consolidate

LLFourn:
The consolidation review now speaks the send flow's visual language rather than inventing its own — same plain `labelLarge` row labels and monospace amounts, same `Fee [1.0 sat/vB]` chip, the feerate in send's own ETA button (tappable, with the same estimate spinner), and send's `Select Signers` card with its `N required` counter, key icons, no-nonces subtitle and `Sign transaction` button. The row label and chip moved to `theme.dart` so the two surfaces can't drift, and the signer card is driven by the same `BuildTxState` send uses, so selection and nonce handling are one implementation.

One deliberate difference: the feerate row stays live here instead of locking at the signer step, because replanning a fixed input set is free and the feerate is the only input a consolidation has.

consolidation review

LLFourn:
End-to-end on regtest, real bitcoind and electrs, device-signed — the whole path in one clip.

stranded change recovered and consolidated

A wallet receives 1 BTC at receive index 0. A transaction built **outside the app** — from the wallet's own exported descriptor, via RPC, never through the send flow — then spends that coin, paying 0.25 away and sending the change to **internal index 800**.

Index 800 is far past the 50 scripts the chain source subscribes to, so the server never reports it. The app holds the transaction anyway, because it owns one of its spent inputs, and the indexer derives 1000 scripts past the frontier — so the change is attributed and the balance is right: 0.74998 BTC. Before this PR that wallet showed **zero**.

The coin still sits past the gap a restore would crawl, so the banner offers to consolidate. The remedy is a transaction with no recipient at all: one coin in, one change output back.

The device pages are worth watching, and they are #546's disclosure doing its job on exactly the shape it was built for — change with no foreign recipient:

- `Send Amount #1 — 0.74997889 BTC — To Self`
- `To Address #1 — Change #0`
- `Network Fee 0.00000111 at 1.0 sat/vB` — with nothing leaving the wallet, the fee is the entire cost

Signed, broadcast, confirmed: the value lands on change index 0, inside any restore's reach, and the banner clears itself.

nickfarrow (approved):
Working great, did some manual testing and recoveries.

Non-blocking suggestions:
https://github.com/LLFourn/frostsnap/pull/23

frostsnap_coordinator/src/bitcoin/send.rs
```diff
@@ -61,6 +61,23 @@ impl SendPlan {
...
     }
 
+    pub fn input_count(&self) -> usize {
+        self.selected.len()
+    }
+
+    /// The total value the plan spends. The plan pins outpoints, not values, so this is the
+    /// accounting identity fee + outputs rather than a wallet lookup — it is what the committed
+    /// transaction consumes even if the wallet's view moves after planning.
+    pub fn input_total(&self) -> u64 {
```
LLFourn, line 108: I feel like it'd be better to just store the input value in "selected" and then have a constructor for this double checks everything adds up.

frostsnap_coordinator/src/bitcoin/send.rs
```diff
@@ -275,6 +292,124 @@ impl CoordSuperWallet {
...
+            })
+    }
+
+    /// Plan a consolidation: every coin the nudge counts goes in, and ONE output — change,
+    /// allocated by [`Self::commit_send`] like any other — comes back. There is no recipient
+    /// and no coin selection; the input set is the point. A coin effective-negative at
+    /// `feerate` is still included — the user is explicitly rescuing it, and leaving it behind
+    /// would leave the nudge standing after its own remedy. Refuses when the coins cannot pay
+    /// the fee and still leave a usable output.
+    pub fn plan_consolidate(
```
LLFourn, line 487: I think the api for a general consolidate should be the utxos to consolidate. In this case the outpoints should just be passed in and you can look them up from tx graph.

PR #551, opened 2026-08-18
Consolidate (n): whole-wallet consolidation from More Actions

LLFourn:
Stacked on #550 (review the tip commit).

#550 gives the stranded-coins nudge a remedy: merge the coins a restore would miss into one, with no payment involved. This generalises it to the obvious user request — **merge everything** — because nothing downstream of choosing the input set cares why the coins were chosen.

## Change

- `plan_consolidate` takes a `ConsolidationScope`: `Stranded` (the nudge's set, behaviour unchanged) or `All` (every unspent coin). The fee arithmetic, sub-dust refusal, commit boundary and review page are shared verbatim.
- `utxo_count` for the menu's badge and its gate.
- **"Consolidate (n)" in More Actions**, visible only past one coin, opening the same `ConsolidatePage` the banner opens — the page already took its input set as a planner closure, so this needed no page changes beyond an optional description line.

Two policy decisions, both deliberate and pinned by tests:

- **`All` includes coins effective-negative at the chosen rate.** The user asked to clean the wallet; leaving dust behind defeats that. The total-level fee/dust guard remains the safety boundary.
- **`All` refuses a single coin** — merging one coin into one coin only pays a fee — enforced in the wallet, not the menu, so a direct call or a tap racing a spend can't build it. `Stranded` stays unbounded there: moving one stranded coin back within a restore's reach is exactly the point.

Whole-wallet consolidation links every coin on-chain, so both surfaces say so rather than leaving it implied.

## Screenshots

The menu entry counting three coins, with the linking note:

menu entry

The shared review page — the send flow's own vocabulary (row labels, fee chip, ETA feerate button, signer card): 1.5 BTC in, 226 sat fee at 1.0 sat/vB, 1.49999774 returning:

review

## Testing

- Coordinator: `All` plans every coin (reachable, stranded, and dust not worth its own weight at the rate) into one output with the accounting identity holding and planning revealing nothing; the whole-wallet remedy also clears the stranded nudge and leaves exactly one coin; the single-coin refusal alongside `Stranded` still rescuing one; the sub-dust refusal is scope-independent and its wording no longer calls reachable coins "stranded".
- End-to-end on regtest (real bitcoind + electrs, device-signed): the action is absent at one coin, present as "Consolidate (3)", opens the shared page, signs, and after one confirmation the action gates itself off again while `scantxoutset` over the wallet's own exported descriptor sees exactly one unspent output. Both stranded-path scenarios still pass unchanged.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/send.rs
+++ b/frostsnap_coordinator/src/bitcoin/send.rs
@@ -342,6 +342,47 @@ impl CoordSuperWallet {
             .collect()
     }
 
+    /// The coins a whole-wallet consolidation spends: every one the wallet holds that is worth more
+    /// than it costs to move at `feerate`.
+    ///
+    /// A coin below its own spend cost is left where it is, because moving it destroys value —
+    /// 300 sats carried against roughly 575 of input cost at 10 sat/vB — and forfeits spending it
+    /// later at a rate where it does pay. That is the same coin [`Self::plan_send`] declines to
+    /// force and [`Self::calculate_avaliable_value`] leaves out of the spendable figure, so
+    /// including it here would spend value the wallet does not count as spendable.
+    ///
+    /// Filtered at the rate the user picked rather than a nominal one: a manual consolidation runs
+    /// behind the ordinary feerate picker, so the caller holds the real rate when it composes the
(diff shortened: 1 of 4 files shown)


nickfarrow (approved):
LGTM

PR #552, opened 2026-08-18
Only treat a PSBT output as ours if our key owns it

nickfarrow:
`index_of_spk` looks a script up across every wallet loaded in the app and tells you which one derives it. The PSBT output loop threw that answer away and stamped the signing wallet's key on it instead.

So with two wallets open, signing a PSBT that pays wallet B's address produced a transaction paying wallet A at the same index. Nothing is stealable here, but the transaction signed isn't the one the PSBT asked for.
 
Two commits: the first shares the test wallet fixture send.rs already had inline; the second is the fix, one file plus the first tests over `psbt_to_tx_template`. The test asserts that the signed transaction's outputs still equal the PSBT's, rather than how the output gets labelled -- remove the guard and it fails on exactly that.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/wallet.rs
+++ b/frostsnap_coordinator/src/bitcoin/wallet.rs
@@ -612,24 +612,26 @@ impl CoordSuperWallet {
             owned_count += 1;
         }
 
-        for (i, _) in psbt.outputs.iter().enumerate() {
-            let txout = &rust_bitcoin_tx.output[i];
+        for txout in &rust_bitcoin_tx.output {
+            // An owned output keeps a derivation path, not the script, and the template
+            // re-derives it: our appkey on another wallet's address repoints the payment at us.
             match self
                 .tx_graph
                 .index
                 .index_of_spk(txout.script_pubkey.clone())
             {
-                Some(&((_, account_keychain), index)) => template.push_owned_output(
-                    txout.value,
-                    LocalSpk {
-                        master_appkey,
-                        bip32_path: BitcoinBip32Path {
-                            account_keychain,
-                            index,
+                Some(&((owner, account_keychain), index)) if owner == master_appkey => template
+                    .push_owned_output(
+                        txout.value,
+                        LocalSpk {
+                            master_appkey,
+                            bip32_path: BitcoinBip32Path {
+                                account_keychain,
+                                index,
+                            },
                         },
-                    },
-                ),
-                None => {
+                    ),
+                _ => {
                     template.push_foreign_output(txout.clone());
                 }
             }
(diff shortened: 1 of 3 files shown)


LLFourn:
Ran this on `a539958b` and it holds up — the whole coordinator suite passes (36), and the bug is real.

The part worth spelling out, because the diff alone doesn't show it: an owned output stores a *derivation path*, and the template re-derives the script from it. So discarding `index_of_spk`'s answer about which wallet owns the address wasn't a labelling slip — it repointed the payment. Another loaded wallet's address at index *n* came back out as ours at index *n*, and the transaction signed isn't the one the PSBT asked for. Your comment on the guard says exactly that, which is the right thing for it to say.

Agreed on the severity too: `index_of_spk` only matches wallets loaded in the app, so both addresses are the user's own, and a genuinely foreign output takes the `None` branch untouched. Wrong destination, not a stealable one.

I checked the test discriminates rather than just passing — removed the `owner == master_appkey` guard and exactly one test fails, `a_psbt_output_paying_another_loaded_wallet_stays_foreign`. Asserting the signed outputs equal the PSBT's, rather than how an output gets labelled, is what makes it survive a refactor of the labelling.

One thing to know rather than fix: this conflicts with #540 in `wallet.rs`, which retypes indices across the same code. Whichever lands second needs a rebase — nothing wrong with either.

frostsnap_coordinator/src/bitcoin/wallet.rs
```diff
@@ -741,8 +743,72 @@ pub struct ConfirmationTime {
...
+        tweak::{AppTweak, DerivationPathExt},
+    };
+
+    fn appkey() -> MasterAppkey {
+        MasterAppkey::derive_from_rootkey(Point::random(&mut rand::thread_rng()))
+    }
+
+    #[test]
+    fn a_psbt_output_paying_another_loaded_wallet_stays_foreign() {
+        let (mut wallet, _handler) = test_wallet(bitcoin::Network::Bitcoin);
```
LLFourn, line 759: the testing approach here is suboptimal. We're dealing with internal details like lazily_initialize_key. The logic to be tested should be extracted into its own function.

LLFourn:
I am taking over this one.

LLFourn:
Superseded by #558, which removes `index_of_spk` from the PSBT output path entirely rather than correcting the key it stamps.

The output loop there takes its derivation from `PSBT_OUT_TAP_BIP32_DERIVATION` the way the input loop already does, and proves the claimed path by deriving the script from the master appkey. The fingerprint check that has always rejected a sibling wallet's *input* is then the same one that rejects its output, so this bug stops being expressible rather than being guarded against.

The test here — that the signed transaction's outputs still equal the PSBT's — is carried over, along with the wallet-free ones #558 adds around it.

Closing in favour of that; happy to reopen if you'd rather land this first.

PR #553, opened 2026-08-18
Keep the keychain when converting a bip32 path

nickfarrow:
A Frostsnap address path is four numbers: `account_kind / account_index / keychain / address_index`.

From<BitcoinBip32Path> for DerivationPath produced three. It reached through to the inner BitcoinAccount, which yields two segments, instead of BitcoinAccountKeychain, which yields three, then appended the index.

Receive `#7` and change `#7` therefore converted to the same path, and neither round-tripped: from_u32_slice requires exactly four segments.

Nothing was broken by this. No production code calls the conversion, and address derivation doesn't use it.

The fix uses the path's own four-segment method, the same one already feeding `AddressInfo.derivation_path` and the persistence layer. The test asserts receive and change convert differently and that both round-trip.

### DIFF
--- a/frostsnap_core/src/tweak.rs
+++ b/frostsnap_core/src/tweak.rs
@@ -62,13 +62,7 @@ impl BitcoinBip32Path {
 
 impl From<BitcoinBip32Path> for DerivationPath {
     fn from(bip32_path: BitcoinBip32Path) -> Self {
-        DerivationPath::from_normal_path_segments(
-            bip32_path
-                .account_keychain
-                .account
-                .path_segments_from_bitcoin_appkey()
-                .chain(core::iter::once(bip32_path.index)),
-        )
+        DerivationPath::from_normal_path_segments(bip32_path.path_segments_from_bitcoin_appkey())
     }
 }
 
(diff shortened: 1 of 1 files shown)


LLFourn:
Ran this on `e9770731` — `tweak` tests pass, and reverting the fix fails exactly the new test while `bip32_derivation_matches_rust_bitcoin` stays green. It discriminates.

I also checked the claim that carries the severity, since it's the difference between a latent defect and an address-derivation bug: there's no caller of `DerivationPath::from(BitcoinBip32Path)` outside `tweak.rs` in `frostsnap_core`, `frostsnap_coordinator` or `frostsnapp/rust`. So "nothing was broken by this" holds, and it's worth having said in the description rather than leaving a reader to guess how bad it was.

Asserting both that receive and change convert differently *and* that each round-trips is the right pair — either alone would pass on a conversion that was wrong in the other direction.

Heads up: this conflicts with #540 in `tweak.rs`, which retypes the index handling in the same file. Whichever lands second rebases.

nickfarrow:
Closing in favour of https://github.com/frostsnap/frostsnap/pull/555 (rebased onto ongoing changes)

PR #555, opened 2026-08-18
Delete the unused bip32 path to DerivationPath conversion

nickfarrow:
`From<BitcoinBip32Path> for DerivationPath` reached through to the account, so it emitted three segments and dropped the keychain. Receive `#7` and change `#7` converted to the same path, and neither round-tripped through `from_u32_slice`.

This PR originally fixed it. Deleting is better.

### DIFF
--- a/frostsnap_core/src/tweak.rs
+++ b/frostsnap_core/src/tweak.rs
@@ -160,18 +160,6 @@ impl BitcoinBip32Path {
     }
 }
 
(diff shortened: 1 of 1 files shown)


LLFourn:
If nothing is using this can we just delete it?

nickfarrow:
> If nothing is using this can we just delete it?

Updated this PR to delete.

PR #557, opened 2026-08-18
[coord] Don't panic on a signature share for an unknown session

average-gary:
The session lookup for an incoming SignatureShare was an unguarded `.expect()` on a session_id the device chooses. The id is checked after framing and decode but before any crypto verification, so nothing authenticates it: any connected device can panic the coordinator process by naming a session the coordinator doesn't hold. It is also reachable benignly, when the coordinator has forgotten a session and the device sends its share anyway.

Return `Error::coordinator_invalid_message`, as the neighbouring arms of the same match already do for a device sending something that doesn't fit the coordinator's state. The only production caller logs and continues, so a device losing this race isn't disconnected.

### DIFF
--- a/frostsnap_core/src/coordinator.rs
+++ b/frostsnap_core/src/coordinator.rs
@@ -871,10 +871,12 @@ impl FrostCoordinator {
                     ref replenish_nonces,
                 },
             ) => {
-                let active_sign_session = self
-                    .active_signing_sessions
-                    .get(&session_id)
-                    .expect("inavariant");
+                let active_sign_session = self.active_signing_sessions.get(&session_id).ok_or(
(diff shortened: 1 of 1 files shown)


LLFourn:
@average-gary thanks gary! That's definitely only an "invariant" if the device is honest or not misbehaving. I think that the test adding more code than it's worth. The better thing to do would be to write a proptest that actually generates a bunch of malformed messages and see what falls out. Right now we only have a proptest that genrates correct messages. I'll take the change without the test though and malformed/malicious proptest can come in another PR since it's a larger task.

average-gary:
Removed the tests in an amended commit.

LLFourn (approved):

PR #558, opened 2026-08-19
Read PSBTs and place signatures using the template, not the wallet index

LLFourn:
Three commits, each a self-contained fix, all with the same cause: asking a wallet index
what the transaction template already knew.

Supersedes #552, which fixes the first of these against the index rather than removing it.

### 1. Read a PSBT using only the PSBT and the key

`psbt_to_tx_template` took `&mut self` and never mutated. It read `self` exactly once, to
ask the BDK indexer which key derives a script pubkey — the only reason the logic was
welded to a wallet, and the reason it could only be exercised by standing one up.

Outputs now take their derivation from `PSBT_OUT_TAP_BIP32_DERIVATION`, exactly as inputs
already took theirs from `PSBT_IN_TAP_BIP32_DERIVATION`, and the claimed path is proved by
deriving the script from the master appkey and matching it against the output's own. Both
loops walk one shared ladder, so they cannot drift apart on what counts as ours.

That removes #552's bug rather than guarding it. `index_of_spk` searches every wallet
loaded in the app, so it could answer "whose is this?" about a key that is not signing —
with two wallets open, a PSBT paying wallet B was signed as paying wallet A. A fingerprint
on a PSBT-supplied path cannot do that: the check that now rejects a sibling wallet's
output is the same one that has always rejected a sibling wallet's input.

It also lifts the `revealed + 50` lookahead ceiling, which had made an output past it read
as a stranger's while an input at the same depth stayed signable.

Evidence for relying on the PSBT: a Sparrow PSBT and three from Bitcoin Core v31 on
regtest, committed as fixtures. Core annotates owned outputs on both keychains — the
payment as well as the change — and writes no derivation for an output it does not own.
The fixtures double as an independent check on our derivation: Core built those scripts
from the wallet descriptor, we rebuild them from the appkey and the claimed path, and a
mismatch is exactly what the new proof rejects.

### 2. Put each signature on the input it signed

A signing session produces one signature per *locally owned* input, so a signature's
position in the list is not an input index. Consumers paired the list against every input
with `zip`, which is correct only when the wallet owns all of them — true for the app's own
send flow, false for a PSBT. With a foreign input first, the signature was witnessed onto
the foreign input and ours was left empty.

`TransactionTemplate` already knew the answer and was never asked: `iter_locally_owned_inputs`
yields the index. The pairing now lives there, beside the `sign_items` that establishes the
order. A count that does not match is an error rather than the silent truncation `zip` was
performing.

The PSBT path had reconstructed ownership from `is_mine`, which answers "is this spk ours"
— a different question, and one that a foreign input paying a script we also send to gets
wrong.

### 3. Build every signing-session transaction from the template that signed it

`with_signatures` still had that positional zip, and it builds the transaction that gets
broadcast.

`UnsignedTx::details` asked the wallet what the template knew, twice: it classified scripts
with `spk_index`, so an owned output past the lookahead displayed as a stranger's; and it
sourced prevouts from the wallet's graph, which cannot hold a PSBT's foreign inputs. One
missing prevout makes `_sum_inputs` return `None`, so the review screen for those PSBTs had
no fee and no feerate. Both verified against real PSBTs rather than assumed.

Three sites assembled a `Transaction` field by field, two identical and the third quietly
different — which is how the divergent one survived. There is one constructor now.

### Testing

Behaviour is pinned where it is decided: the pure transform and the signature mapping in
`frostsnap_core` / `frostsnap_coordinator`, with the bridge layer testing only what those
cannot express — a missing template, and that `Transaction` sources from the template
rather than the wallet. `with_signatures` asserts agreement with the primitive rather than
restating it.

Every test was checked by reverting the behaviour it covers and confirming it fails.

### Note for #509

`nostr-taipei` routes both broadcast and PSBT export through `UnsignedTx::complete`, whose
`api/signing.rs` is `+0 -0` on that PR — so it currently funnels everything through the
positional zip. Commit 2 fixes that function, so the two should compose. `complete` is
deliberately kept here despite being unused on master.

The remaining conflict with #509 is `is_mine`'s value type. `owned_spks` returns the full
`BitcoinBip32Path`, so that PR's `(keychain, index)` change becomes a projection at the
call site rather than a change to the derivation.


LLFourn:
Replaced by #560 — same work with the `.clank/` plan files and commit tags stripped, rebased on current master, and `psbt_to_tx_template` reshaped into `TransactionTemplate::from_psbt`.

PR #562, opened 2026-08-19
[widgets,device] Show the firmware version on the standby screens

LLFourn:
A device could not tell you which firmware it was running. This puts the version
on the standby screens.

## Where the constant lives

An image can never contain its own digest, so the running version is not
recoverable from `KNOWN_FIRMWARE_VERSIONS` and has to be a literal until v0.5.0
embeds it at build time.

`FIRMWARE_VERSION` lives in the device crate — the running version is not a
protocol fact, so the coordinator and app never need it. A const assertion ties
it to `EARLIEST_ACCEPTABLE` so a release bump cannot move the downgrade floor
past the version this tree builds; it is checked at firmware build time because a
no_std riscv32 crate has nowhere to run a host test.

The literal itself sits in `device/firmware_version.rs` and is included verbatim
by both the device and the widget demos. `frostsnap_widgets` cannot depend on the
device crate — device depends on widgets — so inclusion is what lets the
simulator render the real version rather than keeping a second literal that would
drift at the next release. `Standby` takes the version to display rather than
reading a constant, so the widget stays a display layer.

## Before / after

Before on the left, after on the right.

**Welcome**

welcome

**Has key**

has key

**Recovery**

recovery

## Why the corner, not the column

The version first went at the foot of each content column. On the recovery
screen — warning row + key name + `Key #N` + device name + version — that
overflowed 280px and clipped the version against the bottom edge. As a top-right
overlay on `Standby` it fits every state, leaves both content columns exactly as
they were, and lives in one place rather than two.

It is `FONT_SMALL` in `text_secondary`, the same treatment `Key #N` gets, so it
reads as provenance rather than competing with the key name or the URL. Because
it hangs off `Standby` rather than the `FadeSwitcher`, it is also visible during
the startup state.

## Verification

- `frostsnap_widgets/tests/standby.rs` renders all four states and asserts the
  version corner is painted and that no state touches the bottom edge. Both were
  checked against mutations: removing the overlay fails the first, and moving the
  version into the content columns fails both, clipping on recovery.
- The const assertion was checked by setting `FIRMWARE_VERSION` below
  `EARLIEST_ACCEPTABLE`, which fails the firmware build as intended.
- The shared literal was checked by setting `device/firmware_version.rs` to
  `(9, 9, 9)` and confirming the widgets side renders `9.9.9`.
- riscv32 firmware builds from this branch: `cd device && cargo build --release --bin frontier`.
- `cargo test -p frostsnap_comms` (11) and `cargo test -p frostsnap_widgets` (34) pass.

Captures were taken with `tools/widget_simulator`, whose `screenshot` stdin
command already existed; they are hosted on a throwaway `standby-captures` branch
on my fork purely so they render here.

### DIFF
--- a/device/firmware_version.rs
+++ b/device/firmware_version.rs
@@ -1,0 +1,4 @@
+// Included verbatim by device/src/lib.rs and by frostsnap_widgets' demos, which
+// cannot depend on this riscv32-only crate. See FIRMWARE_VERSION in
+// device/src/lib.rs for why the version is a literal at all.
+(0, 4, 0)
--- a/device/src/frosty_ui.rs
+++ b/device/src/frosty_ui.rs
@@ -62,7 +62,10 @@ impl<'a> FrostyUi<'a> {
         use embedded_graphics::geometry::Size;
         use frostsnap_widgets::debug::EnabledDebug;
 
-        let root_widget = RootWidget::new(WidgetTree::Standby(Box::new(Standby::new())), 200);
+        let root_widget = RootWidget::new(
+            WidgetTree::Standby(Box::new(Standby::new(crate::FIRMWARE_VERSION))),
+            200,
+        );
         let debug_config = EnabledDebug {
             logs: cfg!(feature = "debug_log"),
             memory: cfg!(feature = "debug_mem"),
@@ -182,9 +185,11 @@ impl<'a> UserInteraction for FrostyUi<'a> {
 
         // Convert workflow to widget tree
         let new_page = match workflow {
-            Workflow::Startup => WidgetTree::Standby(Box::new(Standby::new())),
+            Workflow::Startup => {
+                WidgetTree::Standby(Box::new(Standby::new(crate::FIRMWARE_VERSION)))
+            }
             Workflow::None => {
-                let mut standby = Standby::new();
+                let mut standby = Standby::new(crate::FIRMWARE_VERSION);
                 standby.set_welcome();
                 WidgetTree::Standby(Box::new(standby))
             }
@@ -192,7 +197,7 @@ impl<'a> UserInteraction for FrostyUi<'a> {
                 device_name,
                 held_share,
             } => {
-                let mut standby = Standby::new();
+                let mut standby = Standby::new(crate::FIRMWARE_VERSION);
                 standby.set_key(device_name.to_string(), held_share);
                 WidgetTree::Standby(Box::new(standby))
             }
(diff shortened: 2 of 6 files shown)


frostsnap_comms/src/firmware_version.rs
```diff
@@ -80,6 +80,14 @@ impl FirmwareFeatures {
     }
 }
 
+/// The version of the firmware built from this tree.
+///
+/// A literal because an image can never contain its own digest, so a running
+/// firmware cannot look itself up in the table below. It is accurate because
+/// exactly one binary is ever signed as a given release. From v0.5.0 the
+/// version is embedded in the image at build time and this constant goes away.
+pub const FIRMWARE_VERSION: VersionNumber = VersionNumber::new(0, 4, 0);
```
LLFourn: i dont think this should be in comms. probably just in the device crate for now

LLFourn: Moved it to `device/src/lib.rs`. That also let `frostsnap_widgets` drop the `frostsnap_comms` dependency it had taken on solely to read the constant — `Standby::new` now takes the version to display, so the widget stays a display layer and the device supplies the fact.

The drift guard came along as a const assertion in the device build rather than a host test, since the crate is no_std riscv32. Verified by setting `FIRMWARE_VERSION` below `EARLIEST_ACCEPTABLE`, which fails the firmware build.

One duplication this forces, flagged rather than hidden: the demos in `demo_widget.rs` now carry their own literal, because `widget_simulator` builds them and cannot depend on a riscv32-only crate. Both halves retire together at v0.5.0.

LLFourn: Follow-up: the demo literal I flagged is gone. `device/firmware_version.rs` now holds the only `(0, 4, 0)` and is `include!`d by both `device/src/lib.rs` and the widget demos — no dependency edge, works on riscv32 and host alike, and the simulator renders the real version again.

Verified by setting the file to `(9, 9, 9)` and confirming the widgets side renders `9.9.9`, so they genuinely share one literal rather than agreeing by coincidence.

If the cross-crate `include!` is too cute for your taste, the alternative is the demos showing an obvious placeholder instead of a real version — say the word.

nickfarrow (approved):
Test improvement suggestion:
https://github.com/LLFourn/frostsnap/pull/24

nickfarrow:
One addition could be to have a dev device show its build. It could do so from the hash it calculates at boot.

PR #560, opened 2026-08-19
Read PSBTs and place signatures using the template, not the wallet index

LLFourn:
Reads a PSBT from the PSBT itself, and places every signature on the input that produced it. Three bugs with one cause — the wallet's script index was being asked *"do I know this script?"* when the question was *"is this ours, in this transaction?"*.

## The bugs

**Signatures landed on the wrong inputs.** `with_signatures` did `tx.input.iter_mut().zip(signatures)`, but signatures come back per *owned* input, in owned order. Any transaction with a foreign input — a PSBT from Sparrow, say — put our signature on someone else's input and left ours unsigned. The return type was `RTransaction`, not `Result`, so there was no error path: it handed you an invalid transaction and you broadcast it.

**#552 — a PSBT paying wallet B was labelled as paying wallet A.** Ownership came from `index_of_spk`, which searches every wallet loaded in the app, and the result went to the *infallible* `push_owned_output`. So an unproven claim became a label.

**An owned output past the lookahead was shown as a stranger's.** The index can only answer for scripts it has already derived.

## What changed

`TransactionTemplate::from_psbt` takes ownership from the PSBT's own key origins and *proves* each claim — the claimed path is derived from the key it names and matched against the actual script — so a PSBT cannot make us attribute a script to a key that does not produce it. The template is then the single authority on what is ours, and `signatures_by_input_index` is the only thing that decides where a signature goes.

`TransactionTemplate` gained a scope type parameter. `Unscoped` is the form that travels; `ScopedTo` is one narrowed to a key by `as_seen_by`, and every reader that answers an ownership question exists only there — asking an unscoped template whose inputs are whose is a compile error. Three compile-fail doctests pin it. The wire encoding is unchanged and a fixture proves it, because firmware parses those bytes.

`from_psbt` takes `&[MasterAppkey]`, so `Local` means "one of these, and the template records which" rather than being scoped in fact but not in type. Keys are indexed by fingerprint once; the map holds candidates because a fingerprint is four bytes, and the script decides.

[...]

PR #563, opened 2026-08-19
Make a transaction template's key scope a type

LLFourn:
Follow-up to #560, which this is stacked on — review that one first.

#560 made every `TransactionTemplate` reader correct by normalising the template to the
signing key in `WireSignTask::check`. That works, but the guarantee is a *convention*: one
call, in one function, and nothing stops the next reader taking a template that never went
through it. Both forms are the same type, so the compiler cannot tell them apart.

This makes it a type.

```rust
pub struct TransactionTemplate<S = Unscoped> { …, scope: S }

pub struct Unscoped;                  // the form that travels
pub struct ScopedTo(MasterAppkey);    // narrowed to one key by as_seen_by
```

Readers that answer an ownership question — `iter_locally_owned_*`, `owned_spks`,
`signatures_by_input_index`, `to_signed_rust_bitcoin_tx`, `has_any_inputs_to_sign`,
`feerate`, `foreign_recipients`, `user_prompt`, `attach_signatures_to_psbt` — exist only on
`TransactionTemplate<ScopedTo>`. Asking an unscoped template whose inputs are whose is a
compile error. `fee`, `txid`, `net_value` and `to_rust_bitcoin_tx` need no key and stay on
both.

**`ScopedTo` implements no bincode.** A scoped template cannot be serialized at all, so the
form that goes over the wire is the general one by construction rather than by discipline.
`WireSignTask` carries `Unscoped`; `SignTask`, which only exists after `check`, carries
`ScopedTo`.

### The wire bytes do not move

Established before anything was built on it, because firmware parses them.
`frostsnap_core/tests/wire_format.rs` pins the bincode encoding of a template using both
sides and both owner kinds, captured before the type parameter existed; it still
round-trips. `Unscoped` is a unit struct whose derived `Encode` writes nothing.

That fixture is worth having regardless of this PR: nothing previously pinned the encoding,
so an accidental change to `Input`, `Output`, `SpkOwner` or `LocalSpk` would have shipped
and broken un-upgraded firmware silently.

### What the compiler found that the convention had hidden

- Three `as_seen_by` calls in `check` re-scoping an already-scoped template.
- Every reader on the app's display path refused to compile until handed a scoped template
  — the same sites an earlier sweep had flagged as reading ownership from a template
  nothing had checked.
- `Transaction::from_template` lost its `master_appkey` parameter: the scope carries it.
- `details` and `effect` lost arguments that could *disagree* with the template they were
  reading. `effect` took a key while reading a template that names its own; disagreement was
  not benign, and its "we don't support spending from multiple different keys" arm is now
  unreachable and deleted.

### Testing

Two `compile_fail` doctests pin the guarantees themselves, with expected error codes
(`E0599`, `E0277`) so they cannot pass for an unrelated reason — verified by moving a reader
onto the general impl and watching them fail.

`frostsnap_core --no-default-features` still builds, since the device parses the wire form.

### DIFF
--- a/frostsnap_core/src/bitcoin_transaction.rs
+++ b/frostsnap_core/src/bitcoin_transaction.rs
@@ -14,8 +14,22 @@ use crate::{
     MasterAppkey,
 };
 
-/// Invalid state free representation of a transaction
-/// The scope a [`TransactionTemplate`] is in.
+/// Marks a template that has not been narrowed to any one key: the form that travels.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, bincode::Encode, bincode::Decode)]
+pub struct Unscoped;
+
+/// Marks a template narrowed to one key by [`TransactionTemplate::as_seen_by`], where
+/// `Local` can only mean that key.
+///
+/// Deliberately not `bincode::Encode`/`Decode`: a scoped template is a local conclusion, and
+/// making it unserializable means the form that goes over the wire is the general one by
+/// construction rather than by discipline.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub struct ScopedTo(pub MasterAppkey);
+
+/// Invalid state free representation of a transaction.
+///
+/// The type parameter is the scope: which key's transaction this is.
 ///
 /// An ownership question — which inputs are ours, which recipients are foreign — has no
 /// answer until the template names a key, so those readers exist only on
@@ -40,19 +54,6 @@ use crate::{
 /// There is deliberately no way back either: scoping erases another key's ownership, so a
 /// widened template would carry less than the one it came from. Anything that must send
 /// holds the unscoped form and narrows to read.
-/// Marks a template that has not been narrowed to any one key: the form that travels.
-#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, bincode::Encode, bincode::Decode)]
-pub struct Unscoped;
-
-/// Marks a template narrowed to one key by [`TransactionTemplate::as_seen_by`], where
-/// `Local` can only mean that key.
-///
-/// Deliberately not `bincode::Encode`/`Decode`: a scoped template is a local conclusion, and
-/// making it unserializable means the form that goes over the wire is the general one by
-/// construction rather than by discipline.
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
-pub struct ScopedTo(pub MasterAppkey);
-
 #[derive(Clone, Debug, bincode::Encode, bincode::Decode, Eq, PartialEq, Hash)]
 pub struct TransactionTemplate<S = Unscoped> {
     scope: S,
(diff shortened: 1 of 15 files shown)


frostsnap_core/src/bitcoin_transaction.rs
```diff
@@ -202,54 +311,45 @@ impl TransactionTemplate {
         self.push_owned_output(txout.value, owner);
         Ok(())
     }
+}
 
-    pub fn to_rust_bitcoin_tx(&self) -> bitcoin::Transaction {
-        bitcoin::Transaction {
+impl TransactionTemplate<Unscoped> {
```
nickfarrow: nit: continued impls for Unscoped, don't need to close and re-open

frostsnap_core/tests/wire_format.rs
```diff
@@ -0,0 +1,97 @@
...
+    schnorr_fun::fun::{g, G},
+    tweak::{BitcoinBip32Path, NormalIndex},
+    MasterAppkey,
+};
+
+fn idx(n: u32) -> NormalIndex {
+    NormalIndex::new(n).unwrap()
+}
+
+/// Both sides, both owner kinds, non-default version and locktime.
```
nickfarrow: `set_version` is never called, so the template carries `new()`'s default `Version::TWO`

frostsnap_core/src/bitcoin_transaction.rs
```diff
@@ -88,9 +128,82 @@ impl Default for TransactionTemplate {
...
+
+    pub fn fee(&self) -> Option<u64> {
+        self.inputs
+            .iter()
+            .map(|input| input.value)
+            .sum::<u64>()
+            .checked_sub(self.outputs.iter().map(|output| output.value).sum())
+    }
+
+    pub fn net_value(&self) -> BTreeMap<RootOwner, i64> {
```
nickfarrow: `fn net_value()` gives a different answer depending on which template you call it on. On the unscoped one, a second wallet of ours shows up as ours. On the scoped one, it shows up as a stranger.

`fn effect()` takes on the Scoped path: after pulling out our own key it treats everything else as a recipient. Safe because line 147 narrows first, what the comment on 157 of `signing.rs` are getting at.

`effect()` is the only caller and it already narrows, so `net_value` could live on `impl TransactionTemplate<ScopedTo>`.

nickfarrow (approved):

frostsnap_core/src/bitcoin_transaction.rs
LLFourn: Good idea I'm going to rework the naming of these methods

frostsnap_core/src/psbt.rs
LLFourn: shouldn't this one's return value be scoped since we've passed in the MasterAppKey.

On the other hand we could convert this to a &[MasterAppKey] and pass in all the keys of the wallet and then it could be unscoped. This would be fine since devices scope it to themselves while validating. Also any effects of the transaction would be scoped or unscoped and clear in the situation.

This doesn't mean we'd necessarily have to pass all that exist in this PR we can just pass one but it makes the API less weird if we generalize it to taking many keys.

frostsnap_core/src/bitcoin_transaction.rs
LLFourn: There should no longer be ambiguity.

PR #567, opened 2026-08-20
An output with no address is a fact, not a panic

LLFourn:
`user_prompt` unwrapped `Address::from_script` for every disclosed output. An
OP_RETURN has no address rendering, `from_psbt` records anything it does not
recognise as foreign, and `WireSignTask::check` validates only the paths of
outputs claimed as ours — so a PSBT carrying a data output reached the signing
prompt and panicked the device.

`PromptRecipient` carries a `PromptDestination` now, either an `Address` or an
`UnrecognizedScript` holding the script. The variant is named for what is known:
`from_script` failing means rust-bitcoin has no address form, not that nobody can
spend the output — a bare multisig fails it and is spendable — and the one screen
a user is meant to trust must not claim more than it knows. There is a test using
exactly that script, asserting `!is_op_return`, so the name cannot drift back.

The device gets a page for it, so an output it cannot render still takes its place
in the sequence and the page arithmetic is unchanged; an output the signer never
sees would be the one outcome worse than an awkward one. The script's bytes are
not drawn — `AddressDisplay` earns trust by being checkable against another
source, and arbitrary hex cannot be.

`EffectOfTx.foreign_receiving_addresses` was the same defect on the app side and
three more besides: addresses pre-rendered to `String`, keyed by script so two
payments to one address merged into one row, netted and filtered to `> 0` so a
party who netted zero vanished. It is `foreign_outputs: Vec<TxOutInfo>` now, built
from `Transaction::recipients()` — not a new type, the one the neighbouring view
already used, whose `address(network)` answers `Option`. All four defects went at
once because each was a consequence of keeping a second model rather than a
separate mistake. `foreign_net_values` lost its only caller and is deleted, and
`effect` stopped taking a network: an address is a rendering, and rendering
belongs to whoever displays.

That closes the device/app divergence rather than documenting it. The assertion in
`another_keys_scripts_become_foreign` that pinned the two surfaces disagreeing
about one address — 150_000 gross against 50_000 net — now pins them agreeing.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/frostsnap_core/src/bitcoin_transaction.rs
+++ b/frostsnap_core/src/bitcoin_transaction.rs
@@ -465,31 +465,6 @@ impl TransactionTemplate<ScopedTo> {
                 .sum::<i64>()
     }
 
-    /// Net value flowing to each script that is not ours, negative where a foreign party
-    /// spends into this transaction.
-    ///
-    /// Distinct from [`Self::foreign_recipients`], which reports what an output pays
-    /// regardless of what its owner also spends.
-    pub fn foreign_net_values(&self) -> BTreeMap<ScriptBuf, i64> {
-        let mut net: BTreeMap<ScriptBuf, i64> = Default::default();
-
-        for input in &self.inputs {
-            if let SpkOwner::Foreign(spk) = &input.owner {
-                *net.entry(spk.clone()).or_default() -=
-                    i64::try_from(input.value).expect("value ridiculously large");
-            }
-        }
-
-        for output in &self.outputs {
-            if let SpkOwner::Foreign(spk) = &output.owner {
-                *net.entry(spk.clone()).or_default() +=
-                    i64::try_from(output.value).expect("value ridiculously large");
-            }
-        }
-
-        net
-    }
-
     pub fn foreign_recipients(&self) -> impl Iterator<Item = (&Script, u64)> {
         self.outputs
             .iter()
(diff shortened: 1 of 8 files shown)


LLFourn:
Screen recording of the device prompt for a transaction whose outputs have no address form.

device signing prompt for an OP_RETURN and a bare multisig

One regtest transaction carrying three disclosed outputs: a plain payment (the control), an OP_RETURN, and a bare multisig — recorded together so the two unrenderable cases sit in the same page sequence. All eight pages:

all eight prompt pages

**It does what it says.** No panic, and no output disappears: 8 pages, exactly `recipients*2 + 2`. bitcoind built the spend (`walletcreatefundedpsbt` against the watch-only descriptor), it came in through the QR scanner, two of three devices signed, and it broadcast and confirmed.

**The two cases are indistinguishable** — 68 pixels apart, and those are the index digit. An unspendable data carrier and a spendable bare multisig get the identical screen. Deliberate, and honest, but it does mean the page says nothing about which one you are signing.

**The header contradicts the body.** The page reads `To Address #2` above `Unrecognized script`. For an OP_RETURN there is no address and there never could be, so that header tells a signer a recipient address exists which the device failed to parse — it reads as a malfunction rather than "this is a data output". The body is the honest half; `UnrecognizedScriptPage::new` keeps `AddressPage`'s `format!("To Address #{}", index + 1)`, which claims the one thing the page is trying not to claim.

The nearest real hint is on a different page: the amount reads `0.00 000 000 BTC`. That is the strongest available signal it is a data output, but you have to hold two pages in your head to draw it.

Two notes on the artifact: it is stitched device framebuffers (240×280, 2× nearest-neighbour), not a video — `screenrecord` is Android-only and there is no emulator on this machine. And the `1.9 sats/vB` on the fee page is the harness, not this PR: bitcoind sized the fee for a placeholder output before the bare-multisig script was written over it.

nickfarrow (approved):

PR #571, opened 2026-08-20
Don't add TOFU certs to the PKI root store

nickfarrow:
`create_combined_cert_store` put every TOFU-trusted certificate into the base verifier's `RootCertStore`, where webpki treats it as an unconstrained CA -- basicConstraints and keyUsage are never checked on an anchor. A leaf trusted for one host could then mint a certificate for any other host, with no TOFU prompt.

Nothing is lost: `verify_server_cert` matches stored certs byte-for-byte and returns before the base verifier runs, so the root store only ever applied to hosts with no TOFU entry.

### DIFF
--- a/frostsnap_coordinator/src/bitcoin/tofu/connection.rs
+++ b/frostsnap_coordinator/src/bitcoin/tofu/connection.rs
@@ -172,8 +172,12 @@ async fn connect_with_tofu(
     host: &str,
     trusted_certificates: &mut Persisted<TrustedCertificates>,
 ) -> Result<TlsStream<TcpStream>, TofuError> {
-    // Create combined certificate store with PKI roots and TOFU certs
-    let root_store = trusted_certificates.create_combined_cert_store();
+    // webpki roots only. TOFU certs must never go in here: webpki treats every root as an
+    // unconstrained CA regardless of its basicConstraints, so a TOFU'd leaf would let whoever
+    // holds its key mint a cert for any other host. TOFU trust is per-host exact match in
+    // `TofuCertVerifier`, which runs before this verifier is ever consulted.
+    let mut root_store = rustls::RootCertStore::empty();
+    root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
     let base_verifier = WebPkiServerVerifier::builder(Arc::new(root_store))
         .build()
         .map_err(|e| TofuError::Other(anyhow!("Failed to create verifier: {:?}", e)))?;
--- a/frostsnap_coordinator/src/bitcoin/tofu/trusted_certs.rs
+++ b/frostsnap_coordinator/src/bitcoin/tofu/trusted_certs.rs
@@ -1,7 +1,6 @@
 use anyhow::Result;
 use bdk_chain::{bitcoin, rusqlite_impl::migrate_schema};
 use rusqlite::params;
-use rustls::RootCertStore;
 use rustls_pki_types::CertificateDer;
 use std::collections::HashMap;
 use std::time::{SystemTime, UNIX_EPOCH};
@@ -102,25 +101,6 @@ impl TrustedCertificates {
             },
             mutations,
         );
-    }
-
-    pub fn create_combined_cert_store(&self) -> RootCertStore {
-        // Start with standard PKI certificates from webpki-roots
-        let mut store = RootCertStore::empty();
-        store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
-
-        // Add user-trusted certificates
-        for (server_url, trusted_cert) in &self.certificates {
-            if let Err(e) = store.add(trusted_cert.certificate.clone()) {
-                tracing::warn!(
-                    "Failed to add trusted certificate for {}: {:?}",
-                    server_url,
-                    e
-                );
-            }
-        }
-
-        store
     }
 
     /// Find a certificate by its SHA256 fingerprint (raw hex, no colons)
--- a/frostsnap_coordinator/src/bitcoin/tofu/verifier.rs
+++ b/frostsnap_coordinator/src/bitcoin/tofu/verifier.rs
@@ -89,7 +89,8 @@ impl From<anyhow::Error> for TofuError {
 ///    - Most real-world attacks would produce certificates that fail validation entirely
 ///
 /// 4. **What We DO Guarantee**:
-///    - Certificates are properly scoped to specific servers (no certificate confusion)
+///    - Certificates are properly scoped to specific servers (no certificate confusion). This
+///      relies on TOFU certs never entering the PKI root store — see `connect_with_tofu`.
 ///    - Once trusted, we detect any changes to the certificate
 ///    - We don't capture certificates with structural problems
 pub struct TofuCertVerifier {
@@ -413,8 +414,9 @@ mod tests {
 
     #[test]
     fn test_certificate_confusion_attack_prevention() {
-        // This test should FAIL if the vulnerability exists
-        // A certificate accepted for one server should NOT be accepted for a different server
+        // A certificate accepted for one server should NOT be accepted for a different server.
+        // Only covers byte-for-byte reuse: TEST_CERT1 has no SAN for "bank.com" so it fails name
+        // checking anyway, which is why this can't detect PKI root store leakage on its own.
 
         let mut trusted_certs =
             TrustedCertificates::new_for_test(bdk_chain::bitcoin::Network::Bitcoin);


frostsnap_coordinator/src/bitcoin/tofu/verifier.rs
```diff
@@ -413,8 +414,9 @@ mod tests {
 
     #[test]
     fn test_certificate_confusion_attack_prevention() {
-        // This test should FAIL if the vulnerability exists
-        // A certificate accepted for one server should NOT be accepted for a different server
+        // A certificate accepted for one server should NOT be accepted for a different server.
+        // Only covers byte-for-byte reuse: TEST_CERT1 has no SAN for "bank.com" so it fails name
+        // checking anyway, which is why this can't detect PKI root store leakage on its own.
```
LLFourn: This comment is honest and it is also the thing I would most want addressed: after this PR, **nothing in the test suite fails if the TOFU certs go back into the root store**. The change is a deleted function plus prose, so the property it establishes is untested.

I reproduced both directions against this head to be sure the fix is the fix:

- **Pre-PR store** (webpki roots + `store.add(tofu_cert)`, i.e. what `create_combined_cert_store` returned): a self-signed leaf for `attacker.com` with `basicConstraints critical CA:FALSE` and a `keyUsage` *without* `keyCertSign` was accepted by `RootCertStore::add` as an anchor. A `bank.com` leaf signed with that key then returned `Ok(ServerCertVerified(()))` with **no TOFU error captured**.
- **This PR's store** (webpki roots only): `Err(InvalidCertificate(UnknownIssuer))`, and the cert is captured for a TOFU prompt.

A regression test that fails loudly if the combined store ever returns — new file `frostsnap_coordinator/tests/tofu_root_store.rs`:

```rust
use frostsnap_coordinator::bitcoin::tofu::trusted_certs::TrustedCertificates;
use frostsnap_coordinator::bitcoin::tofu::verifier::TofuCertVerifier;
use frostsnap_coordinator::persist::Persist;
use rustls::client::danger::ServerCertVerifier;
use rustls::client::WebPkiServerVerifier;
use rustls_pki_types::{CertificateDer, ServerName, UnixTime};
use std::sync::Arc;
use tempfile::NamedTempFile;

const ATTACKER: &[u8] = include_bytes!("certs/attacker_leaf.der");
const BANK: &[u8] = include_bytes!("certs/bank_leaf.der");

/// Inside the fixtures' validity window (2025-01-01 .. 2027-01-01).
fn now() -> UnixTime {
    UnixTime::since_unix_epoch(std::time::Duration::from_secs(1_767_225_600))
}

fn attacker_trusted() -> TrustedCertificates {
    let tf = NamedTempFile::new().unwrap();
    let mut conn = rusqlite::Connection::open(tf.path()).unwrap();
    TrustedCertificates::migrate(&mut conn).unwrap();
    let mut tc = TrustedCertificates::load(&mut conn, bdk_chain::bitcoin::Network::Signet).unwrap();
    tc.add_certificate(
        CertificateDer::from(ATTACKER.to_vec()),
        "attacker.com".to_string(),
        &mut Vec::new(),
    );
    tc
}

/// A TOFU'd leaf must not be able to mint a certificate for a host it was never trusted for.
/// Fails if TOFU certs are ever put back into the verifier's `RootCertStore`.
#[test]
fn a_tofud_cert_cannot_vouch_for_another_host() {
    let mut roots = rustls::RootCertStore::empty();
    roots.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
    let base = WebPkiServerVerifier::builder(Arc::new(roots)).build().unwrap();
    let v = TofuCertVerifier::new(base, attacker_trusted());

    let bank = CertificateDer::from(BANK.to_vec());
    let name = ServerName::try_from("bank.com").unwrap();
    assert!(v.verify_server_cert(&bank, &[], &name, &[], now()).is_err());
    assert!(
        v.take_tofu_error("bank.com").is_some(),
        "an unvouched-for cert belongs in a TOFU prompt, not in the trust store"
    );
}

/// The other half: removing the certs from the root store loses nothing.
#[test]
fn a_tofud_cert_is_still_accepted_for_its_own_host() {
    let mut roots = rustls::RootCertStore::empty();
    roots.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
    let base = WebPkiServerVerifier::builder(Arc::new(roots)).build().unwrap();
    let v = TofuCertVerifier::new(base, attacker_trusted());

    let attacker = CertificateDer::from(ATTACKER.to_vec());
    let name = ServerName::try_from("attacker.com").unwrap();
    assert!(v.verify_server_cert(&attacker, &[], &name, &[], now()).is_ok());
}
```

Both pass on this head; the first returns `Ok` (i.e. fails) if you restore `create_combined_cert_store`. I ran them here with no new dependency — the fixtures follow the existing `tests/certs/*.der` pattern:

```sh
# attacker.cnf: CN=attacker.com, basicConstraints=critical,CA:FALSE,
#   keyUsage=critical,digitalSignature,keyEncipherment, extendedKeyUsage=serverAuth,
#   subjectAltName=DNS:attacker.com
openssl ecparam -genkey -name prime256v1 -noout -out attacker.key
openssl req -new -x509 -key attacker.key -out attacker.pem -sha256 -set_serial 1 \
  -config attacker.cnf -not_before 20250101000000Z -not_after 20270101000000Z
openssl x509 -in attacker.pem -outform DER -out tests/certs/attacker_leaf.der

# bank.cnf: CN=bank.com, same extensions but subjectAltName=DNS:bank.com
openssl ecparam -genkey -name prime256v1 -noout -out bank.key
openssl req -new -key bank.key -out bank.csr -config bank.cnf
openssl x509 -req -in bank.csr -CA attacker.pem -CAkey attacker.key -set_serial 2 \
  -extfile bank.cnf -extensions v3 -sha256 \
  -not_before 20250101000000Z -not_after 20270101000000Z -out bank.pem
openssl x509 -in bank.pem -outform DER -out tests/certs/bank_leaf.der
```

Note `openssl verify` rejects that chain outright (`key usage does not include certificate signing`) — which is the point: webpki accepts it only because the signer is an *anchor*, and anchors skip those checks.

While you are here, the test name above this comment still says `test_certificate_confusion_attack_prevention` while the comment now admits it does not cover the confusion attack that actually existed. Worth renaming it to what it does test (byte-for-byte reuse across hosts) once the real one exists.

frostsnap_coordinator/src/bitcoin/tofu/connection.rs
```diff
@@ -172,8 +172,12 @@ async fn connect_with_tofu(
...
     trusted_certificates: &mut Persisted<TrustedCertificates>,
 ) -> Result<TlsStream<TcpStream>, TofuError> {
-    // Create combined certificate store with PKI roots and TOFU certs
-    let root_store = trusted_certificates.create_combined_cert_store();
+    // webpki roots only. TOFU certs must never go in here: webpki treats every root as an
+    // unconstrained CA regardless of its basicConstraints, so a TOFU'd leaf would let whoever
+    // holds its key mint a cert for any other host. TOFU trust is per-host exact match in
+    // `TofuCertVerifier`, which runs before this verifier is ever consulted.
+    let mut root_store = rustls::RootCertStore::empty();
+    root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
```
LLFourn: This is the right change and the comment above it is a good one. The structural problem is that the comment is now the *only* thing holding the invariant, and it is attached to two lines buried inside a connection helper.

`root_store.extend(webpki_roots::TLS_SERVER_ROOTS...)` appears eight times in this crate: here, five times in `verifier.rs` tests (322, 360, 389, 429, 525) and twice in `tests/tofu_tests.rs` (35, 63). Each test builds its own equivalent store, which means **no test exercises the store this function builds**. Put the certs back on line 180 and the suite stays green.

Giving the invariant one home fixes both halves at once — something like:

```rust
/// Trust anchors for hosts with no TOFU entry.
///
/// TOFU certs must never go in here: webpki treats every anchor as an unconstrained CA
/// regardless of its basicConstraints, so a TOFU'd leaf would let whoever holds its key
/// mint a cert for any other host. TOFU trust is per-host exact match in `TofuCertVerifier`,
/// which runs before this verifier is ever consulted.
pub(crate) fn pki_base_verifier() -> Result<Arc<WebPkiServerVerifier>, VerifierBuilderError> {
    let mut root_store = rustls::RootCertStore::empty();
    root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
    WebPkiServerVerifier::builder(Arc::new(root_store)).build()
}
```

Then `connect_with_tofu` and all seven test sites call it, the tests are testing production's store rather than a lookalike, and the regression test in the other comment actually guards the code that ships. It also deletes seven copies of the same four lines.

frostsnap_coordinator/src/bitcoin/tofu/trusted_certs.rs
```diff
@@ -104,25 +103,6 @@ impl TrustedCertificates {
...
-                    e
-                );
-            }
-        }
-
-        store
-    }
-
     /// Find a certificate by its SHA256 fingerprint (raw hex, no colons)
     pub fn find_certificate_by_fingerprint(
```
LLFourn: Good deletion — this removes a second, weaker representation of the same trust, which is what made the bug possible: a TOFU cert was both a per-host pin *and* an unconstrained PKI anchor, and only one of those was per-host.

The lookup immediately below is the last API on this type with that shape. `find_certificate_by_fingerprint` finds a certificate by fingerprint alone, across every host in the store — trust keyed by something other than the host, which is exactly the confusion you just closed off. It also has no callers anywhere in the repo:

```
$ grep -rn --include='*.rs' --include='*.dart' find_certificate_by_fingerprint .
frostsnap_coordinator/src/bitcoin/tofu/trusted_certs.rs:107:    pub fn find_certificate_by_fingerprint(
```

Worth deleting in this PR rather than leaving a host-agnostic lookup sitting in the store's public API waiting to be wired into an accept path. `get_all_certificates` (line 117) is in the same position — only ever called from tests — though that one is at least harmless and is the obvious basis for a future "manage trusted certificates" screen, so I would keep it.

frostsnap_coordinator/src/bitcoin/tofu/verifier.rs
LLFourn: 🤖codex🤖 Please reconcile this sample with the structural comment before publishing. As shown, the integration test manually constructs a webpki-only store, so restoring create_combined_cert_store on the production path leaves the test green; the claim that it fails if the combined store returns is therefore false. Have the regression test call the same production helper as connect_with_tofu, or move it into an in-crate test module where a pub(crate) helper is accessible.

frostsnap_coordinator/src/bitcoin/tofu/connection.rs
LLFourn: 🤖codex🤖 The one-home direction is right, but this exact helper cannot support the stated call sites: pub(crate) is inaccessible from tests/tofu_root_store.rs, and a helper returning Arc<WebPkiServerVerifier> cannot replace the two tests/tofu_tests.rs sites that need a RootCertStore for with_root_certificates. Please make the recommendation internally consistent—for example, keep the regression test in an in-crate module and call the pub(crate) helper there, and do not claim the unrelated integration-test sites can all call this verifier helper.

LLFourn (approved):

PR #583, opened 2026-08-22
Fill the progress bar background so it erases

nickfarrow:


LLFourn:
This one is not necessary right now since progress never goes backwards.

PR #587, opened 2026-08-23
app: support Base64 and Hex encoded PSBTs (fixes #569)

### LINKED ISSUE #569: App can't read base64 PSBTs
Bitcoin Core's `walletcreatefundedpsbt` emits base64.

7-falseparfait:
Fixes #569

Bitcoin Core and several other wallets export PSBTs as Base64 text by default (`walletcreatefundedpsbt`), but the app previously expected raw binary starting with `psbt\xff`.

## Changes

- Added `deserialize_psbt` in `frostsnapp/rust/src/api/bitcoin.rs` to auto-detect the input format: it checks for the raw magic bytes first, then falls back to decoding trimmed Base64 or hex text before handing off to rust-bitcoin.
- Updated `psbt.dart` to route through `deserializePsbt`.
- Added the `base64` and `hex` crates to `frostsnapp/rust/Cargo.toml`.
- Added 4 unit tests covering binary, Base64, hex, and invalid payloads.

## Testing

- `cargo test -p rust_lib_frostsnapp`: all tests passing.
- Replicated the CI jobs locally: `lint-ordinary`, `lint-device`, firmware build + stack check (frontier), and `lint-app` all green; full `test-ordinary --release --all-features --locked` suite passes.

### DIFF
--- a/frostsnapp/rust/Cargo.toml
+++ b/frostsnapp/rust/Cargo.toml
@@ -8,6 +8,7 @@ crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
 anyhow = "1"
+base64 = "0.22"
 frostsnap_coordinator = { workspace = true }
 frostsnap_core = { workspace = true }
 lazy_static = "1.4"
@@ -24,6 +25,7 @@ rand = "0.8"
 time = { version = "0.3.36", features = ["formatting"] }
 bip21 = {version = "0.5.0", default-features = false}
 futures.workspace = true
+hex = "0.4"
 
 [build-dependencies]
 frostsnap_secure_boot.workspace = true
--- a/frostsnapp/rust/src/api/bitcoin.rs
+++ b/frostsnapp/rust/src/api/bitcoin.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use base64::{engine::general_purpose::STANDARD as BASE64_STANDARD, Engine};
 pub use bitcoin::Transaction as RTransaction;
 pub use bitcoin::{
     psbt::Error as PsbtError, Address, Network as BitcoinNetwork, OutPoint, Psbt, ScriptBuf, TxOut,
@@ -546,10 +547,35 @@ impl AddressExt for bitcoin::Address {
 impl Psbt {
     #[frb(sync)]
     pub fn serialize(&self) -> Vec<u8> {}
+}
 
-    #[frb(sync)]
-    #[allow(unused)]
-    pub fn deserialize(bytes: &[u8]) -> Result<Psbt, PsbtError> {}
+const PSBT_MAGIC: &[u8] = b"psbt\xff";
+
+/// Deserializes a binary PSBT or its Base64/hex text representation.
+#[frb(sync)]
+pub fn deserialize_psbt(bytes: &[u8]) -> std::result::Result<Psbt, PsbtError> {
+    if bytes.starts_with(PSBT_MAGIC) {
+        return Psbt::deserialize(bytes);
+    }
+
+    // `from_utf8_lossy` leaves valid text unchanged and lets malformed input
+    // gracefully fall through to the usual rust-bitcoin deserialization error.
+    let text = String::from_utf8_lossy(bytes);
+    let text = text.trim().trim_matches('"').trim();
+
+    if let Ok(decoded) = BASE64_STANDARD.decode(text) {
+        if decoded.starts_with(PSBT_MAGIC) {
+            return Psbt::deserialize(&decoded);
+        }
+    }
+
+    if let Ok(decoded) = hex::decode(text) {
+        if decoded.starts_with(PSBT_MAGIC) {
+            return Psbt::deserialize(&decoded);
+        }
+    }
+
+    Psbt::deserialize(bytes)
 }
 
 #[frb(external)]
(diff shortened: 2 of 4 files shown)


7-falseparfait:
hey @LLFourn, just following up... checks are passing, and i wanted to see if there's anything i should address or update on my end.

frostsnapp/rust/src/api/bitcoin.rs
```diff
@@ -546,10 +547,35 @@ impl AddressExt for bitcoin::Address {
...
+}
 
-    #[frb(sync)]
-    #[allow(unused)]
-    pub fn deserialize(bytes: &[u8]) -> Result<Psbt, PsbtError> {}
+const PSBT_MAGIC: &[u8] = b"psbt\xff";
+
+/// Deserializes a binary PSBT or its Base64/hex text representation.
+#[frb(sync)]
+pub fn deserialize_psbt(bytes: &[u8]) -> std::result::Result<Psbt, PsbtError> {
```
LLFourn: Could we decide the encoding from the first few bytes instead of trial-decoding the whole input? That's what Core, Electrum and Coldcard do: `psbt\xff` is binary, `cHNidP` is base64 (it's the base64 of `psbt\xff`), and anything else is rejected straight away.

As written, a file that isn't a PSBT at all (someone picks a photo by mistake) goes through `from_utf8_lossy`, then a base64 decode, then a hex decode, and all of that happens inside a `#[frb(sync)]` call on the UI thread. In a release build, 4 MiB of non-text bytes took about 9 ms and 64 MiB about 145 ms (debug builds are much slower). Before this change the same file failed on a magic-byte compare.

While you're here, stripping ASCII whitespace and a leading UTF-8 BOM before decoding would also make these work:
- line-wrapped base64. `base64 tx.psbt` wraps at 76 columns and `openssl base64` at 64, and `BASE64_STANDARD.decode` rejects the embedded newlines.
- a file saved by Notepad. `str::trim` doesn't remove U+FEFF.

It also lets us drop the `trim().trim_matches('"').trim()` stack. Something like:

```rust
pub fn parse_imported_psbt(bytes: &[u8]) -> std::result::Result<Psbt, PsbtError> {
    let text = bytes.strip_prefix(b"\xEF\xBB\xBF").unwrap_or(bytes).trim_ascii_start();
    if text.starts_with(b"cHNidP") {
        let compact: Vec<u8> = text.iter().copied().filter(|b| !b.is_ascii_whitespace()).collect();
        if let Ok(decoded) = BASE64_STANDARD.decode(compact) {
            return Psbt::deserialize(&decoded);
        }
    }
    Psbt::deserialize(bytes)
}
```

I'd also rename it to something like `parse_imported_psbt`. The file picker is its only caller, and this leniency is meant for user-supplied files. A name that says so means a future caller re-reading our own output won't pick it up by accident.

frostsnapp/rust/src/api/bitcoin.rs
```diff
@@ -546,10 +547,35 @@ impl AddressExt for bitcoin::Address {
...
+    let text = String::from_utf8_lossy(bytes);
+    let text = text.trim().trim_matches('"').trim();
+
+    if let Ok(decoded) = BASE64_STANDARD.decode(text) {
+        if decoded.starts_with(PSBT_MAGIC) {
+            return Psbt::deserialize(&decoded);
+        }
+    }
+
+    if let Ok(decoded) = hex::decode(text) {
```
LLFourn: On hex: BIP-174 defines two encodings, binary and base64. As far as I can find, no GUI wallet writes hex PSBT files. The only producer I know of is Coldcard's `ckcc sign -x`, an opt-in flag on their CLI. I'd lean towards dropping this branch and keeping the change to what #569 asked for.

If you'd rather keep it, sniff the `70736274ff` prefix the same way as base64 instead of trial-decoding.

(Side note: our own copy button in `psbt.dart` copies the signed PSBT as hex, and `bitcoin-cli decodepsbt` rejects that. That's on us to switch to base64. This PR doesn't need to fix it.)

frostsnapp/rust/Cargo.toml
```diff
@@ -24,6 +25,7 @@ rand = "0.8"
 time = { version = "0.3.36", features = ["formatting"] }
 bip21 = {version = "0.5.0", default-features = false}
 futures.workspace = true
+hex = "0.4"
```
LLFourn: If hex stays, `bitcoin` already re-exports a hex crate as `bitcoin::hex` (`use bitcoin::hex::FromHex; Vec::<u8>::from_hex(s)`), so we don't need to add this one. If hex goes, this line goes with it.

`base64 = "0.22"` is fine: 0.22.1 is already in the lockfile.

frostsnapp/rust/src/api/bitcoin.rs
```diff
@@ -910,4 +936,38 @@ mod test {
...
+    }
+
+    #[test]
+    fn deserialize_psbt_accepts_binary_psbt() {
+        let bytes = valid_psbt_bytes();
+        assert!(deserialize_psbt(&bytes).is_ok());
+    }
+
+    #[test]
+    fn deserialize_psbt_accepts_base64_psbt() {
```
LLFourn: These tests use an empty PSBT with no inputs or outputs. There are real Core-produced PSBTs in `frostsnap_coordinator/tests/fixtures/` (e.g. `core_selfsend.psbt`), and base64-encoding one of those would be a more realistic fixture. It'd also be worth adding cases for base64 wrapped at 76 columns with `\n` and for a BOM + CRLF file. Those are the inputs people actually end up with when they write a file by hand.

PR #591, opened 2026-08-28
fix: prevent duplicate wallet restorations

### LINKED ISSUE #298: Able to recover same wallet in multiple instances
The recovery workflow currently allows multiple instances of recovering the same wallet to be active at the same time.
You can start recovering a wallet and add keys, then start a new recovery flow of the same wallet and add other keys.

We should probably disable recovering a wallet with the same name as a wallet you already have or that is already in the process of being recovered. Also if we try to add a key during a recovery workflow and the key is already added to an existing wallet in recovery, we should probably detect that and let them know (for the case where they are recovering from physical backups and have named the wallet)

LakunleD:
## Summary
Fixes #298 
Prevent the coordinator from restoring the same wallet more than once, whether the wallet is already complete or another restoration is in progress.

## Changes

- Add explicit share search outcomes for found, absent and locked/uncheckable wallets.
- Use access structure identity to locate device shares without decrypting wallets.
- Reject duplicate restorations with recoverable errors instead of assertions.
- Preserve valid restoration behavior when unrelated wallets are locked.
- Propagate restoration errors through the Flutter Rust bridge.
- Add regression coverage for duplicate restorations, conflicting shares, locked wallets and physical backups.
- Update existing share location tests for the new lookup API.

## Testing

Added `frostsnap_core/tests/duplicate_wallet.rs` and updated the existing share location test suite.

### DIFF
--- a/frostsnap_core/src/coordinator.rs
+++ b/frostsnap_core/src/coordinator.rs
@@ -90,6 +90,31 @@ pub enum KeyLocationState {
     },
 }
 
+/// The outcome of [`FrostCoordinator::find_share`].
+#[derive(Debug, Clone, PartialEq)]
+pub enum ShareSearch {
+    Found(ShareLocation),
+    /// The share is in no wallet or restoration this coordinator could check.
+    NotFound,
+    /// The share is in no restoration and in no complete wallet that could be checked, but
+    /// these complete wallets could not be unlocked with the key given, so they could not
+    /// be ruled out.
+    CouldNotCheck {
+        locked_key_names: Vec<String>,
+    },
+}
+
+impl ShareSearch {
+    /// The location if the share was found. A wallet that could not be checked counts as
+    /// not found here, so only use this where that is acceptable.
+    pub fn found(self) -> Option<ShareLocation> {
+        match self {
+            ShareSearch::Found(location) => Some(location),
+            ShareSearch::NotFound | ShareSearch::CouldNotCheck { .. } => None,
+        }
+    }
+}
+
 impl CompleteKey {
     pub fn coord_share_decryption_contrib(
         &self,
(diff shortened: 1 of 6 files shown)


frostsnap_core/src/coordinator/restoration.rs
```diff
@@ -555,10 +572,24 @@ impl FrostCoordinator {
...
-            self.find_share(recover_share.held_share.share_image, encryption_key)
+        // Check if the share exists elsewhere. When this restoration already knows its
+        // wallet and the share names the same one, identity is settled: another restoration
+        // claiming that wallet (a duplicate from before that was prevented) must not stop
+        // this one continuing, so only a restoration actually holding the share counts.
+        let share_image = recover_share.held_share.share_image;
+        let share_ref = recover_share.held_share.access_structure_ref;
+        let found = if share_ref.is_some()
+            && share_ref == restoration.access_structure.access_structure_ref()
         {
```
LLFourn: When the share names the same wallet as this restoration, this branch only asks whether another *restoration* holds the share, so the check for a *complete* wallet is skipped too. That drops a check the code had before this PR.

It shows up when the wallet is completed some other way while this restoration is still open. For example, restoration A is started from a device share, and restoration C is built from physical backups of the same wallet and finishes first. Adding another device share to A then returns `Ok(())`, where before this PR it returned `ShareBelongsElsewhere { Complete }`, telling the user to add the device to the existing wallet instead. The same thing happens with two legacy restorations of one wallet once one of them finishes. I checked it with your helpers from `duplicate_wallet.rs`:

```rust
let shares = device_shares(&run);
run.clear_coordinator();
run.coordinator
    .start_restoring_key_from_recover_share(&shares[0], RestorationId::new(&mut rng))
    .unwrap();
let a = run.coordinator.restoring().next().unwrap().restoration_id;
let c = restoration_holding(&mut run, &without_wallet_identity(&shares[1]), &mut rng);
run.coordinator
    .add_recovery_share_to_restoration(c, &without_wallet_identity(&shares[2]), TEST_ENCRYPTION_KEY)
    .unwrap();
run.coordinator.finish_restoring(c, TEST_ENCRYPTION_KEY, &mut rng).unwrap();
// The wallet is now complete. This returns Ok(()); on master it is ShareBelongsElsewhere { Complete }.
run.coordinator
    .check_recover_share_compatible_with_restoration(a, &shares[1], TEST_ENCRYPTION_KEY)
```

Of the three lookups in `locate_without_key`, this path only needs to skip the last one, "a restoration claims this wallet". The complete-wallet lookup by ref and the lookup for a restoration that holds the image should still run. I'd move the claim lookup out of `locate_without_key` into its own function, e.g. `restoration_claiming(access_structure_ref)`, so that `find_share` means "who holds this share". The callers that care about claims then add it themselves: `start_restoring_key_from_recover_share`, `check_start_restoring_key_from_device_share`, and this function when the share's ref differs from the restoration's. The `if` here then goes away. Could you add the scenario above as a test too?

frostsnap_core/src/coordinator.rs
```diff
@@ -90,6 +90,31 @@ pub enum KeyLocationState {
...
+/// The outcome of [`FrostCoordinator::find_share`].
+#[derive(Debug, Clone, PartialEq)]
+pub enum ShareSearch {
+    Found(ShareLocation),
+    /// The share is in no wallet or restoration this coordinator could check.
+    NotFound,
+    /// The share is in no restoration and in no complete wallet that could be checked, but
+    /// these complete wallets could not be unlocked with the key given, so they could not
+    /// be ruled out.
+    CouldNotCheck {
```
LLFourn: Every non-test caller of `find_share` calls `.found()` right away, so `CouldNotCheck` never reaches the app. The comments at those call sites explain why ignoring it is right in each flow, which means nothing acts on the three-way outcome. I'd keep `find_share` returning `Option<ShareLocation>` and drop `ShareSearch`/`found()`. That also removes the `.found()` changes in `share_location.rs`, and `a_locked_wallet_is_reported_rather_than_ignored` can check for `None`.

The new `access_structure_ref` parameter is the real improvement here, and it works without this enum. If we later want to tell the user that a locked wallet couldn't be checked, the variant can come back together with the UI that shows it.

frostsnap_core/src/coordinator/restoration.rs
```diff
@@ -810,9 +848,22 @@ impl FrostCoordinator {
             .purpose
             .ok_or(StartRestorationFromShareError::MissingMetadata)?;
 
-        assert!(!self.restoration.restorations.contains_key(&restoration_id));
-        if let Some(access_structure_ref) = held_share.access_structure_ref {
-            assert!(self.get_access_structure(access_structure_ref).is_none());
+        if self.restoration.restorations.contains_key(&restoration_id) {
+            return Err(StartRestorationFromShareError::RestorationIdInUse);
```
LLFourn: Every caller passes a freshly random `RestorationId::new(rng)`, so a collision here would be a bug in the caller, not something the user can do anything about. That is why it was an `assert!`. As a public error variant it also shows up in the FRB mirror, where Dart can never show it meaningfully. I'd keep the `assert!` for this one. The other one, on the access structure ref, was the right one to turn into `ShareBelongsElsewhere`.

(Good catch, by the way, that the FFI wrapper used to ignore this `Result`. `MissingMetadata` returned the id of a restoration that was never created, and the new `?` fixes that.)

frostsnapp/rust/src/coordinator.rs
```diff
@@ -813,19 +813,12 @@ impl FfiCoordinator {
...
-                    .is_some()
-                {
-                    return Err(anyhow!("we already know about this access structure"));
-                }
-            }
             let mut db = self.db.lock().unwrap();
             coordinator.staged_mutate(&mut *db, |coordinator| {
                 let restoration_id = RestorationId::new(&mut rand::thread_rng());
-                coordinator.start_restoring_key_from_recover_share(recover_share, restoration_id);
+                // Refuses a wallet that already exists or is already being restored.
```
LLFourn: nit: in this codebase comments explain *why*, not what the code does. This one restates the call below, so it can go. Several of the new comments in `coordinator.rs`/`restoration.rs` are the same, e.g. the docs on `share_location` and `locate_in_restorations`. The `find_share` doc is long enough that it's hard to find the contract in it. Cutting these down to the reasons you can't see from the code (e.g. why a device's ref decides whether a share belongs to a complete wallet) would make this much easier to maintain.

PR #594, opened 2026-09-01
Upgrade bdk_electrum_streaming to 0.7, persist its Cache

### LINKED ISSUE #577: Incorrect `bdk_electrum_streaming::Cache` handling
## Description

https://github.com/frostsnap/frostsnap/blob/bdaa8be311e4ed01c89984a0f442c5588aa36304/frostsnap_coordinator/src/bitcoin/chain_sync.rs#L165

`Cache` should be rebuilt from wallet data otherwise tx evictions will not be detected while chain-sync is offline.

Fields that must be populated to avoid incorrect behavior:

* `Cache::spk_txids`

Fields that can be populated to save bandwidth:

* `Cache::txs`
* `Cache::anchors`
* `Cache::headers` (we aren't persisting headers currently - a good to have).

## To Replicate (not tested)

* Send pending incoming tx.
* Wait for wallet to sync.
* Shut wallet down completely.
* Cancel incoming tx.
* Start wallet up again, wait for sync to finish.

Result: cancelled tx is still there.

evanlinjin:
Fixes #577.

## Summary

Upgrades `bdk_electrum_streaming` 0.5.3 → 0.7 and gives its `Cache` the handling #577 asks for: the derived half rebuilt from wallet data on startup, the rest persisted.

**This PR fixes:**
- Txs evicted when disconnected to electrum server not being detected after restart.
- Last used index not updating correctly.
- Txs potentially missing anchors after reorg.

**This PR improves:**
- Reduce bandwidth use by persisting cache.

## ⚠️ Needs physical testing

0.7 restructures how work is scheduled, so **transactions may take longer to appear in the wallet during a sync** than they did on 0.5.3. This is a behavioural change in the upstream architecture, not something this PR configures, and it is not covered by the unit tests here — it needs testing against a real server and a real device.

If the delay turns out to be bad enough to be user-visible, the follow-up is a sync progress indicator so the wallet isn't silently empty-looking while work is still in flight.

## Against #577's checklist

| Field | #577 says | This PR |
| --- | --- | --- |
| `spk_txids` | must be populated, else evictions are missed | seeded from the wallet on every startup |
| `txs` | can be populated to save bandwidth | seeded (as before) |
| `anchors` | can be populated to save bandwidth | seeded (as before) |
| `headers` | not persisted currently — good to have | persisted |

## Changes

- **`Cache` API move**: `Cache::txs`/`anchors` moved under a new `tx_cache` sub-struct in 0.7, so `chain_sync.rs` seeds them at `self.cache.tx_cache.{txs,anchors}`. `tx_cache` is never persisted (it's `#[serde(skip)]` upstream) — it's wallet-derived, so it is rebuilt on startup instead.
- **`spk_txids` seeding** (the correctness half of #577): eviction is reported by diffing the txids we already hold for a script against the history the server now gives, so a txid we were never told about cannot go missing. Nothing was seeding that set, so after a restart the first history fetch had nothing to diff against and an eviction that happened while we were offline was invisible until that script's history changed again. Now rebuilt from the wallet via `CoordSuperWallet::spk_txid_cache()`, mirroring Electrum's definition of a script's history (a tx is in it if it pays the script or spends an output of it).
- **Cache persistence** (the bandwidth half): new `frostsnap_coordinator/src/bitcoin/electrum_cache.rs` wraps the cache in an `ElectrumCache` implementing this crate's `Persist<rusqlite::Connection>`, storing `subscriptions` + `headers` as a per-network `bincode` blob. Loaded in `ChainClient::new` (replacing the `Cache::default()` the issue links to), written back on every exit from a connection (graceful stop, error, reconnect, or shutdown) as a best-effort checkpoint.
- **`ChainClient::new`** now takes `network` and derives `genesis_hash` from it, rather than taking both.

## Test plan
- [x] `cargo test -p frostsnap_coordinator -p rust_lib_frostsnapp --lib` — 58 + 10 pass
- [x] New test encodes #577's repro (pending tx → shutdown → tx dropped → restart): asserts the txid comes back in `evicted_ats`. Verified it fails (`got {}`) with the `spk_txids` seeding removed, so it is not vacuous.
- [x] New round-trip test asserting `headers` persist and `tx_cache` never does
- [x] `just build` — full Linux app build succeeds
- [ ] **Physical test against a real server + device**: how long txs take to show up during a sync (see above)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -350,9 +350,9 @@ dependencies = [
 
 [[package]]
 name = "bdk_electrum_streaming"
-version = "0.5.3"
+version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a16576b1efa0960f4c6d65d588005889ee9bee600d47540adca62d21003154bc"
+checksum = "0b76decece975a438ab263082cb223757dd4137b0f3ff7aebdd868a75cb46875"
 dependencies = [
  "anyhow",
  "bdk_core",
@@ -360,6 +360,7 @@ dependencies = [
  "futures",
  "futures-timer",
  "miniscript",
+ "serde",
  "serde_json",
  "tracing",
 ]
--- a/frostsnap_coordinator/Cargo.toml
+++ b/frostsnap_coordinator/Cargo.toml
@@ -23,7 +23,7 @@ rsa.workspace = true
 serde = { version = "1.0", features = ["derive"] }
 
 bdk_chain = { version = "0.23.3", features = ["rusqlite"] }
-bdk_electrum_streaming = { version = "0.5.3" }
+bdk_electrum_streaming = { version = "0.7" }
 bdk_coin_select = { version = "0.4.1" }
 
 
--- a/frostsnap_coordinator/src/bitcoin.rs
+++ b/frostsnap_coordinator/src/bitcoin.rs
@@ -1,4 +1,5 @@
 pub mod chain_sync;
+mod electrum_cache;
 mod handler_state;
 pub mod outgoing;
 pub mod psbt;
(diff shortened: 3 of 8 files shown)


evanlinjin:
Replacement PR incoming.

PR #599, opened 2026-09-09
[app] Make validate_amount network-independent

### LINKED ISSUE #598: refactor(wallet): decouple network validation from `validate_amount` and flatten parsing logic
- `validate_amount` previously required a `&self` reciever on `BitcoinNetwork` to enforce network checks before calculating dust thresholds. 

- a silent bug existed where address network mismatches returned `None`, falsely indicating to the caller that an invalid or misconfigured amount was valid .
- deeply nested `match` statements mad the addrss parsing flow hard to read and maintain

7-falseparfait:
fixes #598 

Changes

* **Signature:** moved `validate_amount` out of `BitcoinNetworkExt` into a standalone `pub fn validate_amount(address: &str, value: u64) -> Option<String>`, resolving the long-standing `FIXME` ("doesn't need to be on the network").

* **Network Decoupling:** replaced `.require_network(*self)` with `.assume_checked()`. The dust check only needs the script pubkey, which the address payload determines on its own...this fixes a silent bypass where a wrong-network address previously returned `None` (no error warning) instead of being dust-checked.
* **Flow Control:** flattened nested `match` blocks into a single `let Ok(address) = ... else { return None; }` guard.
* **Bindings:** re-ran FRB codegen, updating the Dart API to a top-level `validateAmount(...)` function instead of an extension method on `BitcoinNetwork`. no existing call sites in rust or dart required updates.

Verification: 
* **`cargo check` & `cargo test`:** Passed cleanly across all workspace default members.
* **Formatting & Analysis:** `cargo fmt --check` and `flutter analyze` both passed with zero issues.
* **`cargo clippy -Dwarnings`:** Clean across all touched crates (pre-existing `uninlined_format_args` warnings in untouched `desktop_camera`/`frost_backup` remain).

### DIFF
--- a/frostsnapp/rust/src/api/bitcoin.rs
+++ b/frostsnapp/rust/src/api/bitcoin.rs
@@ -74,9 +74,6 @@ pub trait BitcoinNetworkExt {
     fn bdk_file(&self, app_dir: impl AsRef<Path>) -> PathBuf;
 
     #[frb(sync)]
-    fn validate_amount(&self, address: &str, value: u64) -> Option<String>;
-
-    #[frb(sync)]
     fn supported_networks() -> Vec<BitcoinNetwork>;
 }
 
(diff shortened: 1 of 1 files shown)


frostsnapp/rust/src/api/bitcoin.rs
```diff
@@ -151,30 +148,28 @@ impl BitcoinNetworkExt for BitcoinNetwork {
...
-                    } else {
-                        None
-                    }
-                }
-                Err(_e) => None,
-            },
-            Err(_e) => None,
-        }
+#[frb(sync)]
+pub fn validate_amount(address: &str, value: u64) -> Option<String> {
```
LLFourn: Thanks for picking up this FIXME, and for the clear write-up. The `let … else` version does read better than the nested match.

I think the better fix is to delete `validate_amount` entirely, because nothing calls it. Its last caller, the amount-field validator in `wallet_send.dart`, was removed in c270d929 ("Rework Send Bitcoin flow"). I regenerated the bindings at this head and `grep -rn validateAmount frostsnapp/lib` finds only the generated definition. (That's also why no call sites needed updating, as your PR description says.)

The send flow does both checks elsewhere:
- **Network:** `BitcoinNetworkExt::validate_destination_address` (L106–L130) calls `require_network` on both the BIP21 and the plain-address branches. A `tb1…` address on mainnet comes back as `Err("Wrong network: …")`.
- **Dust:** `SendState::amount` in `api/transaction.rs` (L538–L541) computes `minimal_non_dust()` on the recipient's already-checked `Address` and returns `AmountError::AmountBelowDust`.

So this function's behaviour change affects no user, and the "silent bypass" it fixes was never reachable. Keeping a public, network-free `validateAmount` in the Dart API also leaves a trap: for a wrong-network address above dust it returns `None`, which a caller reads as "valid". If someone reached for it later instead of `SendState`, they'd lose the network check.

Suggested change: remove both the trait method and its impl (the whole FIXME block), rather than moving it. That still closes #598.

LLFourn (changes requested):

7-falseparfait:
thanks for the catch, makes total sense....since network and dust checks are already handled in the send flow, deleting the dead code /TODO block is way cleaner.

i've removed `validate_amount` entirely, rebased onto upstream master, and regenerated the FRB bindings.

PR #603, opened 2026-09-23
[device] Upgrade to esp-hal 1.2

LLFourn:
Moves the device firmware from esp-hal 0.22 (and our fork of it) to upstream esp-hal 1.2.1, so low-level code such as OTA and partition handling can move onto what Espressif curates.

Supersedes #457 and #601. It is the same final tree as #601 (the #457 port rebased onto master without the unrelated core-module-move commit, plus the review fixes below), restructured into a few clean commits:

1. **[deps,nix] Move to Rust 1.95**: esp-hal 1.2 requires it. Also bumps the nix rust-overlay pin, whose age was what broke the deterministic build on #457.
2. **[device,widgets] Reduce the largest stack frames**: 1.2 leaves less stack (~28 KB vs ~32 KB), so this makes room before the upgrade and every commit passes the 25% stack-check.
3. **[device] Upgrade to esp-hal 1.2**: the port itself, with fixes for behaviour 1.2 changed under our code folded in:
   - A UART RX overflow now panics naming the UART. 1.2 resets the whole FIFO on overflow, and the port was silently swallowing that.
   - The UART TX drain wait happens outside the critical section. The port's blocking flush masked the other UART's RX interrupt meanwhile.
   - One UART config function builds every UART config.
   - The touch I2C bus timeout is restored to 0.22's `BusCycles(10)`.
   - The panic handler no longer heap-allocates its display buffer.
   - SPI stays at 80 MHz, as on master.
4. **[device] Read the partition table with esp-bootloader-esp-idf**.
5. **[factory,nix] Build the bootloader from ESP-IDF v5.5.4**: lands after the upgrade, because v5.4+ validates the app descriptor that the upgrade adds.
6. **[device] Remove unused cfg_example.toml**.

No wire-protocol or on-device data format changes. `frostsnap_core` is untouched.

**Checked locally**
- Every commit passes device lint, the legacy and frontier builds, and `stack-check --max-pct 25` (tip: 22.8%).
- The tip passes `lint-ordinary` and `test-ordinary` (253 tests).

**Still needs hardware**
- OTA across a chain of 2+ devices, and hot-plugging a device mid-chain (neither should hit the overflow panic).
- Touch works normally.
- A forced panic draws the error screen.
- `genuine-check` on a provisioned device for DS signing.
- A legacy device boots after `just legacy-flash`.

**Open**
The `#[inline(never)]` on `CheckBackupScreen::draw` (in commit 2) passes the single-frame stack gate by splitting a frame, but deepens the check-backup draw path. A static worst-case call-depth analysis is under way separately and will decide whether it stays.


frostsnap_widgets/src/backup/check_backup.rs
```diff
@@ -629,6 +629,8 @@ impl DynWidget for CheckBackupScreen {
 impl Widget for CheckBackupScreen {
     type Color = Rgb565;
 
+    // Keep this screen's rendering and construction temporaries out of WidgetTree::draw.
+    #[inline(never)]
```
LLFourn, line 633: **No change in this PR; follow-up PR requested.** Keep this attribute for now.

**Note:** this passes the single-frame gate by splitting `WidgetTree::draw`'s 7,504 B frame into 4,320 + 4,384. The static depth analysis (`stack_check depth`, `DEPTH_ANALYSIS.md` on `stack-depth-analysis`) finds that it makes the check-backup path about 1.2 KB *deeper* (21.7 KB → 22.9 KB). That path isn't the deepest, so it doesn't change the worst case, but it trades real depth for a proxy. The underlying problem is that the port takes ~4 KB of stack away (≈32 KB → 28,156 B) while direct-call chains alone are estimated at 84–95% of what is left, before 407 unfollowed indirect calls. The single-frame gate can't see that.

**Follow-up PR:** build an enhanced stack depth checker. Make `stack_check depth` a fail-closed CI gate on the deepest call chain rather than the largest frame:
- every reachable frame is known;
- every indirect call is resolved, including the `dyn` widget dispatch;
- recursion is bounded, including the three cycles in our own code;
- interrupt and panic routes are counted: interrupt on the main path ~784 B, panic ~2,992 B, panic inside the UART ISR ~3.3 KB.

With a real depth number in CI, that PR decides the budget on evidence: shrink the largest UI construction frames (`DeviceLoop::poll`, `FeePage::new`, `set_workflow`, `QuizButtons::new`), and/or move stack back from the 256 KB heap if measured heap peaks allow. Then this attribute can go.

device/src/uart_interrupt.rs
```diff
@@ -182,9 +194,28 @@ impl UartHandle {
     }
 }
 
+/// Panics if either UART's RX FIFO has overflowed, which lost up to a FIFO's worth of received
+/// bytes. The ISR only latches the overflow because it must not panic; call this from task context.
+pub fn panic_on_rx_overflow() {
```
LLFourn, line 199: **Change requested:** remove the new overflow handling and let the error go up the existing read path.

**Change:** remove the `RX_OVERFLOWED` latch, `panic_on_rx_overflow()`, and its calls in `DeviceLoop::poll` and `SerialIo::read_byte`. They're new functionality that master didn't have, and I'd keep this PR to the port. Dropped data is getting its own story.

**Change:** 1.2 does report the overflow, and that's fine to keep. Have `drain_chunk` note it for the port, return it from the task-side read as an error, and let it come out of `SerialInterface::receive()` like any other bad read. `esp32_run` already decides what happens then: upstream panics (`"upstream read fail"`), and downstream reports it and drops the connection. The OTA loop reads `SerialIo::read_byte` directly, so give it a plain `.expect()` there.

**PR description:** add one line, since the upgrade can't avoid it. On overflow, 1.2's `read_buffered` resets the whole RX FIFO (`check_for_errors_and_reset_fifo`), where 0.22 kept the 128 bytes already in it and dropped only the excess.

frostsnap_factory/bootloader/flake.nix
```diff
@@ -9,11 +9,12 @@
...
       pkgs = import nixpkgs { inherit system; };
-      # Override to ESP-IDF v5.3.1 — v5.4+ added esp_app_desc_t efuse_blk_rev validation
-      # that rejects bare-metal esp-hal images missing a proper app descriptor.
+      # Pin ESP-IDF to v5.5.4. v5.4+ enforces esp_app_desc_t efuse_blk_rev
+      # validation, rejecting images without a proper app descriptor; every
+      # firmware binary declares one with esp_app_desc!().
       esp-idf-riscv = (nixpkgs-esp-dev.packages.${system}.esp-idf-riscv).override {
-        rev = "v5.3.1";
-        sha256 = "sha256-hcE4Tr5PTRQjfiRYgvLB1+8sR7KQQ1TnQJqViodGdBw=";
+        rev = "v5.5.4";
```
LLFourn, line 16: **Change requested (docs) + hardware checks before merge.**

**Note:** the bootloader never changes on a device after provisioning, because OTA writes only `ota_0`/`ota_1`. So the bootloader that matters for this PR is the one shipped devices already have. `strings frostsnap_factory/bootloader/prod/signed-bootloader.bin` reports **ESP-IDF v5.1.6**, which is neither the v5.3.1 this flake used to build nor the v5.5.4 it builds now. `bootloader-legacy.bin` is v5.1.6 too.

**Hardware check (before merge):** moving dev devices to v5.5.4 makes them **less** like production, because a dev device provisioned from this branch runs the new firmware under a bootloader no customer has. OTA this firmware onto a device running the v5.1.6 production bootloader (or a dev device reflashed with a v5.1.6 dev bootloader), and boot it from both slots.

**Change (`PROVISIONING.md`, in this PR):** the flake can't reproduce the committed prod `signed-bootloader.bin`, and before this PR it couldn't either. Add a line saying prod stays on the committed v5.1.6 binary until someone re-signs, so the v5.5.4 bump isn't read as a production change.

**Hardware check + change (`PROVISIONING.md`):** a v5.5.4 bootloader rejects images without an app descriptor, so a device provisioned from this branch can't boot any pre-#603 firmware. The app refuses downgrades, so users can't reach this state, but `just flash` of an older commit onto such a dev device will fail to boot that slot. Confirm on hardware that it falls back to the other slot instead of boot-looping, and note the limitation in `PROVISIONING.md`.

device/src/bin/frontier.rs
```diff
@@ -6,21 +6,20 @@
 extern crate alloc;
 
 use core::cell::RefCell;
-use esp_hal::entry;
+use esp_hal::main;
 use esp_storage::FlashStorage;
 use frostsnap_device::{esp32_run, peripherals::DevicePeripherals, resources::Resources};
 
-#[entry]
+esp_bootloader_esp_idf::esp_app_desc!();
```
LLFourn, line 13: **Change requested:** pin `SOURCE_DATE_EPOCH` in the root `.cargo/config.toml` (below).

**Note:** `esp_app_desc!()` embeds a build time and date. esp-bootloader-esp-idf 0.5's build script takes them from `SOURCE_DATE_EPOCH`, or from `Timestamp::now()` when that is unset. The nix dev shell sets `SOURCE_DATE_EPOCH=1704067200`, so `deterministic-build.sh` should stay reproducible. Outside nix, clean builds now differ from one another, because the build script's output is cached and only a fresh one picks up a new time. A local `just build-firmware frontier` here embeds `2026-09-24` / `03:02:13` (`strings target/riscv32imc-unknown-none-elf/release/frontier`). That includes the `dev-frontier.bin` CI builds and bundles into the app.

**Notes, no change needed in this PR.** Two more things make this easy to miss:
- The deterministic-build workflow is `continue-on-error` and never compares hashes, so its green tick shows only that the build succeeded, not that two builds match.
- The build script parses `SOURCE_DATE_EPOCH` with `Timestamp::from_microsecond`, so the nix build's descriptor says 1970-01-01 00:28:24. That's harmless here. **Optional:** report it upstream.

**Change:** pin it next to the partition offset in the root `.cargo/config.toml`, which removes this timestamp as a source of nondeterminism whether or not the build runs under nix. Byte-for-byte reproducibility would still need two clean builds whose hashes match. `[env]` doesn't override a variable that is already set, so nix's value still wins there:

```toml
[env]
ESP_BOOTLOADER_ESP_IDF_CONFIG_PARTITION_TABLE_OFFSET = "0xD000"
SOURCE_DATE_EPOCH = "1704067200"
```

One caveat: the build script emits no `rerun-if-env-changed=SOURCE_DATE_EPOCH`, so this only takes effect after `cargo clean -p esp-bootloader-esp-idf`.

device/src/partitions.rs
```diff
@@ -29,42 +30,41 @@ impl<'a> Partitions<'a> {
...
-        for row in table.iter_storage(&mut *flash.borrow_mut(), false) {
-            let row = match row {
-                Ok(row) => row,
-                Err(_) => panic!("unable to read row of partition table"),
-            };
-            assert_eq!(row.offset % FlashStorage::ERASE_SIZE as u32, 0);
-            match row.name() {
+        // Boxed: ~3KB, too big for the stack of a boot-time function.
+        let mut pt_mem = alloc::boxed::Box::new([0u8; partitions::PARTITION_TABLE_MAX_LEN]);
+        let pt = partitions::read_partition_table(&mut *flash.borrow_mut(), &mut *pt_mem)
```
LLFourn, line 40: **No code change; PR description line requested.**

**Note:** this makes the partition table stricter at boot. The old code called `iter_storage(.., false)`, which skipped the MD5 check. esp-bootloader-esp-idf 0.5 has its `validation` feature on by default: it requires an MD5 row and a matching hash, and otherwise returns `Err(Invalid)`, which this `expect` turns into a panic on every boot, before any UI.

I checked that no device we've flashed can have a table without that row:
- **Frontier:** tables are written from `device/partitions.bin`. It has only ever had one committed version (`44de3e8c`), and that version has a valid MD5 row at index 5.
- **Legacy:** `just legacy-flash` regenerates the table from `partitions.csv` on every flash. espflash 3.2.0, the version pinned in CI, produces a table byte-identical to `partitions.bin`. espflash 2.x–4.x all generate tables through esp-idf-part, whose `to_bin` writes the MD5 row unconditionally (0.1.0 onwards), and espflash 1.5–1.7 wrote it too. Only espflash ≤1.4 (April 2022) lacked it, a year before the custom table at 0xD000 existed (`a84f3776`, April 2023).

**Note:** keeping validation on is right, and there's no compatibility concern.

**PR description:** add one line saying partition-table MD5 validation is now enforced at boot.

device/src/peripherals.rs
```diff
@@ -170,93 +170,86 @@ fn extract_entropy(
...
         let mut delay = Delay::new();
 
         // Initialize SHA256 early for entropy extraction
-        let mut sha256 = esp_hal::sha::Sha::new(&mut peripherals.SHA);
+        let mut sha256 = esp_hal::sha::Sha::new(peripherals.SHA);
 
         // Get initial entropy from hardware RNG mixed with SHA256
-        let mut trng = esp_hal::rng::Trng::new(&mut peripherals.RNG, &mut peripherals.ADC1);
+        let trng_source =
+            esp_hal::rng::TrngSource::new(peripherals.RNG.reborrow(), peripherals.ADC1.reborrow());
```
LLFourn, line 195: **No code change; PR description line requested.**

**PR description:** add a line saying esp-hal 1.x now enforces Espressif's minimum spacing between RNG reads, 16 APB cycles or 200 ns on the C3 (`rng/ll.rs`, `rng.apb_cycle_wait_num = 16`).

**Note:** 0.22 had no limiter, so the spacing in `extract_entropy` came entirely from the SHA update between chunks. Now the HAL guarantees it, and a future change to `extract_entropy` can't silently lose it. `TrngSource` keeps the SAR ADC source on for the whole extraction, and dropping it afterwards is fine because nothing reads the hardware RNG again.

device/src/frosty_ui.rs
```diff
@@ -307,20 +298,24 @@ impl<'a> UserInteraction for FrostyUi<'a> {
                 widget: sign_prompt,
                 phase,
             } => {
-                if sign_prompt.is_confirmed() {
-                    if let Some(phase_data) = phase.take() {
-                        return Some(UiEvent::SigningConfirm { phase: phase_data });
-                    }
+                if let Some(phase_data) = if sign_prompt.is_confirmed() {
```
LLFourn, line 301: **Nit (optional change):** `if let Some(x) = if c { phase.take() } else { None }` takes a second read. If it was written for 1.95's `collapsible_match`, `Option::take_if` says the same thing directly and is clippy-clean on 1.95 and current stable (checked in a scratch crate). Same at L313:

```rust
if let Some(phase_data) = phase.take_if(|_| sign_prompt.is_confirmed()) {
    return Some(UiEvent::SigningConfirm { phase: phase_data });
}
```

LLFourn (commented):
clankers round 1

